PR #4, opened 2019-07-19
Port of downstream operator

sthaha:
Initial port of OpenShift/pipelines-operator

This is the initial version of the tekton operator that installs
0.5.2 version of tekton pipelines and is based on the
openshift-pipeline-operator:
see: openshift/tektoncd-pipeline-operator@039bd4c

The operator automatically installs pipelines when started by
- creating a `config.operator.tekton.dev` CR with the name `cluster`
- Once the cluster `cr` is created, it installs tekton-pipelines and
  updates the status of the `cluster` CR to indicate the status of
  installation of pipelines.

Signed-off-by: sthaha@redhat.com


test/presubmit-tests.sh
```diff
@@ -27,10 +27,45 @@ export DISABLE_MD_LINTING=1
...
+  GO111MODULE=on operator-sdk build gcr.io/tekton-nightly/tektoncd-operator
+}
+
+install_operator_sdk() {
+  local sdk_rel="v0.9.0"
+  mkdir -p tmp/bin/
+
+  curl -JL \
+    https://github.com/operator-framework/operator-sdk/releases/download/${RELEASE_VERSION}/operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu \
+    -o tmp/bin/operator-sdk
```
vdemeester: Put it in /use/bin so that you don’t need to play around with the path 👼 it’s simpler

test/presubmit-tests.sh
```diff
@@ -27,10 +27,45 @@ export DISABLE_MD_LINTING=1
...
+  chmod +x tmp/bin/operator-sdk
+}
+
+post_build_tests() {
+  ## TODO: enable this back on the next pr
+  #golangci-lint run
+  GO111MODULE=on operator-sdk test local ./test/e2e  \
```
vdemeester: This should be in an integration-tests function instead of a post build. The build job only build, the e2e tests are meant to run in the intergration job (in parallel with build) 👼

vincent-pli:
🎉 🎉 🎉

Issue #11, opened 2019-08-07
Add ko support to build & publish the image, even install the operator


houshengbo:
# Expected Behavior
When we run the command `ko apply -f config`, operator should be installed from the source code.

ko should be the major command to build, publish, and install tekton projects.

# Actual Behavior
We do not support. We rely on operator sdk to do everything.

# Steps to Reproduce the Problem

1. No support. No command available to run.

# Additional Info

PR #21, opened 2019-08-10
Enable ko support to build & install

houshengbo:
# Changes
* This PR adds the ko support, so that operator can be installed with the command `ko apply -f config/`.
* We should set GO111MODULE=on, when in need. This should not work as the default option.
* A role for `rbac.authorization.k8s.io` is added for tekton-operator.

Fixes: #11 #40

# Submitter Checklist

review them:

# Release Notes

```
Describe any user facing changes here, or delete this block.

Examples of user facing changes:
- API changes
- Bug fixes
- Any changes in behavior

```


houshengbo:
I will create the the new integration tests to verify if the CR can be installed, after operator is installed. Remove the testmain to make sure the integration test job is happy.

sthaha:
@houshengbo @vdemeester @nikhil-thomas , Could we hold this PR until the switch away from `operator-sdk` can be discussed in a WG meeting and we get inputs from how knative operator is shipped into operatorhub.io without relying on operator-sdk and related tools?

vdemeester:
/hold
@sthaha yes. I'm willing to prioritize the features (addons support, …), testing coverage and publishing a first version of the operator before change the tooling. 

Operator developers are either used to `operator-sdk` or `controller-runtime` more than `knative/pkg`. We might need to discuss that in a call (maybe we can setup a operator sync, the same way `tektoncd/triggers` does).

houshengbo:
@vdemeester We can cancel the hold as well, since ko support does not influence the operator-sdk. It uses a different operator.yaml under config/.

config/operator.yaml
```diff
@@ -0,0 +1,28 @@
...
+  template:
+    metadata:
+      labels:
+        name: tekton-operator
+    spec:
+      serviceAccountName: tekton-operator
+      containers:
+        - name: tekton-operator
+          image: github.com/tektoncd/operator/cmd/manager
+          imagePullPolicy: IfNotPresent
```
vdemeester: Not sure about this one :angel: (we don't set this on pipeline, …)

houshengbo: removed it.

test/testgroups/cluster_crd.go
```diff
@@ -3,50 +3,15 @@ package testgroups
...
 	defer ctx.Cleanup()
 
-	err := deployOperator(t, ctx)
-	helpers.AssertNoError(t, err)
-
 	t.Run("auto-installs-pipelines", testsuites.ValidateAutoInstall)
 	t.Run("delete-pipelines", testsuites.ValidateDeletion)
 }
-
-func deployOperator(t *testing.T, ctx *test.TestCtx) error {
```
vdemeester: Why is this removed ?

houshengbo: function tekton_setup() in the script will launch the operator, so we do not need to deploy the operator here.

deploy/operator.yaml
```diff
@@ -15,11 +15,8 @@ spec:
       serviceAccountName: tekton-operator
```
vincent-pli: Should restructure the `/deploy` or `/config`,  add number as prefix of files, like 500-operator.yaml, it will keep the order when run `ko apply`,  avoid exception caused by the wrong order.

README.md
```diff
@@ -9,7 +9,7 @@ The following steps will install [Tekton Pipeline](https://github.com/tektoncd/p
     
 2. Apply Operator CRD
 
-    `kubectl apply -f deploy/crds/*_crd.yaml`
+    `kubectl apply -f deploy/crds/`
```
vincent-pli: If we turn to `ko`, the `crd` should also handle by `ko apply`, so the `crd` definition file should put in `/deploy/`

houshengbo: We do not need to use ko on top of the crds, because there is no image building or pulling involved so far.

deploy/operator.yaml
houshengbo: So far, the CRD installation still goes before the operator installation as prerequisite. We may need to do the ordering later if we decide to move the CRD into the operator installation as well.

README.md
nikhil-thomas: @vincent-pli @houshengbo
I agree with vincent-pli
as we are providing `ko` support, it will be better if we can move all manifests to a single location and number them.

my only concern was the `operator-sdk olm-catalog gen-csv` command which generates a ClusterServiceVersion (CSV) file to be published on OperatorHub. By default, the command assumes the operator.yaml to be in deploy/ and crds in deploy/crd/.

However, the command can accept a config file where we can specify the paths,
eg: https://github.com/operator-framework/operator-sdk/blob/master/internal/pkg/scaffold/olm-catalog/testdata/deploy/olm-catalog/csv-config.yaml

nikhil-thomas: @houshengbo could you make this update

in fact, if we decide to move to knative-pkg injection-controller, this part can stay the same.

nikhil-thomas: >  there is no image building or pulling involved so far.

true, but if we keep them separate we will have to do a `kubectl apply -f <crd path>` and `ko apply config`
i would prefer haveing a single directory with numbered files

houshengbo: @nikhil-thomas I can use the file link instead of the physical file. We can still keep crd under deploy/crds, but file-link it under config. openrator-sdk works with deploy dir and ko works with config dir.

test/e2e-tests.sh
```diff
@@ -14,15 +14,35 @@
...
+  header "Installing Tekton Operator"
+  kubectl create namespace $TEST_NAMESPACE
+  kubectl apply -n $TEST_NAMESPACE -f deploy/crds/
+  ko apply -n $TEST_NAMESPACE -f config/
+  wait_until_pods_running $TEST_NAMESPACE
+}
+
+function tekton_teardown() {
+  header "Removing Tekton Operator"
+  kubectl delete --ignore-not-found Config cluster
```
nikhil-thomas: use fully qualified name
`config.opeator.tekton.dev`

test/e2e-tests.sh
```diff
@@ -14,15 +14,35 @@
...
-export GO111MODULE=on
 source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
 
+# Namespace used for tests
+readonly TEST_NAMESPACE="operator-tests"
+
+function tekton_setup() {
+  header "Installing Tekton Operator"
+  kubectl create namespace $TEST_NAMESPACE
+  kubectl apply -n $TEST_NAMESPACE -f deploy/crds/
```
nikhil-thomas: why do we need `-n $TEST_NAMESPACE` while applying CRDs as CRDs are clusterwide

test/e2e-tests.sh
```diff
@@ -14,15 +14,35 @@
...
+  ko apply -n $TEST_NAMESPACE -f config/
+  wait_until_pods_running $TEST_NAMESPACE
+}
+
+function tekton_teardown() {
+  header "Removing Tekton Operator"
+  kubectl delete --ignore-not-found Config cluster
+  ko apply -n $TEST_NAMESPACE -f config/
+  kubectl delete -n $TEST_NAMESPACE -f deploy/crds/
```
nikhil-thomas: why do we need `-n $TEST_NAMESPACE` while applying CRDs as CRDs are clusterwide

test/e2e-tests.sh
```diff
@@ -14,15 +14,35 @@
...
+  kubectl create namespace $TEST_NAMESPACE
+  kubectl apply -n $TEST_NAMESPACE -f deploy/crds/
+  ko apply -n $TEST_NAMESPACE -f config/
+  wait_until_pods_running $TEST_NAMESPACE
+}
+
+function tekton_teardown() {
+  header "Removing Tekton Operator"
+  kubectl delete --ignore-not-found Config cluster
+  ko apply -n $TEST_NAMESPACE -f config/
```
nikhil-thomas: is this necessary in tear_down

test/e2e-tests.sh
nikhil-thomas: or should it be ` kubectl apply -n $TEST_NAMESPACE -f deploy/crs/`

nikhil-thomas:
@houshengbo 
as you said you might be going to fix it in a different patch.
However, `./test/e2e-tests.sh`
results in
```
error: set $PROJECT_ID or use --gcp-project to select the GCP project where the tests are run
```

houshengbo:
@nikhil-thomas per ./test/e2e-tests.sh, this is an existing old issue with this script all across knative and tekton projects. We have to set an env var PROJECT_ID or --gcp-project to run the script, no matter we use google cloud or not.      I will get to it to see, if we can resolve it in a wide range.

test/e2e-tests.sh
houshengbo: Since I create the CRDs available under config/, I can use kubectl apply -n $TEST_NAMESPACE -f config/ to install everything, including the crds and the deployment. So I can drop this line.

test/e2e-tests.sh
houshengbo: Again, I have the link files of crds under config/, the crds will be removed if I run kubectl delete -n $TEST_NAMESPACE -f config/. I will change the tear_down as well.

test/e2e-tests.sh
houshengbo: This is my fault, it should be delete.

test/e2e-tests.sh
houshengbo: done

test/e2e-tests.sh
nikhil-thomas: > @nikhil-thomas per ./test/e2e-tests.sh, this is an existing old issue with this script all across knative and tekton projects. We have to set an env var PROJECT_ID or --gcp-project to run the script, no matter we use google cloud or not. I will get to it to see, if we can resolve it in a wide range.

it set the env var PROJECT_ID, but then i need `gcloud` binary. In addtion, I have to provide a valid PROJECT_ID not any random value. :(

is that how it works at present

nikhil-thomas:
> @nikhil-thomas per ./test/e2e-tests.sh, this is an existing old issue with this script all across knative and tekton projects. We have to set an env var PROJECT_ID or --gcp-project to run the script, no matter we use google cloud or not. I will get to it to see, if we can resolve it in a wide range.

it set the env var PROJECT_ID, but then i need gcloud binary. In addtion, I have to provide a valid PROJECT_ID not any random value. :(

is that how it works at present

test/e2e-tests.sh
nikhil-thomas: it  makes more sense now :+1:

houshengbo:
@nikhil-thomas It seems so at this moment. To run the complete script, we need to set up a google cloud account, and associate with the billing account. 

However, you can verify the script running, by the integration tests: https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_operator/21/pull-tekton-operator-integration-tests/1184851383740796930/. All the google clusters with credentials have been set up there, so the script can run the operator installation and the tests against a kube cluster in google cloud.

Issue #66, opened 2020-01-30
Tekton operator security improvements


davidkarlsen:
# Expected Behavior

1. operator creates tekton-pipelines-controller and tekton-pipelines-webhook deployments, but it should be possible to define a custom registry for the images (to use in air-gap envs)
2. securityContext for the deployments should be configurable, as of now they are `{}` - which means root, privileged, etc

# Actual Behavior

# Steps to Reproduce the Problem

1. install the standard tekton operator using current defined config

# Additional Info

vincent-pli:
@davidkarlsen 
So, these are two requirements.

For the first one: `custom registry`, I personal prefer the design in [knative/serving-operator](https://github.com/knative/serving-operator/blob/5897cd8d8238eebfbeed7305b3e86a462f716b35/pkg/apis/serving/v1alpha1/knativeserving_types.go#L85), they defined a map to replace related image url in the `yaml`s, also consider the `pullsecret`, I think we can follow the same design.
@vdemeester @houshengbo @sthaha @nikhil-thomas

sthaha:
I agree, it will nice to use a config-map that we can use to `manifest.Transform` images that matches a regex to a destination image.

PR #68, opened 2020-02-13
Introduce new feature: image replacement

vincent-pli:
# Changes
Fix issue mentioned in #66 
Introduce new feature: `image replacement`

Enable `image replacement` for both `Config` and `Addon`.
A sample for `Config`:
```
kind: Config
metadata:
  name: cluster
spec:
  targetNamespace: tekton-pipeliness
  registry:
    override:
      tekton-pipelines-controller: index.docker.io/vincentpli/haha:v0.0.1
      -kubeconfig-writer-image: index.docker.io/vincentpli/kubeconfigwriter:v0.10.0
```
Almost same in `Addon`.

In code lever, not hard code in `Reconcile`, introduce concept ` Extension` and implements the first build in `Extension`: `imagereplacement`.
The design is easy for extending new features, maybe we will have more `Extension` in future.

Still not consider `pullsecret`, will add later.

# Submitter Checklist

review them:

# Release Notes

```
Describe any user facing changes here, or delete this block.

Examples of user facing changes:
- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/pkg/apis/operator/v1alpha1/addon_types.go
+++ b/pkg/apis/operator/v1alpha1/addon_types.go
@@ -8,6 +8,12 @@ import (
 // +k8s:openapi-gen=true
 type AddonSpec struct {
 	Version string `json:"version"`
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	// ExtensionWapper containes attribute Extension used
+	// +optional
+	ExtensionWapper `json:",inline"`
 }
 
 // AddonStatus defines the observed state of Addon
@@ -58,4 +64,10 @@ type AddonList struct {
 
 func init() {
 	SchemeBuilder.Register(&Addon{}, &AddonList{})
+}
+
+func (addonSpec AddonSpec) ConvertExtensionwapper() *ExtensionWapper {
+	return &ExtensionWapper{
+		Registry: addonSpec.Registry,
+	}
 }
--- a/pkg/apis/operator/v1alpha1/config_types.go
+++ b/pkg/apis/operator/v1alpha1/config_types.go
@@ -9,6 +9,12 @@ import (
 type ConfigSpec struct {
 	// namespace where pipelines will be installed
 	TargetNamespace string `json:"targetNamespace"`
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	// ExtensionWapper containes attribute Extension used
+	// +optional
+	ExtensionWapper `json:",inline"`
 }
 
 // ConfigStatus defines the observed state of Config
@@ -77,4 +83,10 @@ type ConfigList struct {
 
 func init() {
 	SchemeBuilder.Register(&Config{}, &ConfigList{})
+}
+
+func (configSpec ConfigSpec) ConvertExtensionwapper() *ExtensionWapper {
+	return &ExtensionWapper{
+		Registry: configSpec.Registry,
+	}
 }
--- a/pkg/controller/common/extensions.go
+++ b/pkg/controller/common/extensions.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package common
+
+import (
+	mf "github.com/jcrossley3/manifestival"
+	v1alpha1 "github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("common")
+
+// Activity is a interface, func Configure will be implemented by each Extension as a register. It will decide if
+// the Extension will be add to process chain.
+type Activity interface {
+	Configure(client.Client, *runtime.Scheme, *v1alpha1.ExtensionWapper) (*Extension, error)
+}
+
+// Activities is array of Activity
+type Activities []Activity
+
+// Extensions is array of Extension
+type Extensions []Extension
+
+// Extension is a plugin, could be remove/add easily
+type Extension struct {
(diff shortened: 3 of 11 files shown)


vincent-pli:
@nikhil-thomas @sthaha 
Could you take a look, thanks.

nikhil-thomas:
@vincent-pli thanks from bumping it up. This pr in my list and it was taking me for ever to reach it.
i shall review it right away.

nikhil-thomas:
@vincent-pli
This is very useful. I love it.

I have a suggestion. Could we define `Activity` as an interface,
something like 
```
type Activity interface {
    Configure(client.Client, *runtime.Scheme, *v1alpha1.ExtensionWapper) (*Extension, error)
}
```

Then each new extension that we add to `pkg/controller/extensions/` will have a uniform structure.

In addition, each new extension can be a type of its own which is customizable in each `add_imagereplacement.go `

something like,
in https://github.com/tektoncd/operator/pull/68/files#diff-fe73266e8939732fb98e4183e3ace7ec (pkg/controller/config/add_imagereplacement.go)
```
imgReplace := imagereplacement.New()  //returns something that implements Activity interface
activities = append(activities, imageReplace)
```

At present, the types `Extension`, `Activities` etc makes it a bit confusing to understand the flow. And each new extension's `Activity type func` can be named differently.

vincent-pli:
@nikhil-thomas 
Thanks for your comments, I will make some enhancement based on the comments, back here soon, thanks.

vincent-pli:
@nikhil-thomas 
Done some enhancement based on your comments, please take a look, thanks

pkg/controller/extensions/imagereplacement/imagereplacement.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// New creates a new Imagereplacement
+func New() (*Imagereplacement, error) {
+	return &Imagereplacement{}, nil
+}
+
+// Imagereplacement define Activity to replace image url in the yaml
+type Imagereplacement struct {
+	Extension        common.Extension
```
nikhil-thomas: can we use 
`[]mf.Transformer` instead of `common.Extension` here

pkg/controller/common/extensions.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return result, err
+		}
+		if ext != nil {
+			result = append(result, *ext)
+		}
+	}
+	return
+}
+
```
nikhil-thomas: Could you tell me why we need to break the logic here into `Extend` and `Transform` functions. 
Will it be better if we can return `transformers` directly from `Extend`.
```
func (activities Activities) Extend(c client.Client, scheme *runtime.Scheme, extensionWapper *v1alpha1.ExtensionWapper) (result []mf.Transformer, err error) {
.
.
.
if ext != nil {
			result = append(result, *ext.Transformers...)
		}
...
```
 If we do that we can avoid declaring the new type `Extension` and `Extensions` all together.

nikhil-thomas:
@vincent-pli Thank you for the updates.

One thing that is not clear to me is the need for the types `Extension` and `Extensions`
why can't we directly return `transformers` during the `Extend()` call.

vincent-pli:
@nikhil-thomas 
Thanks again for your comments.
It's my bad to make some confuse, I can explain it.
```
type Extension struct {
	Transformers []mf.Transformer
}
```
See the `Extension`, the `Transformers` is array of 
```
type Transformer func(u *unstructured.Unstructured) error
```
That's means the thing `Transformers` should do is modifying the `Unstructured` files (in our case, it's release.yaml), based on its function declaration.
but in my mind, it's not enough, maybe we need to check configuration of k8s before `transfer`, maybe modify k8s configuration after `transfer`, these are all not `Unstructured` work. 
So, we can extend `Extension` like this:
```
type Extension struct {
        preTransfer()
	Transformers []mf.Transformer
        postTransfer()
}
```
That will make it good organized, so keep `Extension` is for extensibility in the future. 
Need your comments, thanks

pkg/apis/operator/v1alpha1/addon_types.go
```diff
@@ -8,6 +8,12 @@ import (
 // +k8s:openapi-gen=true
 type AddonSpec struct {
 	Version string `json:"version"`
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	// ExtensionWapper containes attribute Extension used
+	// +optional
+	ExtensionWapper `json:",inline"`
```
sthaha: nit: typo? what does wapper mean?

pkg/apis/operator/v1alpha1/addon_types.go
```diff
@@ -59,3 +65,9 @@ type AddonList struct {
 func init() {
 	SchemeBuilder.Register(&Addon{}, &AddonList{})
 }
+
+func (addonSpec AddonSpec) ConvertExtensionwapper() *ExtensionWapper {
```
sthaha: may be `ConvertExtensionWapper` to match `ExtensionWapper` ?

pkg/apis/operator/v1alpha1/config_types.go
```diff
@@ -9,6 +9,12 @@ import (
 type ConfigSpec struct {
 	// namespace where pipelines will be installed
 	TargetNamespace string `json:"targetNamespace"`
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
sthaha: I am not sure if we need to add // INSERT ADDITIONAL ...

pkg/controller/config/config_controller.go
```diff
@@ -199,25 +200,28 @@ func (r *ReconcileConfig) reconcileInstall(req reconcile.Request, res *op.Config
...
+	extensions, err := activities.Extend(r.client, r.scheme, extensionWrapper)
+	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	if err := r.manifest.ApplyAll(); err != nil {
-		log.Error(err, "failed to apply release.yaml")
+	tfs = append(tfs, extensions.Transform()...)
+	err = r.manifest.Transform(tfs...)
+	if err == nil {
```
sthaha: could you please convert 

```
if err == nil {
 green path... 
}
if err != nil {
  ....
 return 
}
```
to 

```
if err != nil {
   ....
   return 
}
green path ...
```

pkg/apis/operator/v1alpha1/addon_types.go
```diff
@@ -8,6 +8,12 @@ import (
 // +k8s:openapi-gen=true
 type AddonSpec struct {
 	Version string `json:"version"`
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
```
sthaha: curious if you ran it ? shouldn't it generate the deepcopy?

sthaha:
@vincent-pli @nikhil-thomas I am not sure if I grok the design yet. Please allow me a day or so to wrap my head around it. My knee jerk reaction is if it is really needed but that could be because I fail to see the advantage the Extension and Activities bring over a simple func/method of `Extension` that returns a list of `mf.Transformations` .

vincent-pli:
@sthaha 
Sure, I'm glad to wait for your further comments

pkg/apis/operator/v1alpha1/addon_types.go
vincent-pli: Should be Wrapper I guess 😓

pkg/controller/config/config_controller.go
vincent-pli: Sure, thanks

PR #79, opened 2020-04-14
Upgrade operator-sdk to 0.11.0.

timmyers:
# Changes

- Upgrade the operator-sdk to 0.11.0 following https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md.
- This will fix the issue reference in this comment:
```
// NOTE: need to use a deepcopy since Status().Update() seems to reset the
// APIVersion of the res to "" making the object invalid; may be a mechanism
// to prevent us from using stale version of the object
```
and ultimately allow fixing a major bug related to expecting the cached reader to give up to date results.

# Submitter Checklist

review them:


timmyers:
These diffs are currently very large due to the `vendor` directory, what is the reason for it being in github?

vdemeester:
> These diffs are currently very large due to the `vendor` directory, what is the reason for it being in github?

Mainly to be able to compile using older go version (if needed / if we can), and also because it reduces the time to build, etc.. (aka we don't download the deps on each and every CI run, …).
It not a huge deal for reviews on GitHub as it defaults to hide the vendor folder :wink:

sthaha:
@timmyers thank you for the patch. Could you please address the integration-test failure? 🙏

pkg/apis/operator/v1alpha1/addon_types.go
```diff
@@ -38,7 +38,7 @@ type AddonCondition struct {
 
 // Addon is the Schema for the addons API
 // +k8s:openapi-gen=true
-// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=addons,scope=Namespaced
```
sthaha, line 41: aren't the addons cluster-scoped?

timmyers: Sorry this should be reverted, it was an attempt to fix the bug that did not do what I expected.

timmyers:
> Could you please refactor these into 2 PR 
Sure, I will update this PR to just bump the sdk version since its a prerequisite for fixing the bug.

PR #81, opened 2020-04-17
fix: major bugs in addon_controller.go

timmyers:
# Changes

Code was relying on refreshCR function to update the CR during reconciliation. This does not work becaues the operator SDk client uses a cache so this is almost always out of date, resulting in tons of errors as we resources are updated with stale resource versions. Remove all use of that and instead rely on the fact that client Update() calls will update the CR pointer correctly. Updating operator-sdk and deps in #79 fixed the bug that was mentioned in a comment where GroupVersionKind is lost.

# Submitter Checklist

review them:

### DIFF
--- a/pkg/controller/addon/addon_controller.go
+++ b/pkg/controller/addon/addon_controller.go
@@ -136,13 +136,11 @@ func (r *ReconcileAddon) ensureAddonVersion(res *op.Addon) (bool, error) {
 		return false, err
 	}
 
-	tmpRes := res.DeepCopy()
-	tmpRes.Spec.Version = version
-	err = r.client.Update(context.TODO(), tmpRes)
+	res.Spec.Version = version
+	err = r.client.Update(context.TODO(), res)
 	if err != nil {
 		return false, err
 	}
-	r.refreshCR(res)
 
 	return true, nil
 }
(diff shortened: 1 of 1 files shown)


timmyers:
There shouldn't be an update with more than one `update` in a reconcile loop.  The `update` function modifies the CR object that is passed to it.

From the `controller-runtime` package:
```
// Update updates the given obj in the Kubernetes cluster. obj must be a
// struct pointer so that obj can be updated with the content returned by the Server.
```

vincent-pli:
@timmyers 
Yes, but when run `update`, the `resource version` will changed, if more than one `update` in once `reconcile` the subsequent `update` will hit `version` issue, since it based on a stale resource version(since in one `reconcile`).
Correct me if I missing something, thanks

timmyers:
`r.client.Update` will update the resource version on the CR so you can use it multiple times in a row i.e.:

```
func (r *ReconcileAddon) test(res *op.Addon) {
	fmt.Println(res.ResourceVersion) // 1
	r.client.Update(context.TODO(), res) // Updates res.ResourceVersion if successful
	fmt.Println(res.ResourceVersion) // 2
}
```

vincent-pli:
@timmyers 
I think you are right, maybe that different between pure `client-go` and `controller-runtime`, thanks.

/lgtm
/approve

nikhil-thomas:
cc @sthaha

sthaha:
/lgtm 
/meow 
❤️

sthaha:
@timmyers we need the CLA to be signed 🙏

timmyers:
Trying to figure out how to sign the new CLA...

timmyers:
@sthaha I had to push a whitespace comment change to trigger the CLA bot, will need a new LGTM 🙏

PR #114, opened 2020-07-13
WIP: Add ko command support to build the image

houshengbo:
# Changes

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/operator/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

PR #121, opened 2020-07-17
Add ko based workflow

nikhil-thomas:
# Changes

- Add `ko apply -f config` support.
  This will help us reuse already existing Release Pipelines and Tasks used by other Tekton Projects

- Update tests and Make targets to accommodate the change

- Make ko workflow work with current test mechanism
    
- Add `--no-setup` flag in `operator-sdk test` command, so that we can setup the test using the updated paths (deploy->config)
    
- Update Makefile targets
    
- Add an improved WaitforCR function in e2e test suite
    
- Add a `.ko.yaml`

**Note:** This patch **cherry-picks  5 commits from #114** and make it work with current test setup.

# Submitter Checklist

review them:

# Release Notes

    ```release-note
    NONE
    ```


.ko.yaml
```diff
@@ -0,0 +1 @@
+defaultBaseImage: registry.access.redhat.com/ubi8/ubi-minimal:latest
```
sthaha, line 1: should we use a different image here 🙏 ?

nikhil-thomas:
> looks good to me but we may want to switch the base image to a different one so that additional changes can be made easily.

that makes sense,

shall we use `gcr.io/distroless/static:latest` 

or is it recommended to use anyother image from this pipeline/.ko.yaml`https://github.com/tektoncd/pipeline/blob/master/.ko.yaml

@sthaha   @vdemeester @houshengbo

.ko.yaml
nikhil-thomas: done :+1:

.ko.yaml
```diff
@@ -0,0 +1 @@
+defaultBaseImage: gcr.io/distroless/static:latest
```
vdemeester, line 1: Can we try `gcr.io/distroless/static:nonroot` just to see if it would work too ? :upside_down_face:

Makefile
```diff
@@ -1,35 +1,46 @@
...
 .PHONY: clean
 clean:
-	-kubectl delete -f deploy/
-	-kubectl delete -f deploy/crds/
-	-kubectl delete namespace tekton-pipelines
-
-.PHONY: local-dev
-local-dev: dev-setup
-	GO111MODULE=on operator-sdk run --local --watch-namespace "" --operator-flags '--zap-encoder=console'
+	-kubectl delete -f config/
```
vdemeester, line 8: could be `ko` too here :stuck_out_tongue_closed_eyes:

.ko.yaml
nikhil-thomas: done

Makefile
nikhil-thomas: done

Issue #133, opened 2020-08-13
TektonPipeline Resource status.conditions always appended


bkmeneguello:
# Expected Behavior

I expect only relevant conditions to be appended to the list

# Actual Behavior

There are multiple copies of the same status appended:

```yaml
...
status:
  conditions:
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
  - code: installed
    version: v0.15.1
  - code: installing
    version: v0.15.1
```

# Steps to Reproduce the Problem

1. Just install the operator
2. Check the resource

# Additional Info

There is no check before append at [pipeline_controller.go#L325](https://github.com/tektoncd/operator/blob/cf1f2174a092cc3d76a42ecead30ef600b553d1e/pkg/controller/pipeline/pipeline_controller.go#L325)

PR #146, opened 2020-09-09
Avoid multiple useless condition in cr: tektonpipeline

vincent-pli:
# Changes

Fix issue: #133 

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/controller/pipeline/pipeline_controller.go
+++ b/pkg/controller/pipeline/pipeline_controller.go
@@ -322,6 +322,12 @@ func (r *ReconcileTektonPipeline) updateStatus(res *op.TektonPipeline, c op.Tekt
 	// to prevent us from using stale version of the object
 
 	tmp := res.DeepCopy()
+	for _, condition := range tmp.Status.Conditions {
+		if condition.Code == c.Code && condition.Version == c.Version {
+			return nil
+		}
+	}
+
 	tmp.Status.Conditions = append([]op.TektonPipelineCondition{c}, tmp.Status.Conditions...)
 
 	if err := r.client.Status().Update(context.TODO(), tmp); err != nil {


pkg/controller/pipeline/pipeline_controller.go
```diff
@@ -322,6 +322,12 @@ func (r *ReconcileTektonPipeline) updateStatus(res *op.TektonPipeline, c op.Tekt
 	// to prevent us from using stale version of the object
 
 	tmp := res.DeepCopy()
+	for _, condition := range tmp.Status.Conditions {
+		if condition.Code == c.Code && condition.Version == c.Version {
+			return nil
+		}
+	}
+
```
nikhil-thomas, line 330: could you make this check `tmp.Status.Conditions[0].code == c.Code && ...`

otherwise, the loop will match the condition from the past, and skip a relevant update

nikhil-thomas (changes requested):

pkg/controller/pipeline/pipeline_controller.go
vincent-pli: @nikhil-thomas  if we only check the [0], it will not work, since when reconcile is triggered, the new coming `condition` will always: 
`installing`
then
`installed`

nikhil-thomas: @vincent-pli oh, i see.. the status oscillates between `installing` and `install`. 
i think we can handle this better in the reconcile function it self.

could you move the check in this patch to the reconcile function (right before the `reconcileInstall` call).

ref: https://github.com/tektoncd/operator/blob/98cbaf7ed72855750c9d2b877f75515cc11d92f8/pkg/controller/pipeline/pipeline_controller.go#L193

nikhil-thomas: so `reconcileInstall` will not be called if the status is `installed` (the oscillating status updates won't be there)

nikhil-thomas:
/hold
https://github.com/tektoncd/operator/pull/146#discussion_r486806382

pkg/controller/pipeline/pipeline_controller.go
vincent-pli: @nikhil-thomas Thanks for your reply.
I guess we should not block the `reconcileInstall`, besides the first installation, there is two kinds of events will trigger `reconcileInstall`.
- Any change on `TektonPipeline` and related `Deployment`, since we watch on them:
https://github.com/tektoncd/operator/blob/98cbaf7ed72855750c9d2b877f75515cc11d92f8/pkg/controller/pipeline/pipeline_controller.go#L112-L126
`Operator` try to handle the modification no matter it's on purpose or not.

- Regular event from apiserver (not sure, correct me if incorrect)

nikhil-thomas: i see, it makes sense.
:+1:

PR #184, opened 2020-12-08
Propagate proxy configuration to underlying deployments

vdemeester:
# Changes

This propagate any proxy configuration set on the operator
deployment (through environment variables). This applies to any
deployment managed by the different addons (TektonPipeline, …).

/kind feature
/hold

/cc @nikhil-thomas @piyush-garg @savitaashture 

# Submitter Checklist

review them:

# Release Notes

```release-note
Propagate proxy configuration that are set on the operator to underlying deployments
```


vdemeester:
On hold because it needs a proper test / e2e test somehow :upside_down_face:

pkg/reconciler/kubernetes/tektonpipeline/controller.go
```diff
@@ -57,6 +57,7 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
 			logger.Fatalw("Error creating client from injected config", zap.Error(err))
 		}
 		mflogger := zapr.NewLogger(logger.Named("manifestival").Desugar())
+		// FIXME(vdemeester) why ignoring error ?
 		manifest, _ := mf.ManifestFrom(mf.Slice{}, mf.UseClient(mfclient), mf.UseLogger(mflogger))
```
nikhil-thomas, line 61: no reason. it should atleast be logged.
I shall take care of it in a different patch in all NewExtendedController Functions. :+1:

nikhil-thomas:
@savitaashture

pkg/reconciler/common/proxy.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		for _, e := range proxyEnv {
+			if e.Value == "" {
+				continue
+			}
+			envvar := map[string]interface{}{
+				"name":  e.Name,
+				"value": e.Value,
+			}
+			currentEnv = append(currentEnv, envvar)
```
piyush-garg, line 68: Considering the scenario if there is an env already with the same name, there will two env then

pkg/reconciler/common/proxy_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	actual := unstructuredDeployment(t)
+	expected := unstructuredDeployment(t, withEnv(proxyEnv))
+
+	defer env.PatchAll(t, proxyEnv)()
+	if err := ApplyProxySettings(actual); err != nil {
+		t.Fatal(err)
+	}
+
+	assert.DeepEqual(t, actual, expected)
```
savitaashture, line 57: Bit confused with testcase here
As we have not set env in `actual` how are we getting  `actual` and `expected` as equal ?

pkg/reconciler/common/proxy.go
vdemeester: Indeed. There is also the case where the env are there and we want to remove them. Will update 😉

pkg/reconciler/common/proxy_test.go
vdemeester: Because `ApplyProxySettings` does its job 😉 — aka ready the proxy env variable (set by `env.PatchAll`) and add them to the list of environment variables.

vdemeester:
Note: custom ca support will be a follow up 😉

vdemeester:
Oh the error is legit… `panic: interface conversion: interface {} is nil, not string` 😅

PR #221, opened 2021-01-28
Restructure proxy installation

piyush-garg:
Currently proxy webhook is installed in operator namespace as part of
operator. this refactoring will make it to install as a component like
pipeline and triggers in the tekton namespace and installation will be
done along with pipeline component.

This fixes the issue getting faced as mutatuingwebhookcnfiguration is not
supported in CSV if name is specified.

# Submitter Checklist

review them:

# Release Notes

```release-note
Restructure proxy installation as a component
```


pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -105,7 +106,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
 		return err
 	}
 	stages := common.Stages{
-		common.AppendTarget,
+		r.appendPipelineTarget,
```
nikhil-thomas, line 109: shall we keep these 2 as separate stages. ie, leave common.AppendTarget as it is. Then make r.appendPipelineTarget next stage and make it handle only what `addProxy` is doing.

In addition, after separating we can also rename the `r.appendPipelineTarget` to something like `r.appendProxy`

savitaashture (commented):
/lgtm

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
piyush-garg: I will work on this in a follow-up PR. Thanks

PR #312, opened 2021-05-24
Moves rbac controller logic to OpenShift extention

sm43:
This moves the rbac logic to OpenShift extension in TektonConfig
controller and adds TektonConfig as owner to all rbac resources.
So that when TektonConfig instance is deleted, all rbac resources
get deleted.

The TektonConfig controller watches namespace, whenever a ns event
occurs it will enqueue event with name `config` in work queue to trier
the TektonConfig controller.
This is required because we want to get our TektonConfig reconciler triggered
for already existing and new namespaces, without manual intervention like adding
a label/annotation on namespace to make it manageable by Tekton controller.

# Submitter Checklist

review them:

# Release Notes


pkg/reconciler/kubernetes/tektonconfig/controller.go
```diff
@@ -88,3 +93,13 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
 		return impl
 	}
 }
+
+// customEnq adds `config` in work queue so that whenever a namespace
+// event occurs, the reconciler get triggered. The reconciler will only
+// get triggered only when a TektonConfig instance with name `config`
+// already exist.
```
nikhil-thomas, line 100: mention that this is required because we want to get our TektonConfig reconciler triggerd for already existing and new namespaces, without manualy intervention like `adding a label/annotion` on namespace to make it manageable by Tekton controllers. 

in addition could you take care of:
- `s/adds 'config' in work queue/adds an event with name `config` in work queue/`
- `s/the reconciler get triggered./the TektonConfig reconciler get triggered./`
- remove 
`The reconciler will only// get triggered only when a TektonConfig instance with name `config`// already exist.` because the reconciler will be triggered for TektonConfig CR instances with any name. Inside the reconciler we just have a check to skip reconcile for all instances of TektonConfig with anyother name other than `config`.

pkg/reconciler/kubernetes/tektonconfig/controller.go
```diff
@@ -88,3 +93,13 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
...
+func customEnq(impl *controller.Impl) func(obj interface{}) {
```
nikhil-thomas, line 101: change this name to something descriptive, something like `enqueueName(impl *controller.Impl, name string)`.
so that we can pass the `common.ConfigResourceName`

nikhil-thomas:
@sm43 could you delete the `pkg/reconciler/openshift/tektonconfig/extension/cleanup-clusterrole-rolebinding.go`. This is obsolete and can be removed safely.

nikhil-thomas:
cc @concaf

pkg/reconciler/kubernetes/tektonconfig/controller.go
```diff
@@ -80,6 +83,8 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
 			Handler:    controller.HandleAll(impl.EnqueueControllerOf),
 		})
 
+		namespaceInformer.Informer().AddEventHandler(controller.HandleAll(enqueueCustomName(impl, common.ConfigResourceName)))
```
pradeepitm12, line 86: is this required for kubernetes aswell?

pradeepitm12 (commented):
/lgtm

Issue #311, opened 2021-05-24
Components customization


pvlltvk:
Hi, everyone!

Is there any possibility to set/customize some parameters of Tekton components which are installed by Operator? 
For example, I want to set some feature-flags as described [here](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior)

nikhil-thomas:
At present, we don't expose any of the parameter through our Operator's CRDs.

However, potential workaround is to edit them in the config map of respective components. There is a gotcha though, operator might reset custom changes in successive reconciles.

The issue https://github.com/tektoncd/operator/issues/149 needs to be resolved for the above workaround to work reliably.

Customization options are something we are planning. However, the UX is not defined yet (change it in CR spec or add a new ConfigMap etc).

@pvlltvk how would you prefer to set those customizations if you can pick any ux.

pvlltvk:
Hi @nikhil-thomas!

I would definitely prefer to set it in CR spec.

nikhil-thomas:
How about an updated spec like:
```
apiVersion: operator.tekton.dev/v1alpha1
kind: TektonConfig
Metadata:
name: config
Spec:
  Profile: all
  targetNamespace: tekton-pipelines
  pipelineConfig:
    prunePTRuns: 
    frequency: int # will take the frequency of pruning
    ptRunsToKeep: int # will take max runs to keep
    ptAge: int  # max age allowed of runs
    relatedResource: bool # want to prune related res
    default-param1: #items from https://github.com/tektoncd/pipeline/blob/main/config/config-defaults.yaml
    default-param2:
    ...
    feature-flag1: #items from https://github.com/tektoncd/pipeline/blob/main/config/config-feature-flags.yaml
    feature-flag2:
    ...
  triggersConfig: {}
  dasboardConfig:
    param1:
    param2:
  ```

@pvlltvk 
cc @houshengbo @vincent-pli @vdemeester @savitaashture @pradeepitm12 @sm43

pvlltvk:
@nikhil-thomas Looks good to me!

PR #314, opened 2021-05-31
Provides fields in CRs to customize Pipelines Configmaps

### LINKED ISSUE #149: Tekton ConfigMaps are overwritten by Operator
# Expected Behavior

After installing the Operator and make any modification to Tekton ConfigMaps (config-defaults, feature-flags) they should not be modified, even after operator restarts or upgrades.

# Actual Behavior

If I restart the operator deployment container the config-maps are overwritten with default values, replacing anything I've defined.

- Install Tekton Operator
- Wait for everything to be provisioned
- Modify any tekton-pipelines/feature-flags config map entry
- Scale-down tekton operator deployment to zero
- Scale-up tekton operator deployment to one
- Check the modified tekton-pipelines/feature-flags config map is overwritten

sm43:
This provides options to customize pipelines configmaps through TektonConfig Spec.
Field will be passed from TektonConfig to TektonPipelines which will be replaced in Configmaps while installing them.
Some of the fields have default values so if user doesn't pass them, operator will add them to CR. Fields which
doesn't have default values would be not set till user pass some values.
Documenation can be found at https://github.com/tektoncd/operator/blob/main/docs/TektonConfig.md

Closes #149 

# Submitter Checklist

review them:

# Release Notes

```
User can now customize pipelines's configmap fields through TektonConfig Spec. Documentation for available fields can be found at https://github.com/tektoncd/operator/blob/main/docs/TektonConfig.md
```


pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -134,14 +140,28 @@ func addProxy(manifest *mf.Manifest) error {
...
+	if err != nil {
+		logger.Errorf("failed to fetch configmaps from namespace %s", comp.GetSpec().GetTargetNamespace())
+		return err
+	}
+
+	// If configMaps exists, then preserve the configurations changes if any
+	if len(cms) != 0 {
+		extra = append(extra, common.PreserveExistingConfigMapsConfigurations(cms))
+	}
+
```
nikhil-thomas, line 164: could you move all of this logic into transformers, or to a separate file where transformers are located.
and just add `common.PreserveExistingConfigMapsConfigurations(cmNames)`

nikhil-thomas:
@sm43 looks good overall, could you move all the implementaiton details into the transformer itself and keep only the transformer addition to the transformers list in the main reconciler.

nikhil-thomas:
looks good 👍 

cc @savitaashture @pradeepitm12 @vdemeester

nikhil-thomas:
cc @vdemeester

vdemeester (commented):
I am not sure I fully follow this change though. `app.kubernetes.io/part-of: tekton-pipelines` is on the configmap we install. This changes means once we install a version, the configmap will never change or be updated right ? Or at least current values wouldn't change ever — even if the default upstream changed (e.g. the `HOME`/`WORKDIR` change in pipeline 0.24).
I am not sure that's exactly what we want though. We should allow the user to "set in stone" there configmap (aka saying the operator "it's on me, do not try to manage it") but default would be to manage everything (including the configmap). Does that make sense ?

nikhil-thomas:
> I am not sure I fully follow this change though. `app.kubernetes.io/part-of: tekton-pipelines` is on the configmap we install. This changes means once we install a version, the configmap will never change or be updated right ? Or at least current values wouldn't change ever — even if the default upstream changed (e.g. the `HOME`/`WORKDIR` change in pipeline 0.24).
> I am not sure that's exactly what we want though. We should allow the user to "set in stone" there configmap (aka saying the operator "it's on me, do not try to manage it") but default would be to manage everything (including the configmap). Does that make sense ?

yup, it does. 👍 

how about this. 
We can consider this as a fix for current situation. ie, operator always overwrites changes in Component Config.

Then, we can take care of scenario in your comment as part of Component Customization story (https://github.com/tektoncd/operator/issues/311). We can decide a mechanism to expose customization to user (CR spec, configMap etc)

My preference is to let this merge and handle the rest as an improvement in the near future. 🙂

vdemeester:
> yup, it does. 
> 
> how about this.
> We can consider this as a fix for current situation. ie, operator always overwrites changes in Component Config.
> 
> Then, we can take care of scenario in your comment as part of Component Customization story (#311). We can decide a mechanism to expose customization to user (CR spec, configMap etc)
> 
> My preference is to let this merge and handle the rest as an improvement in the near future.

My worry doing that is that it would create a case where we cannot upgrade values from the configmap on an upgrade. If we just take the approach of using another annotation (that is not here by default) to handle this behavior, we make it explicit instead of implicit. And I'd definitely would enabling this behavior implicitely, from the start.

GijsvanDulmen:
In our case we would definitely always want customizations on the configmap-defaults. So that would probably mean on every upgrade we need to look into the impact from a future version on the currently installed configmap. We would expect the operator to help in the process of updates (but could live without as long as it's explicily mentioned somewhere).

nikhil-thomas:
@sm43 the functionality looks good.

Could you make the TektonPipelie Reconciler watch for configMap changes as well. So that if a user deletes the label, a reconcile shall be triggered and the defaults restored.

cc @vdemeester

sm43:
> @sm43 the functionality looks good.
> 
> Could you make the TektonPipelie Reconciler watch for configMap changes as well. So that if a user deletes the label, a reconcile shall be triggered and the defaults restored.
> 
> cc @vdemeester

@nikhil-thomas  Updated with changes. ptal :)

nikhil-thomas:
/hold
hi, @sm43 please excuse the hold. 😇 
 I want to make sure that this won't affect our attempts for seamless upgrades in the long run. 🧑‍💻 

The transformer implementation is good. However, shall we use a separate source for user customizations. For example, a configMap in `tekon-pipelines` namespace, which is not managed by operator, where user can specify customizations.

somethinglike

```
apiVersion: v1
kind: ConfigMap
label:
  managedBy: tektonconfig or tektonpipeline (for the knative pkg enquee mechanism)
data:
  config-defaults_default-service-account: 'my-sa'
  config-defaults_default-timeout-minute: 90
  feature-flags_disable-affinity-assistant: "false"
  feature-flags_disable-creds-init: "true"
```
@vdemeester @savitaashture @GijsvanDulmen

GijsvanDulmen:
> /hold
> hi, @sm43 please excuse the hold. 😇
> I want to make sure that this won't affect our attempts for seamless upgrades in the long run. 🧑‍💻
> 
> The transformer implementation is good. However, shall we use a separate source for user customizations. For example, a configMap in `tekon-pipelines` namespace, which is not managed by operator, where user can specify customizations.
> 
> somethinglike
> 
> ```
> apiVersion: v1
> kind: ConfigMap
> label:
>   managedBy: tektonconfig or tektonpipeline (for the knative pkg enquee mechanism)
> data:
>   config-defaults_default-service-account: 'my-sa'
>   config-defaults_default-timeout-minute: 90
>   feature-flags_disable-affinity-assistant: "false"
>   feature-flags_disable-creds-init: "true"
> ```
> 
> @vdemeester @savitaashture @GijsvanDulmen

From a user perspective I wouldn't mind. Sounds like the easiest solution during an upgrade. In our case this configuration would be in Git and synced with ArgoCD as part of a "cluster helm chart". But I guess a lot of operators make this config part of the CRD's. So the configuration could be part of the "TektonPipeline" for example (a seperate config property in there or such).

vdemeester:
> /hold
> hi, @sm43 please excuse the hold. 
> I want to make sure that this won't affect our attempts for seamless upgrades in the long run. 
> 
> The transformer implementation is good. However, shall we use a separate source for user customizations. For example, a configMap in `tekon-pipelines` namespace, which is not managed by operator, where user can specify customizations.

@nikhil-thomas How would it affect seamless upgrade in the long run ? (naive question I know 😬 )

nikhil-thomas:
> @nikhil-thomas How would it affect seamless upgrade in the long run ? (naive question I know grimacing ) 

if we make one of the resources, created/managed by ther oeprator, behave differently with an annotation (giving control to user). Then the operator will not be free to delete/recreate  those resources, as it will have to do additional checks to read user provided specs, store them, peform updates and update them.

So i think, making one of the managed resource like config-defaults config map the source of user provided config, makes reconcile complex. Instead, it would be better if we can make our users specify configguration through a configMap or CR spec, which is untocuhed by operatorReconciles.

vdemeester:
>  Instead, it would be better if we can make our users specify configguration through a configMap or CR spec, which is untocuhed by operatorReconciles.

Yeah, for me in the spec of the CR is the way to go 🙃

nikhil-thomas:
> > Instead, it would be better if we can make our users specify configguration through a configMap or CR spec, which is untocuhed by operatorReconciles.
> 
> Yeah, for me in the spec of the CR is the way to go 🙃

@sm43 ☝️ 🙂

sm43:
> > > Instead, it would be better if we can make our users specify configguration through a configMap or CR spec, which is untocuhed by operatorReconciles.
> > 
> > 
> > Yeah, for me in the spec of the CR is the way to go upside_down_face
> 
> @sm43 point_up slightly_smiling_face

yep. noted.
I will be updating this pr to pass the configuration through spec.
so the cr would look like
```
kind: TektonConfig
metadata:
  name: config
spec:
  pipeline:
    enable-custom-tasks: "false"
    enable-tekton-oci-bundles: "false"
```

sm43:
this is on top of #330 :)

adjavaherian:
Not sure if I'm seeing the same problem here, but basically after installing Tekton-pipelines via operator, I've applied a custom feature-flags.yaml  and it only sticks for a few hours before something sets it back to the defaults.  Not sure if this is done by Operator?

```
kubectl get configmap feature-flags -n tekton-pipelines -o yaml         
apiVersion: v1
data:
  disable-affinity-assistant: "true"
  disable-creds-init: "false"
...
```

Basically, I'm applying a brand new ConfigMap with disable-affinity-assistant=true and after a few hours I can pull the same and see its set back to "false".

Would this be from Operator and related to above?

vdemeester:
> Not sure if I'm seeing the same problem here, but basically after installing Tekton-pipelines via operator, I've applied a custom feature-flags.yaml and it only sticks for a few hours before something sets it back to the defaults. Not sure if this is done by Operator?
> 
> ```
> kubectl get configmap feature-flags -n tekton-pipelines -o yaml         
> apiVersion: v1
> data:
>   disable-affinity-assistant: "true"
>   disable-creds-init: "false"
> ...
> ```
> 
> Basically, I'm applying a brand new ConfigMap with disable-affinity-assistant=true and after a few hours I can pull the same and see its set back to "false".
> 
> Would this be from Operator and related to above?

Yes that's the current behavior of the Operator, and that's one of the reason of this PR, to allow users to overwrite configuration on components instead of.. overriding it silently 😓

adjavaherian:
Thanks @vdemeester, would a workaround for now, be to disable Operator (after the pipeline install), by simply scaling down the Operator deployment until next time we want to upgrade?

vdemeester:
> Thanks @vdemeester, would a workaround for now, be to disable Operator (after the pipeline install), by simply scaling down the Operator deployment until next time we want to upgrade?

That could work yes. If running the operator through OLM not sure if that works, but running it alone, yes that seems like a *decent* workaround until we fix it.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -134,10 +140,12 @@ func addProxy(manifest *mf.Manifest) error {
 // transform mutates the passed manifest to one with common, component
 // and platform transformations applied
 func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp v1alpha1.TektonComponent) error {
-	//logger := logging.FromContext(ctx)
+	pipeline := comp.(*v1alpha1.TektonPipeline)
 	images := common.ToLowerCaseKeys(common.ImagesFromEnv(common.PipelinesImagePrefix))
 	instance := comp.(*v1alpha1.TektonPipeline)
 	extra := []mf.Transformer{
+		common.ReplaceConfigMapValues(featureFlag, pipeline.Spec.PipelineProperties),
+		common.AddOptionalConfigMapValues(configDefaults, pipeline.Spec.OptionalPipelineProperties),
```
nikhil-thomas, line 148: hi, could you tell us why we need 2 different transformers here.
From my prespective, at this point, all values can be consirered optional. That is if it is not set in the CR the default from the release.yaml (payload) will be already there in each of the 2 configMaps. so why don't we call `ReplaceConfigMapValues` 2 times but with 2 different configMap names

sm43: for `config-defaults` cm there are not fields defined in configmap, all are commented so we loop n the `OptionalPipelineProperties` and if there is value defined for a field we add it in the configmap

for the other one, we loop on the config map fields and copy the value from CR, 
we set the default values in webhook and then copy them in transformer without any check.

sm43: but now I think we can use `AddOptionalConfigMapValues` for both configmaps where we loop on struct fields and add in configmap

sm43:
@vdemeester @nikhil-thomas need a final review 😅

config/base/300-operator_v1alpha1_addon_crd.yaml
```diff
@@ -48,83 +48,4 @@ spec:
       openAPIV3Schema:
         type: object
         description: Schema for the tektonaddons API
-        properties:
```
vdemeester, line 51: Any reason for removing this schema ?

vdemeester: (Same for the other CRDs)

sm43: if we want to add schema then we would have to add for all field and it is used for of validation of fields too.  if we keep `x-kubernetes-preserve-unknown-fields: true` allowing all fields from here and webhook would validate and set default similar to pipelines seems to be better.

sm43: yeah this does means kubectl explain would not show any info 😄
 should we keep the schema or have docs take care of field explanation?

vdemeester: @sm43 ideally we would make this auto-generated (it should be possible). Also, those schema might be useful for OpenShift (for auto-creating forms from the CRD when you want to create a new one).

sm43: @vdemeester currently there isn't a supported way to auto generate crds in knative/pkg I think.
we do it manually, I have created an issue for this https://github.com/tektoncd/operator/issues/352
will address it separately if that's okay ? 🙂

vdemeester: Thanks for creating the issue. We tried something on `tektoncd/pipeline` (with annotation and stuff), so it is definitely possible but on `tektoncd/pipeline` we are running out of place for them, here it sohuld work fine, I'll try to link the experiment to the issue 😉

sm43:
/hold cancel 
😄

PR #321, opened 2021-06-07
Update rbac changes to support triggers clusterinterceptors

savitaashture:
# Changes
Modified `edit` clusterrole to bind with `clustertriggerbinding` instead of `rolebinding` in order to satisfy triggers new resource `clusterinterceptors`
[rbac example from triggers repo](https://github.com/tektoncd/triggers/blob/main/examples/rbac.yaml#L54)
[Eventlistener Doc](https://github.com/tektoncd/triggers/blob/main/docs/eventlisteners.md#specifying-the-kubernetes-service-account)

**Issue:**
EL pod crashes with below error
```
E0607 13:14:29.078049       1 reflector.go:127] runtime/asm_amd64.s:1371: Failed to watch *v1alpha1.ClusterInterceptor: failed to list *v1alpha1.ClusterInterceptor: clusterinterceptors.triggers.tekton.dev is forbidden: User "system:serviceaccount:default:pipeline" cannot list resource "clusterinterceptors" in API group "triggers.tekton.dev" at the cluster scope
```

/cc @nikhil-thomas @khrm @vdemeester 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/operator/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```release-note
From triggers 0.13.0 a new CRD clusterinterceptors introduced and which requires cluster level permission.
In order to give permission at cluster level created new clusterrole and clusterrolebinding for clusterinterceptors
```

PR #340, opened 2021-07-01
Append OwnerReference if already exists

### LINKED ISSUE #329: Always add ownerReference of parent to the resource
Creating issue to solve https://github.com/tektoncd/operator/pull/321#discussion_r649951551

/cc @nikhil-thomas

vinamra28:
# Changes

As of now we were adding our ownerReference if the existing
ownerReference length is equal to zero.
With this patch we will check if the existing ownerReference is same as
that of ours, if not then append our ownerReference.
See https://github.com/tektoncd/operator/pull/321#discussion_r649951551

closes #329 

Signed-off-by: vinamra28 <vinjain@redhat.com>
# Submitter Checklist

review them:

# Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/openshift/tektonconfig/rbac.go
+++ b/pkg/reconciler/openshift/tektonconfig/rbac.go
@@ -330,10 +330,8 @@ func (r *rbac) updateRoleBinding(ctx context.Context, rb *rbacv1.RoleBinding, sa
 		rb.Subjects = append(rb.Subjects, subject)
 	}
 
-	ownerRef := rb.GetOwnerReferences()
-	if len(ownerRef) == 0 {
-		rb.SetOwnerReferences([]metav1.OwnerReference{r.ownerRef})
-	}
+	ownerRef := r.updateOwnerRefs(rb.GetOwnerReferences())
+	rb.SetOwnerReferences(ownerRef)
 
 	if hasSubject && (len(ownerRef) != 0) {
 		logger.Info("rolebinding is up to date", "action", "none")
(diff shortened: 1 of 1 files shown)


pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -450,6 +450,18 @@ func (r *rbac) updateClusterRoleBinding(ctx context.Context, rb *rbacv1.ClusterR
 		rb.SetOwnerReferences([]metav1.OwnerReference{r.ownerRef})
 	}
 
+	doUpdateOwnerRef := true
```
sm43, line 453: there are few other places in files where we check `len(ownerRef) ==0` and then add owner ref
while updating a resource.
should we add similar for those cases also?

sm43: @nikhil-thomas

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -450,6 +450,18 @@ func (r *rbac) updateClusterRoleBinding(ctx context.Context, rb *rbacv1.ClusterR
...
+			doUpdateOwnerRef = false
+			break
+		}
+	}
+
+	if doUpdateOwnerRef {
+		ownerRef = append(ownerRef, r.ownerRef)
+	}
+
 	if hasSubject && (len(ownerRef) != 0) {
```
sm43, line 465: this would be now 
`if hasSubject && !doUpdateOwnerRef  {` ???

nikhil-thomas: i think `(len(ownerRef) != 0)` will catch both cases (needed update, or didn't need update).

pkg/reconciler/openshift/tektonconfig/rbac.go
nikhil-thomas: @vinamra28 could you add a similar change in https://github.com/tektoncd/operator/blob/f061c934d3caf4b24ffa6f1d85f2d3658860d76c/pkg/reconciler/openshift/tektonconfig/rbac.go#L323

pradeepitm12 (commented):
/lgtm

pradeepitm12:
/hold
Would be nice to have a test for this.

PR #355, opened 2021-07-07
[OpenShift] Add midstream script as extension

sm43:
This adds midstream script as code for pipelines, triggers. this directly does
the changes on upstream release yaml.
This does following changes:
for pipelines: 
    - add monitoring label in namespace
    - add prefix (quay.io/openshift-pipelines/tektoncd-pipelines-) in deployment images and
        images in args with a few exception 
    - removes runAsUser from all deployment containers
for triggers:
    - add prefix (quay.io/openshift-pipelines/tektoncd-triggers-) in deployment images and
        images in args 
    - removes runAsUser runAsGroup from all deployment containers

Signed-off-by: Shivam Mukhade smukhade@redhat.com

# Submitter Checklist

review them:

# Release Notes


cmd/openshift/operator/kodata/tekton-pipeline/0.24.3/00-pipelines.yaml
```diff
@@ -0,0 +1 @@
+./../../../../../../upstream/pipelines/0.24.3/release.yaml
```
vdemeester, line 1: So this would be at the root of the git repository right ? I think we should / could use something a bit different, like `payload` or something ? (`upstream` doesn't speak too much to me 😛 )

pkg/reconciler/common/proxy.go
```diff
@@ -81,24 +79,6 @@ func ApplyProxySettings(u *unstructured.Unstructured) error {
...
-func InjectLabelOnNamespace() mf.Transformer {
-	return func(u *unstructured.Unstructured) error {
-		kind := strings.ToLower(u.GetKind())
-		if kind != "namespace" {
-			return nil
-		}
-		name := u.GetName()
-		if name == "tekton-pipelines" || name == "openshift-pipelines" {
-			labels := u.GetLabels()
-			labels["operator.tekton.dev/disable-proxy"] = "true"
```
vdemeester, line 95: Side note to @piyush-garg, are we also skipping `openshift-*` pods for the proxy ?

pkg/reconciler/openshift/tektonpipeline/extension.go
```diff
@@ -75,9 +83,11 @@ type openshiftExtension struct {
 
 func (oe openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Transformer {
 	return []mf.Transformer{
+		common.InjectLabelOnNamespace(monitoringLabel),
```
vdemeester, line 86: This might need to change with the 1.5 change on disabling the service monitor, right ?

vdemeester (approved):
Few question, but looks good to me

cmd/openshift/operator/kodata/tekton-pipeline/0.24.3/00-pipelines.yaml
sm43: `payload` is fine 😄

pkg/reconciler/openshift/tektonpipeline/extension.go
sm43: yep after this pr I will handle that in openshift extension

PR #38, opened 2019-08-25
Enable the client generation

houshengbo:
# Changes

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/operator/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```
Describe any user facing changes here, or delete this block.

Examples of user facing changes:
- API changes
- Bug fixes
- Any changes in behavior

```

PR #364, opened 2021-07-14
fix autoprune not exectuing for second resource #38

pradeepitm12:
earlier resources["pipelinerun", "taskrun"] executes fine
but somehow resources ["taskrun", "pipelinrun"] do not execute
for the second resource.

problem: command created and passed was an array where each
item was treated as an arugement.

Now the commands are single string and seperated by ';'

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/common/prune.go
+++ b/pkg/reconciler/common/prune.go
@@ -142,7 +142,7 @@ func getPruningContainers(resources, namespaces []string, keep uint, tknImage st
 			Name:                     jobName,
 			Image:                    tknImage,
 			Command:                  []string{"/bin/sh", "-c"},
-			Args:                     cmdArgs,
+			Args:                     []string{cmdArgs},
 			TerminationMessagePolicy: "FallbackToLogsOnError",
 		}
 		containers = append(containers, container)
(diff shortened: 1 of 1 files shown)


pradeepitm12:
cc: @nikhil-thomas @sm43

pkg/reconciler/common/prune.go
```diff
@@ -151,11 +151,11 @@ func getPruningContainers(resources, namespaces []string, keep uint, tknImage st
 	return containers
 }
 
-func deleteCommand(resources []string, keep uint, ns string) []string {
-	cmdArgs := []string{}
+func deleteCommand(resources []string, keep uint, ns string) string {
```
pradeepitm12, line 154: there is a diff uint and int with respect to the downstream.

sm43:
/lgtm
just one ques would `tkn <res> delete` cmd delete a running pr/tr?
if yes, then if user has kept for ex. `keep 1` and right when the puner job is started there is a pipeline running
so would the job delete pr for it?

pradeepitm12:
> /lgtm
> just one ques would `tkn <res> delete` cmd delete a running pr/tr?
> if yes, then if user has kept for ex. `keep 1` and right when the puner job is started there is a pipeline running
> so would the job delete pr for it?

Yes, it will delete a running pr/tr.
As keep=1 would keep 1 pipelinerun which is latest(running) and delete others.

nikhil-thomas:
@pradeepitm12 could you add an issue to track this case

> > /lgtm
> > just one ques would `tkn <res> delete` cmd delete a running pr/tr?
> > if yes, then if user has kept for ex. `keep 1` and right when the puner job is started there is a pipeline running
> > so would the job delete pr for it?
> 
> Yes, it will delete a running pr/tr.
> As keep=1 pruner would keep 1 pipelinerun which is latest and delete others, in this latest should be the one running.

/approve
/lgtm

PR #371, opened 2021-07-29
[OpenShift] Removes external reference in openshift payload

sm43:
This removes external reference for one of pipelines images in
the payload. this removes it and add no image available instead
of a gcr image.

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/openshift/common/testdata/test-prefix-images-expected.yaml
+++ b/pkg/reconciler/openshift/common/testdata/test-prefix-images-expected.yaml
@@ -63,7 +63,7 @@ spec:
             "-pr-image", "quay.io/openshift-pipeline/tektoncd-pipeline-pullrequest-init:v0.24.3",
 
             # This is gcr.io/google.com/cloudsdktool/cloud-sdk:302.0.0-slim
-            "-gsutil-image", "gcr.io/google.com/cloudsdktool/cloud-sdk@sha256:27b2c22bf259d9bc1a291e99c63791ba0c27a04d2db0a43241ba0f1f20f4067f",
+            "-gsutil-image", "no-image-available",
             # The shell image must be root in order to create directories and copy files to PVCs.
             # gcr.io/distroless/base:debug as of Apirl 17, 2021
             # image shall not contains tag, so it will be supported on a runtime like cri-o
--- a/pkg/reconciler/openshift/common/transformer.go
+++ b/pkg/reconciler/openshift/common/transformer.go
@@ -39,7 +39,7 @@ import (
 
 // UpdateDeployments will also remove runAsUser from container
 
-func UpdateDeployments(prefix string, replaceImg map[string]string, skipImages []string) mf.Transformer {
+func UpdateDeployments(prefix string, replaceImg map[string]string) mf.Transformer {
 	return func(u *unstructured.Unstructured) error {
 		if u.GetKind() != "Deployment" {
 			return nil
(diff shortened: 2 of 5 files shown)


vdemeester (approved):

pkg/reconciler/openshift/common/testdata/test-prefix-images-expected.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             "-pr-image", "quay.io/openshift-pipeline/tektoncd-pipeline-pullrequest-init:v0.24.3",
 
             # This is gcr.io/google.com/cloudsdktool/cloud-sdk:302.0.0-slim
-            "-gsutil-image", "gcr.io/google.com/cloudsdktool/cloud-sdk@sha256:27b2c22bf259d9bc1a291e99c63791ba0c27a04d2db0a43241ba0f1f20f4067f",
+            "-gsutil-image", "no-image-available",
```
nikhil-thomas, line 66: let us keep the `quay.io/openshift-pipeline/tektoncd-pipeline-nop:v0.24.3"` from https://github.com/tektoncd/operator/pull/371/files#diff-18670a9fe9564fddf949b33a4dbcfbbda752c2f740fa0927964b1d516f9a2641R61
as the replacement.

sm43: @nikhil-thomas just in case if a user is using PipelineResource and if we keep `quay.io/openshift-pipeline/tektoncd-pipeline-nop:v0.24.3` it might give an impression that it not working before bcoz of some error on there side 
they might not know what a `nop` image is for
if we are not publishing an image for that feature so `no-image-available` would make more sense than passing some image that will not work 
wdyt?

nikhil-thomas: > it might give an impression that it not working before bcoz of some error on there side

i agree with this 👍  

I would still prefer a pullable image because:
- this image is specific to pipelineResource type storage (specifically gcs). 
   this means we can push the possiblity of the error in to the taskrun pod scope 
- a pullable image will ensure that no other validations at the controller level will break (present now or new ones in the future).
- the enduser will get an `gsutil` executable not found error in the logs. so the user won't be completely clueless.

pkg/reconciler/openshift/common/transformer.go
```diff
@@ -59,10 +59,6 @@ func UpdateDeployments(prefix string, replaceImg map[string]string, skipImages [
 
 			// Prefix Images in Args if there
 			for i := 0; i < len(c.Args); i++ {
-				if valueInArr(skipImages, c.Args[i]) {
-					i++
-					continue
-				}
```
nikhil-thomas, line 65: hi, im just curious. 
why does this patch remove this.

pkg/reconciler/openshift/common/testdata/test-prefix-images-expected.yaml
sm43: hardcoding the image for `-gsutil-image` is bit tricky as we will not know the tag to use with the image. it will need a separate transformer which will fetch the tag and replace the image.

pkg/reconciler/openshift/common/transformer.go
sm43: this code was specific to skip an image 
func does 2 things
- add prefix to images 
- replace image with some other in some cases
There was 3rd where we use to skip the image but this code was specific to the 3rd case which only for gsutil image
and as we are removing external ref from payload in future too we will be doing that 
so there won't a case to skip an image.

pkg/reconciler/openshift/common/testdata/test-prefix-images-expected.yaml
nikhil-thomas: it will be alright, as like other images in this args list, we can specify a toplevel Image replacement as env (already existing feature). The version can be upgraded to the latest during a build/productization.

vdemeester: We may also want to see in `tektoncd/pipeline` if there is an issue tracking this and if we can.. have this at pipeline level (aka possibility to not give any image and let pipeline handle that).

PR #376, opened 2021-08-10
Update release pipelines by adding payload fetch Task

nikhil-thomas:
Add `operator-fetch-component-releases` Task to run the
`/hack/fetch-releases.sh` script.

Update release pipeline by adding the payload fetch Task

Add Dashboard and Results support to payload fetch Task

# Changes

# Submitter Checklist

review them:

# Release Notes


vdemeester (approved):

cmd/kubernetes/operator/kodata/tekton-dashboard/0.0.0-latest/00-dashboard.yaml
```diff
@@ -87,6 +87,7 @@ rules:
   - apiGroups:
```
sm43, line 87: this file can be deleted now right?

cmd/kubernetes/operator/kodata/tekton-trigger/0.0.0-latest/01-interceptors.yaml
```diff
@@ -0,0 +1,166 @@
+# Copyright 2020 The Tekton Authors
```
sm43, line 1: same here? do we need to keep this file?

hack/fetch-releases.sh
```diff
@@ -1,86 +1,113 @@
...
+  echo dashboard: ${d_version}
+  echo results: ${r_version}
+  echo ""
+
+  release_yaml pipeline release.notags 00-pipelines ${p_version}
+
+  release_yaml triggers release.notags 00-triggers ${t_version}
+  release_yaml triggers interceptors.notags 01-interceptors ${t_version}
+
+  release_yaml dashboard tekton-dashboard-release 00-dashboard ${d_version}
```
sm43, line 105: in case of openshift we don't need to fetch yaml for result and dashboard 
so probably we need to check here

pkg/reconciler/common/releases.go
```diff
@@ -88,7 +88,7 @@ func ComponentDir(instance v1alpha1.TektonComponent) string {
 	case *v1alpha1.TektonConfig:
 		return filepath.Join(koDataDir, "tekton-config")
 	case *v1alpha1.TektonResult:
-		return filepath.Join(koDataDir, "tekton-result")
+		return filepath.Join(koDataDir, "tekton-results")
```
sm43, line 91: may be we can keep `result` here 😅  similar to others
and add a condition in script similar to what we have for triggers
```
   if [[ $comp == "triggers" ]]; then
     dir="trigger"
    fi
```

hack/fetch-releases.sh
```diff
@@ -1,86 +1,113 @@
-#!/bin/bash
+#!/usr/bin/env sh
```
sm43, line 1: if we are going to user `sh`
then we need update the if conditions
`[[ ]]` is not supported I guess

sm43: yep ci showing error for this

test/e2e-common.sh
```diff
@@ -28,6 +28,8 @@ source $(dirname $0)/../vendor/github.com/tektoncd/plumbing/scripts/e2e-tests.sh
 function install_operator_resources() {
   echo ">> Deploying Tekton Operator Resources"
 
+  make get-releases TARGE=${TARGET:-kubernetes}
```
sm43, line 31: in ci this will run twice now, as we run it with `make apply` also

hack/fetch-releases.sh
vdemeester:
```
#!/usr/bin/env bash
```

pkg/reconciler/common/releases.go
nikhil-thomas: i thought about it, but i think this will be better as the script will be a bit more simple.

hack/fetch-releases.sh
```diff
@@ -1,86 +1,109 @@
...
+    # create a directory
+    dirPath=${comp_dir}/${dirVersion}
+    mkdir -p ${dirPath} || true
+
+    # destination file
+    dest=${dirPath}/${destFileName}.yaml
+    http_response=$(curl -s -o ${dest} -w "%{http_code}" ${url})
+    echo url: ${url}
+
+    if [ $http_response != "200" ]; then
```
vdemeester, line 61: should we use always `[[` for consistency ?

nikhil-thomas: yup 👍 
handled

pradeepitm12 (approved):
/lgtm

PR #377, opened 2021-08-10
[OpenShift] fixes the disable affinity assistant configuration

### LINKED ISSUE #372: AffinityAssistant configuration on OpenShift not taken into account
# Expected Behavior

Changing the TektonPipeline configuration `disable-affinity-assistant` to `false` is taken into account under OpenShift and the affinity assistant is enabled.

# Actual Behavior

Changing the TektonPipeline configuration `disable-affinity-assistant` to `false` is not taken into account, it is forced to `true` by the controller (see [here](https://github.com/tektoncd/operator/blob/main/pkg/reconciler/openshift/tektonpipeline/extension.go#L141))

1. Install the operator
2. Update the `TektonPipeline` CR
3. Look at the `openshift/pipelines` `feature-flags` configmap.

/cc @nikhil-thomas @sm43 to correct me if I am saying something wrong

sm43:
This is always true in case of openshift due to which user would
not able to enable it. so this fixes by removing the defaulting in
Operator CR and adding the field only when user adds it.
By default, the value in configmap is false for kubernetes and will
be changed to true for openshift which user can change if requires.

fixes #372 

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/openshift/tektonpipeline/extension.go
+++ b/pkg/reconciler/openshift/tektonpipeline/extension.go
@@ -33,6 +33,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/injection"
 	"knative.dev/pkg/logging"
+	"knative.dev/pkg/ptr"
 )
 
 const (
@@ -135,10 +136,10 @@ func SetDefault(properties *v1alpha1.PipelineProperties) bool {
 		updated = true
 	}
 
-	// Set `disable-affinity-assistant` to true
-	// webhook will set `false` as default value
-	if properties.DisableAffinityAssistant == nil || !*properties.DisableAffinityAssistant {
-		*properties.DisableAffinityAssistant = DefaultDisableAffinityAssistant
+	// Set `disable-affinity-assistant` to true if not set in CR
+	// webhook will not set any value but by default in pipelines configmap it will be false
+	if properties.DisableAffinityAssistant == nil {
+		properties.DisableAffinityAssistant = ptr.Bool(DefaultDisableAffinityAssistant)
 		updated = true
 	}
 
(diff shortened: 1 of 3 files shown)


vdemeester (approved):

pkg/reconciler/openshift/tektonpipeline/extension.go
```diff
@@ -135,10 +136,10 @@ func SetDefault(properties *v1alpha1.PipelineProperties) bool {
 		updated = true
 	}
 
-	// Set `disable-affinity-assistant` to true
-	// webhook will set `false` as default value
-	if properties.DisableAffinityAssistant == nil || !*properties.DisableAffinityAssistant {
-		*properties.DisableAffinityAssistant = DefaultDisableAffinityAssistant
+	// Set `disable-affinity-assistant` to true if not set in CR
+	// webhook will not set any value but by default in pipelines configmap it will be false
+	if properties.DisableAffinityAssistant == nil {
```
savitaashture, line 141: so from Operator we can configure `DisableAffinityAssistant` only for Openshift and not for K8s then in that case can we add comment here https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonpipeline_types.go#L84 so that it will be more clear

not necessarily in this instead we do it in another PR

WDYT?

vdemeester: @savitaashture it's support on all platform. Only on Openshift the default is different (`true` instead of `false`)

sm43: yep. on kubernetes by default the field will not be set in TektonConfig CR, user can define and alter value.
by default it will be false in configmap.

vdemeester: @sm43 but on k8s, if the user set then in the `TektonConfig` CR, the will be applied to the managed tekton instance right ?

sm43: yes.. user can define the field in k8s and change the value. there is no behavior change

savitaashture:
/lgtm

/hold 
For others to take look

vdemeester (approved):

PR #385, opened 2021-08-18
Refactors TektonPipeline Reconciler

sm43:

# Changes

# Submitter Checklist

review them:

# Release Notes


pkg/apis/operator/v1alpha1/tektonpipeline_lifecycle.go
```diff
@@ -21,13 +21,25 @@ import (
...
 )
 
+const (
+	CrdInstalled         apis.ConditionType = "CrdInstalled"
+	ClustersScoped       apis.ConditionType = "ClusterScopedInstalled"
+	NamespaceScoped      apis.ConditionType = "NamespaceScopedInstalled"
+	DeploymentsInstalled apis.ConditionType = "DeploymentInstalled"
+	WebhookReady         apis.ConditionType = "WebhookReady"
+	ControllerReady      apis.ConditionType = "ControllerReady"
+)
```
nikhil-thomas, line 31:
```suggestion
	CrdInstalled         apis.ConditionType = "CRDsInstalled"
	ClustersScoped       apis.ConditionType = "ClusterScopedResourcesInstalled"
	NamespaceScoped      apis.ConditionType = "NamespaceScopedResourcesInstalled"
	DeploymentsInstalled apis.ConditionType = "DeploymentsAvailable"
	WebhookReady         apis.ConditionType = "WebhooksReady"
	ControllerReady      apis.ConditionType = "ControllersReady"
)
```

So that we can make our Status more `read optimized and meaningful` for our users

pkg/apis/operator/v1alpha1/tektonpipeline_lifecycle.go
```diff
@@ -60,6 +76,85 @@ func (tps *TektonPipelineStatus) MarkInstallSucceeded() {
...
+	tps.MarkNotReady("Controller Deployment not available ")
+	pipelineCondSet.Manage(tps).MarkFalse(
+		ControllerReady,
+		"Error",
+		"Controller: %s", msg)
+}
+
 // MarkInstallFailed marks the InstallationSucceeded status as false with the given
 // message.
 func (tps *TektonPipelineStatus) MarkInstallFailed(msg string) {
```
nikhil-thomas, line 160: shall we remove functions like these as the respective "coditions" are removed in this pr.

pkg/reconciler/installer/init.go
```diff
@@ -0,0 +1,169 @@
...
+		deploymentPred,
+		servicePred,
+	); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (i *Installer) IsInstalled(pred ...mf.Predicate) error {
```
nikhil-thomas, line 100: a better name for this function could be `CreateIfNotExists` 🧑‍💻

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -19,24 +19,33 @@ package tektonpipeline
...
 const (
+	// Pipelines ConfigMap
 	featureFlag    = "feature-flags"
 	configDefaults = "config-defaults"
+
+	proxyLabel = "operator.tekton.dev/disable-proxy=true"
+)
+
+var (
+	lastAppliedManifest mf.Manifest
```
nikhil-thomas, line 48: let us avoid this completely. This means we are storing state information related to an installtion inside a controller.
This might look ok in our case as we can assume, there will be only one instance of our CRD. However, this could be very confusing for anyone new reading this code as this essentially is storing state in a stateless component (controller).

nikhil-thomas: i have added comments on 2 different places in this file where `lastAppliedManifest` could be replaced/eliminated.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -80,13 +89,15 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Tekton
...
-	if err != nil {
-		logger.Error("Unable to fetch installed manifest; no cluster-scoped resources will be finalized", err)
-		return nil
+
+	if reflect.DeepEqual(lastAppliedManifest, mf.Manifest{}) {
+		lastAppliedManifest = r.manifest
+		if err := r.transform(ctx, &lastAppliedManifest, original); err != nil {
+			return err
+		}
 	}
```
nikhil-thomas, line 98:
```suggestion
     lastAppliedManifest = r.manifest
     if err := r.transform(ctx, &lastAppliedManifest, original); err != nil {
	return err
    }
```

can we assume that when the finzlizer is called there will be no change in spec after the `DeletionTimeStamp` is set.
In that case, we can perform the deletion on transformed list of r.manifest.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -96,47 +107,166 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Tekton
...
+		ready := tp.Status.GetCondition(apis.ConditionReady)
+		if ready != nil && ready.Status == corev1.ConditionTrue {
+			tp.Status.MarkNotReady("spec changed, reconciling")
+			return fmt.Errorf("reconcile again")
+		}
+
+		lastAppliedManifest = r.manifest
+		if err := r.transform(ctx, &lastAppliedManifest, tp); err != nil {
+			tp.Status.MarkNotReady("transformation failed: " + err.Error())
+			return err
```
nikhil-thomas, line 163: it looks like here, the `lastAppliedManifest`  is stored for the sake of Finalizer. 

In other words, we are using `lastAppliedManifest` to communicate between 2 different reconcile loops.
Let us avoid doing that, as this makes a counter-intutive assumption, uncommon assumption that a piece of state will be recorded in a variable by a previous invocation of the reconciler.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
nikhil-thomas: In short, any restart of controller pod will reset `lastAppliedManifest` to empty.

nikhil-thomas:
The refacotring is great. Now the reconciler is more Readable 👍 
I have a few comments apart from that everything else looks great.

pkg/apis/operator/v1alpha1/tektonpipeline_lifecycle.go
sm43: some are still there as we user `TektonComponent` so we have to define if we define the interface
but we can do the cleanup as we refactor other components

pkg/apis/operator/v1alpha1/tektonpipeline_lifecycle.go
```diff
@@ -17,111 +17,155 @@ limitations under the License.
...
 
-// SetVersion sets the currently installed version of the component.
 func (tps *TektonPipelineStatus) SetVersion(version string) {
 	tps.Version = version
 }
 
-// GetManifests gets the url links of the manifests.
-func (tps *TektonPipelineStatus) GetManifests() []string {
-	return tps.Manifests
+func (tps *TektonPipelineStatus) MarkInstallSucceeded() {
```
sm43, line 141: this are not used now but we need to define as `extension` uses `TektonComponent` 
but will be removed as refactor other components refactor the TektonComponent

pkg/apis/operator/v1alpha1/common.go
```diff
@@ -31,7 +31,7 @@ const (
 	InstallSucceeded apis.ConditionType = "InstallSucceeded"
 	// DeploymentsAvailable is a Condition indicating whether or not the Deployments of
 	// the respective component have come up successfully.
-	DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
+	// DeploymentsAvailable apis.ConditionType = "DeploymentsAvailable"
```
nikhil-thomas, line 34:
```suggestion
```

pkg/apis/operator/v1alpha1/tektonpipeline_lifecycle.go
```diff
@@ -21,107 +21,145 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var (
-	_ TektonComponentStatus = (*TektonPipelineStatus)(nil)
```
nikhil-thomas, line 25: can we add this check back when we refactor the other reconcilers. 
if yes, could you comment this for now instead of removing it. Also mark it as TODO.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -64,31 +73,10 @@ var _ tektonpipelinereconciler.Finalizer = (*Reconciler)(nil)
...
-	logger.Info("Deleting cluster-scoped resources")
-	manifest, err := r.installed(ctx, original)
-	if err != nil {
-		logger.Error("Unable to fetch installed manifest; no cluster-scoped resources will be finalized", err)
-		return nil
-	}
-	if err := common.Uninstall(ctx, manifest, nil); err != nil {
-		logger.Error("Failed to finalize platform resources", err)
-	}
+
```
nikhil-thomas, line 79: why does this finalizer don't have a Delete call on the associated InstallerSet.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -97,46 +85,213 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Tekton
...
 		return err
 	}
-	stages := common.Stages{
-		r.appendPipelineTarget,
-		r.transform,
-		common.Install,
-		common.CheckDeployments,
+
+	// Mark PreReconcile Complete
+	tp.Status.MarkPreReconcilerComplete()
```
nikhil-thomas, line 108: wow! nice 👍

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -97,46 +85,213 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Tekton
...
+	} else if ready.Status == corev1.ConditionFalse {
+		tp.Status.MarkInstallerSetNotReady(ready.Message)
+		r.enqueueAfter(tp, 10*time.Second)
+		return nil
+	}
+
+	// Mark InstallerSet Ready
+	tp.Status.MarkInstallerSetReady()
+
+	if err := r.extension.PostReconcile(ctx, tp); err != nil {
```
nikhil-thomas, line 226: 👍

pkg/apis/operator/v1alpha1/tektonpipeline_lifecycle.go
sm43: yes we can add this back later, will add a comment

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
sm43: We can have a delete call, but as we have owner ref for InstallerSet, it will be garbage collected.
It will take some time to delete all resources after deleting TektonPipeline.
Deleting TektonPipeline would be when user want to uninstall and delete everything, so allowing it sometime to be garbage collected would be fine I think.
In case of InstallerSet controller might have to create a new so we delete all resources explicitly
wdyt?

nikhil-thomas: let us add the explict deletion to be on the safe side. 
Past experience has shown that we cannot rely 100% on gargbage colelction, as it is non-blocking.

nikhil-thomas: in addition, as we had the explict deletion before refactoring, let us add the explicit deletion. so that we don't get any surprises.
ref: https://github.com/tektoncd/operator/blob/dcf6156122cd8389673623640b3433d94b83d208/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L89

Issue #358, opened 2021-07-12
Changing profile in TektonConfig doesn't reflect on components  


sm43:
# Expected Behavior
changing profile should reflect on components
In case of OpenShift
changing profile from `all` to `basic` should delete the tekton addon resource 

# Actual Behavior

changing profile doesn't have any effect on components.

# Steps to Reproduce the Problem
Tested on Openshift but I think would be similar on Kubernetes platform too
1. Install Operator (by default it will install TektonConfig with all profile)
2. Change profile to basic in TektonConfig
3. Addon resource should get deleted but still accessible

PR #394, opened 2021-08-27
Removes the resources based on the profile set

PuneetPunamiya:
When the Tekton Config profile is changed, based on the profile set 
components are deleted

For example:
If the profile is changed from all to basic:
* In case of Kubernetes: `Dashboard` Component is deleted
* In case of Openshift: `Addon` Component is deleted

Fixes: #358

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/kubernetes/tektonconfig/extension.go
+++ b/pkg/reconciler/kubernetes/tektonconfig/extension.go
@@ -25,6 +25,7 @@ import (
 	operatorclient "github.com/tektoncd/operator/pkg/client/injection/client"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"github.com/tektoncd/operator/pkg/reconciler/kubernetes/tektonconfig/extension"
+	"k8s.io/apimachinery/pkg/api/errors"
 )
 
 func KubernetesExtension(ctx context.Context) common.Extension {
@@ -45,9 +46,21 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
 }
 func (oe kubernetesExtension) PostReconcile(ctx context.Context, comp v1alpha1.TektonComponent) error {
 	configInstance := comp.(*v1alpha1.TektonConfig)
+
 	if configInstance.Spec.Profile == common.ProfileAll {
 		return extension.CreateDashboardCR(comp, oe.operatorClientSet.OperatorV1alpha1())
 	}
+
+	if configInstance.Spec.Profile == common.ProfileLite || configInstance.Spec.Profile == common.ProfileBasic {
+		err := extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+	}
+
 	return nil
 }
 func (oe kubernetesExtension) Finalize(ctx context.Context, comp v1alpha1.TektonComponent) error {
--- a/pkg/reconciler/kubernetes/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/kubernetes/tektonconfig/tektonconfig.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	mf "github.com/manifestival/manifestival"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
@@ -30,6 +31,7 @@ import (
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"github.com/tektoncd/operator/pkg/reconciler/kubernetes/tektonconfig/pipeline"
 	"github.com/tektoncd/operator/pkg/reconciler/kubernetes/tektonconfig/trigger"
+	tt "github.com/tektoncd/operator/pkg/reconciler/kubernetes/tektonconfig/trigger"
 	"knative.dev/pkg/logging"
 	pkgreconciler "knative.dev/pkg/reconciler"
 )
@@ -119,6 +121,13 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 
 	var stages common.Stages
 	if tc.Spec.Profile == common.ProfileLite {
+		err := tt.TektonTriggerCRDelete(r.operatorClientSet.OperatorV1alpha1().TektonTriggers(), common.TriggerResourceName)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
 		stages = common.Stages{
 			r.createPipelineCR,
 		}
--- a/pkg/reconciler/openshift/tektonconfig/extension.go
+++ b/pkg/reconciler/openshift/tektonconfig/extension.go
@@ -30,6 +30,7 @@ import (
 	openshiftPipeline "github.com/tektoncd/operator/pkg/reconciler/openshift/tektonpipeline"
 	openshiftTrigger "github.com/tektoncd/operator/pkg/reconciler/openshift/tektontrigger"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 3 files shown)


pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -45,9 +46,40 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
...
-	if configInstance.Spec.Profile == common.ProfileAll {
+
+	switch configInstance.Spec.Profile {
+	case common.ProfileAll:
 		return extension.CreateDashboardCR(comp, oe.operatorClientSet.OperatorV1alpha1())
+
+	case common.ProfileBasic:
+		_, err := extension.GetDashboard(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+		if err != nil {
+			return err
```
nikhil-thomas, line 57:
```suggestion
			if errors.IsNotFound(err) {
   			    return nil
			}
			return err
```

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -45,9 +46,40 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
...
+		}
+		return extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+
+	case common.ProfileLite:
+		dashboard, err := extension.GetDashboard(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+		if dashboard != nil {
+			err := extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+			if err != nil {
+				return err
```
nikhil-thomas, line 66:
```suggestion
			if errors.IsNotFound(err) {
   			    return nil
			}
			return err
```

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -45,9 +46,40 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
...
+			}
+		}
+		if err != nil {
+			trigger, err := tt.GetTrigger(oe.operatorClientSet.OperatorV1alpha1().TektonTriggers(), common.TriggerResourceName)
+			if trigger != nil {
+				err := tt.TektonTriggerCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonTriggers(), common.TriggerResourceName)
+				if err != nil {
+					return err
```
nikhil-thomas, line 74:
```suggestion
			if errors.IsNotFound(err) {
   			    return nil
			}
			return err
```

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -45,9 +46,40 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
...
+	case common.ProfileBasic:
+		_, err := extension.GetDashboard(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+		if err != nil {
+			return err
+		}
+		return extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+
+	case common.ProfileLite:
+		dashboard, err := extension.GetDashboard(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+		if dashboard != nil {
```
nikhil-thomas, line 63:
```suggestion
			if errors.IsNotFound(err) {
   			    return nil
			}
```
i think a check whether there is no error should be enough to decide to delete

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -94,6 +94,13 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 			return err
 		}
 	}
+
+	if configInstance.Spec.Profile == common.ProfileBasic {
+		addon, _ := extension.GetAddon(oe.operatorClientSet.OperatorV1alpha1().TektonAddons(), common.DashboardResourceName)
```
nikhil-thomas, line 99: this err should also be checked, isn't it :)

PuneetPunamiya: yes 👍🏻

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -94,6 +94,13 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
...
+		if addon != nil {
+			return extension.TektonAddonCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonAddons(), common.AddonResourceName)
```
savitaashture, line 101: Don't we require deletion of triggers here for lite profile :thinking:

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -45,9 +46,40 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
...
+			err := extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+			if err != nil {
+				return err
+			}
+		}
+		if err != nil {
```
savitaashture, line 69: If dashboard exist then above err https://github.com/tektoncd/operator/pull/394/files#diff-a100309e66fd0aa1063e2e110fd0d6d97eaa6516bf10a9f61285828a7a0697c1R62 can be nil in that case this function will never be called right :thinking: or am i missing something

pkg/reconciler/openshift/tektonconfig/extension.go
PuneetPunamiya: Yes it requires, have updated it!

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -45,9 +47,48 @@ func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonCompo
...
+				return nil
+			}
+			return err
+		}
+		return extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+
+	case common.ProfileLite:
+		_, err := extension.GetDashboard(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
+		if err != nil {
+			_, err := tt.GetTrigger(oe.operatorClientSet.OperatorV1alpha1().TektonTriggers(), common.TriggerResourceName)
```
sm43, line 68: triggers is common to both platform could the deletion logic be in reconciler instead of extension 🤔

PuneetPunamiya: yes I think it makes sense to add check for triggers in reconciler 👍🏻

pkg/reconciler/kubernetes/tektonconfig/extension.go
PuneetPunamiya: I have handled this case in the reconciler now

pkg/reconciler/kubernetes/tektonconfig/tektonconfig.go
```diff
@@ -119,6 +121,20 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 
 	var stages common.Stages
 	if tc.Spec.Profile == common.ProfileLite {
+		_, err := tt.GetTrigger(r.operatorClientSet.OperatorV1alpha1().TektonTriggers(), common.TriggerResourceName)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil
```
nikhil-thomas, line 127: here if the error is `NotFound` then we should not do anything, the execution flow down to the `delete` call, shouldn't it.

nikhil-thomas: on second thougt, does this mechanism really need to do a `Get` API call. 
could did directly call `Delete` and ignore `NotFound` error if the item doesnot exist.

PuneetPunamiya: updated !

PR #403, opened 2021-09-06
 [OpenShift] Add param to enable/disable servicemonitor 

sm43:
This adds a param in TektonConfig & TektonPipeline  to disable service monitor
which is a part of TektonPipelines for OpenShift.
Param will be added by default, user can change the value in TektonConfig
to false to disable.

Value is added in openshift extension and not in webhook as this is
platform specific param.

If invalid value is passed then default	value is set for param to avoid
installation failure. Users can pass valid values and change the param.

------------------------------
Also,

 Handle upgrade from older version which has pruner defaults

Before webhook we had default value for pruner's keep as 1
but we expect user to define all values now so if a user has
installed prev version and has not enabled pruner then `keep`
will have a value 1 and after upgrading to newer version,
webhook will fail if keep has a value and other fields are not defined
this handles that case by removing the default for keep if
other pruner fields are not defined

# Submitter Checklist

review them:

# Release Notes


vdemeester (approved):

pkg/apis/operator/v1alpha1/tektonconfig_defaults.go
```diff
@@ -26,4 +26,20 @@ func (tc *TektonConfig) SetDefaults(ctx context.Context) {
...
+
+	// before webhook we had default value for pruner's keep as 1
+	// but we expect user to define all values now
+	// so if a user has installed prev version and has not enabled
+	// pruner then `keep` will have a value 1 and after upgrading
+	// to newer version webhook will fail if keep has a value and
+	// other fields are not defined
+	// this handles that case by removing the default for keep if
+	// other pruner fields are not defined
+	if len(tc.Spec.Pruner.Resources) == 0 {
```
nikhil-thomas, line 38: @pradeepitm12  if the neither `pipelinerun` or `taskrun` is there in the list, is that interpreted as prune both or prune none.

pkg/apis/operator/v1alpha1/tektonpipeline_types.go
```diff
@@ -51,8 +51,8 @@ func (tp *TektonPipeline) GetStatus() TektonComponentStatus {
 
 // TektonPipelineSpec defines the desired state of TektonPipeline
 type TektonPipelineSpec struct {
-	CommonSpec         `json:",inline"`
-	PipelineProperties `json:",inline"`
+	CommonSpec `json:",inline"`
+	Pipeline   `json:",inline"`
```
nikhil-thomas, line 55: a structural suggestion. 

at present the type `Pipeline` is defined in `tektonconfig_types.go`. 

This part of the code will be more readable if we move the `Pipeline` struct into `tektonpipeline_types.go`, as we are using this struct inline in TektonPipelineSpec and in TektonConfigSpec as a named field. 

The main reason why this will be an improvement is because, github directed me to [here](https://github.com/tektoncd/operator/blob/27d571d2daa248856a8e53e361523a79fae10637/vendor/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L212:2). So it took deliberate look to find the right type definition. 🧑‍💻 It would have helped me a lot if the struct definition was in this file. 🙂

pkg/apis/operator/v1alpha1/tektonconfig_defaults.go
```diff
@@ -26,4 +26,20 @@ func (tc *TektonConfig) SetDefaults(ctx context.Context) {
...
+		tc.Spec.Pruner.Keep = nil
+		tc.Spec.Pruner.Schedule = ""
+	} else if tc.Spec.Pruner.Schedule == "" {
+		tc.Spec.Pruner.Keep = nil
+		tc.Spec.Pruner.Resources = []string{}
```
nikhil-thomas, line 43: im wondering whether this must be a defaulting operation or a validation operator where it fails to make the update when there is no `schedule` but there is a `keep`.

May be for the seamless upgrade we need to add a pre-reconcile job specifically to get the old pruneSpec into an acceptable form before the update reconcile.

pkg/reconciler/openshift/tektonpipeline/extension.go
```diff
@@ -66,39 +69,50 @@ type openshiftExtension struct {
...
 		occommon.RemoveRunAsUser(),
 	}
+
+	pipeline := comp.(*v1alpha1.TektonPipeline)
+
+	// Add monitoring label if metrics is enabled
+	value := findParam(pipeline.Spec.Params, enableMetricsKey)
+	if value == "" || value == "true" {
+		trns = append(trns, common.InjectLabelOnNamespace(monitoringLabel))
+	}
```
nikhil-thomas, line 83: I think this logic should be handled in the Pre-Reconcile. I think the `openshift-pipelines` namespace is a special case. 
 As it is not created by the TektonPipeline Reconciler, but by TektonPipeline OpenShift Extension (Prereconcile). This is because the targetnamespace is pre-requisite for other things this pre-recociler does.

I could be wrong. However, making this comment to make sure that we have considered the fact that the target-namespace is created by prereconcile.

nikhil-thomas:
@sm43 hi, let us get some clarity on "who" (extension prereconciler or main reconciler) creates the targetname space in builds for `OpenShift`. The answer could affect https://github.com/tektoncd/operator/pull/385.

pkg/reconciler/openshift/tektonpipeline/extension.go
sm43: main reconciler has release yaml which also has namespace so that will override on this
but extension has to be refactored as currently it will only install resources in openshift-pipelines namespace.

PR #404, opened 2021-09-09
Fixes deployment on OpenShift

sm43:
Moves TektonConfig ExtendedController to a separate pkg

TektonConfig's ExtendedController is used in both platforms k8s and
OpenShift and was in the k8s pkg. K8s controller use Dashboard
informer but OpenShift's doesn't need it. As we import the Dashboard
Informer pkg in k8s it has an init func which get executed and looks
for Dashboard CRD. This was happening with OpenShift where we don't
install Dashboard CRD now. so to avoid dashboard informer getting
initialize for OpenShift moving ExtendedController to a separate pkg.

-------------------------------------------------------------------------------------

 Enable TektonAddon Defaulting & Validation in Webhook

This fixes the failure of TektonAddon on OpenShift because of the
params not getting added to cr. this enables the defaulting
and validation in TektonConfig.

This also adds for k8s platform but it won't have any effect of the
installation of addon on k8s as it is not controller in not added
for k8s. This in plan to enable in upcoming days.

----------------------------------------------------------------------

 [OpenShift] Fixes Operator webhook

Earlier for Openshift, the specific files were not getting merged in
base due to which webhook was looking at tekton-operator ns instead
of openshift-operators, this fixes it.

And also updates the service account and rbac for webhook.

# Submitter Checklist

review them:

# Release Notes


config/openshift/webhook.yaml
```diff
@@ -30,23 +30,23 @@ spec:
         name: tekton-operator-webhook
         app: tekton-operator
     spec:
-      serviceAccountName: tekton-operator
+      serviceAccountName: openshift-pipelines-operator
```
vdemeester, line 33: Same remark as I did to @nikhil-thomas's PR, I think this should stay `tekton-operator` as this is in `tektoncd`, OpenShift Pipelines happens to use this but we shouldn't make it hardcoded I think

vdemeester (changes requested):
Not a fan of the package 😅 `pkg/reconciler/common-reconciler/…` there is `common` and a repetition there. What `pkg/reconciler/shared/…` instead ? (it's the "shared" code for all platform more or less).

Not to take in this PR, but `NewExtendedController` might be better named `NewExtensibleController`, shouldn't it ? 🤔

config/openshift/webhook.yaml
sm43: ah okay. I think that change of service account was not reverted due to which webhook is failing now.
controller uses `openshift-pipelines-operator` and webhook `tekton-operator` 
and on openshift `tekton-operator` is not created anymore

sm43:
> Not a fan of the package sweat_smile `pkg/reconciler/common-reconciler/…` there is `common` and a repetition there. What `pkg/reconciler/shared/…` instead ? (it's the "shared" code for all platform more or less).

@vdemeester 
yep shared is better. was not able to think of any 😅 will update. Thnx

> 
> Not to take in this PR, but `NewExtendedController` might be better named `NewExtensibleController`, shouldn't it ? thinking

yeah `NewExtensibleController` make sense

config/openshift/kustomization.yaml
```diff
@@ -14,43 +14,46 @@
...
+      kind: ClusterRole
+      name: tekton-operator
+  - path: role_binding.yaml
+    target:
+      kind: ClusterRoleBinding
+      name: tekton-operator
+  - path: webhook.yaml
+    target:
+      kind: Deployment
+      name: tekton-operator-webhook
```
savitaashture, line 43: does this change required ?

pkg/reconciler/common-reconciler/tektonconfig/controller.go
```diff
@@ -0,0 +1,97 @@
+/*
+Copyright 2020 The Tekton Authors
```
savitaashture, line 2:
```suggestion
Copyright 2021 The Tekton Authors
```

config/openshift/kustomization.yaml
sm43: ide identation 😅

pradeepitm12:
most of the review comments seem already addressed
/lgtm

config/openshift/webhook.yaml
vdemeester: ah I see. Then we should probably create an issue to track this 😝

vdemeester (approved):

PR #410, opened 2021-09-16
WIP: adds auto-prune per namespace

pradeepitm12:
This adds auto-pruning of taskrun and pipeline run per namespace
configured specifically.
adding unique schedule create a new cronjob in the annotated
namespace.
The default behaviour will persist, on top of it now user can
annotate a namespace with unique pruning configuration.
If user configures a unique schedule value by annotating a ns
a new cron job in that namespace will be created.

Auto puring is now enabled by default for default configurations
from tektonconfig.

Annotations introduced with this pr
 - "operator.tekton.dev/prune.skip"
 - "operator.tekton.dev/prune.resources"
 - "operator.tekton.dev/prune.resources.pipelineNames"
 - "operator.tekton.dev/prune.resources.taskNames"
 - "operator.tekton.dev/prune.keep"
 - "operator.tekton.dev/prune.keepsince"
 - "operator.tekton.dev/prune.schedule"

# Changes

# Submitter Checklist

review them:

# Release Notes


pkg/apis/operator/v1alpha1/tektonconfig_validation.go
```diff
@@ -62,10 +62,12 @@ func (p Prune) validate() *apis.FieldError {
 		errs = errs.Also(apis.ErrMissingField("spec.pruner.resources"))
 	}
 
-	if p.Keep == nil {
-		errs = errs.Also(apis.ErrMissingField("spec.pruner.keep"))
-	} else if *p.Keep == 0 {
+	if p.Keep == nil && p.KeepSince == nil {
+		errs = errs.Also(apis.ErrMissingOneOf("spec.pruner.keep", "spec.pruner.keep-since"))
```
nikhil-thomas, line 66: we also need to check if both are specified together. That should return an error, as we can accept only one of `keep` of `keepSince`

pkg/reconciler/openshift/tektonconfig/controller.go
```diff
@@ -21,43 +21,20 @@ import (
...
-	"knative.dev/pkg/kmeta"
 )
 
 // NewController initializes the controller and is called by the generated code
 // Registers eventhandlers to enqueue events
 func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
 	ctrl := tektonconfig.NewExtensibleController(OpenShiftExtension)(ctx, cmw)
-	namespaceInformer := namespaceinformer.Get(ctx)
-	namespaceInformer.Informer().AddEventHandler(controller.HandleAll(enqueueCustomName(ctrl, common.ConfigResourceName)))
-
```
nikhil-thomas, line 40: why does this patch remove this. this is required for both managing RBAC and for pruning. isn't it.

nikhil-thomas: i see it is moved to common  part of the reconciler. 🧑‍💻 👍

pkg/reconciler/common/prune.go
```diff
@@ -21,75 +21,199 @@ import (
...
+	CronName                    = "resource-pruner"
+	JobsTKNImageName            = "IMAGE_JOB_PRUNER_TKN"
+	ownerAPIVer                 = "operator.tekton.dev/v1alpha1"
+	ownerKind                   = "TektonConfig"
+	pruneSkip                   = "operator.tekton.dev/prune.skip"
+	pruneResources              = "operator.tekton.dev/prune.resources"
+	pruneResourcesPipelineNames = "operator.tekton.dev/prune.resources.pipelineNames"
+	pruneResourcesTaskNames     = "operator.tekton.dev/prune.resources.taskNames"
+	pruneKeep                   = "operator.tekton.dev/prune.keep"
+	pruneKeepSince              = "operator.tekton.dev/prune.keepsince"
```
nikhil-thomas, line 47: Our goal is to make `keep` and ` keepsince` mutually exclusive.
in the TektonConfig CR prune spec, we can make use of webhook validation to make sure that only one of the 2 is set.

To ensure, similar check occurs in namespace lables, we have to break the prune "strategy" and "value" into 2 labels.


```suggestion
	pruneStrategy                   = "operator.tekton.dev/prune-strategy"
	pruneKeep                         = operator.tekton.dev/prune.keep
	pruneKeepSince               = "operator.tekton.dev/prune.keepsince"
	
```

Our user will have to set either `keep` or `keepSince` as the value for `operator.tekton.dev/prune-strategy` .

the the controller will honor only `operator.tekton.dev/prune.keep` if `operator.tekton.dev/prune-strategy == keep` 

and
the the controller will honor only `operator.tekton.dev/prune.keepSince` if `operator.tekton.dev/prune-strategy == keepSince`

pkg/reconciler/common/prune.go
```diff
@@ -170,5 +301,39 @@ func checkAndDelete(k kubernetes.Interface, ctx context.Context, targetNamespace
...
+func (pruner *Pruner) removedFromTektonConfig(prune v1alpha1.Prune) bool {
+	if len(prune.Resources) == 0 || prune.Schedule == "" {
+		return true
+	}
+	return false
+}
+
+func (pruner *Pruner) removedNSAnnotation() ([]string, bool) {
+	var namespaces []string
+	cronJobs, err := pruner.kc.BatchV1beta1().CronJobs("").List(pruner.context, v1.ListOptions{})
```
nikhil-thomas, line 316: let us add a label to all the cronJobs we create eg: `tektonconfig.operator.tekton.dev/pruner: true`

then this cronJob query can be improved by passing the label trough listOptions

pkg/reconciler/common/prune.go
```diff
@@ -170,5 +301,39 @@ func checkAndDelete(k kubernetes.Interface, ctx context.Context, targetNamespace
...
+	if err != nil {
+		return nil, false
+	}
```
nikhil-thomas, line 319: why don't we return error..  if no cronjobs are returned, we can check for  errorNotFound and return nil. [example](https://github.com/tektoncd/operator/blob/a3d280298cf08d090da741e2f6d0ee17b09e9d5f/pkg/reconciler/shared/tektonconfig/tektonconfig.go#L125)

pkg/reconciler/common/prune.go
```diff
@@ -170,5 +301,39 @@ func checkAndDelete(k kubernetes.Interface, ctx context.Context, targetNamespace
...
+	for _, cronjob := range cronJobs.Items {
+		if strings.HasPrefix(cronjob.Name, CronName) {
+			name := strings.TrimPrefix(cronjob.Name, CronName+"-")
```
nikhil-thomas, line 322: if we have already shorlisted the cronjobs using a label as mentioned in the comment above, this logic here will become, ns = cronjob.Namespace.

pkg/reconciler/common/prune.go
```diff
@@ -170,5 +301,39 @@ func checkAndDelete(k kubernetes.Interface, ctx context.Context, targetNamespace
...
+			if _, ok := ano[pruneSchedule]; !ok || ano[pruneSkip] == "true" {
+				namespaces = append(namespaces, name)
+			}
+		}
+	}
+
+	if len(namespaces) > 0 {
+		return namespaces, true
+	}
+	return nil, false
```
nikhil-thomas, line 337: could we make this   
```suggestion
	return namespaces, nil
```

pkg/reconciler/common/prune.go
```diff
@@ -21,75 +21,199 @@ import (
...
+		containers = append(containers, jobContainers...)
+	}
+	return containers
+}
+
+func (pruner *Pruner) createJobs(nsConfig *pruneConfigPerNS, ns string) []corev1.Container {
+	var containers []corev1.Container
+
+	cmdArgs := pruneCommand(nsConfig, ns)
+	jobName := SimpleNameGenerator.RestrictLengthWithRandomSuffix("pruner-tkn-" + ns)
```
nikhil-thomas, line 203: let us avoid the `namespace` suffix. it is redundant data and it doesn't help us with any functional benefits.

Im assuming that this was done to check a cronjob was created by this pruner logic. For that, let us add a label on the CornJob (example `tektonconfig.operator.tekton.dev/pruner: true`) 

This will be more readable. It will also help us filter our query for cronJobs from allnamespaces.

nikhil-thomas: In addition, we will create only 1 pruner CronJob in any given namespace.

pkg/reconciler/common/prune.go
```diff
@@ -21,75 +21,199 @@ import (
...
 )
 
 const (
-	tektonSA         = "tekton-pipelines-controller"
-	CronName         = "resource-pruner"
-	JobsTKNImageName = "IMAGE_JOB_PRUNER_TKN"
-	ownerAPIVer      = "operator.tekton.dev/v1alpha1"
-	ownerKind        = "TektonConfig"
+	tektonSA                    = "tekton-pipelines-controller"
+	CronName                    = "resource-pruner"
```
nikhil-thomas, line 38: let us make this `tektoncd-resorce-pruner`

pkg/reconciler/common/prune.go
```diff
@@ -21,75 +21,199 @@ import (
...
+			pc.config.KeepSince = defaultPruneConfig.KeepSince
+		}
+		if nsAnnotations[pruneKeep] != "" {
+			keep, _ := strconv.Atoi(nsAnnotations[pruneKeep])
+			uintKeep := uint(keep)
+			pc.config.Keep = &uintKeep
+			pc.config.KeepSince = nil
+		} else if defaultPruneConfig.Keep != nil && nsAnnotations[pruneKeepSince] == "" {
+			pc.config.Keep = defaultPruneConfig.Keep
+		}
```
nikhil-thomas, line 171: we have to add `pruneStrategy` logic here
as mentioned in  https://github.com/tektoncd/operator/pull/410/files#r712087264

pkg/reconciler/common/prune.go
```diff
@@ -98,11 +222,11 @@ func createCronJob(k kubernetes.Interface, ctx context.Context, pru v1alpha1.Pru
 			APIVersion: "batch/v1beta1",
 		},
 		ObjectMeta: v1.ObjectMeta{
-			Name:            CronName,
+			Name:            CronName + "-" + targetNs,
```
nikhil-thomas, line 225: https://github.com/tektoncd/operator/pull/410/files#r712104343

pkg/apis/operator/v1alpha1/tektonconfig_validation.go
pradeepitm12: Yeh, it makes sense, Thank You

pkg/reconciler/common/prune.go
pradeepitm12: This could be done. I am only thinking of updates, is there any chance the cronjob name change will create any problem with the updates?

pkg/reconciler/common/prune.go
pradeepitm12: Yes, looks good. Will update it.

pkg/apis/operator/v1alpha1/tektonconfig_validation.go
vdemeester: This check is still needed isn't it ? if we specify both should it fail or ?

pkg/reconciler/common/prune.go
vdemeester: Do we put annotation on the cronjob we create ? If yes, then we could/should probably not rely on the name, and thus changing it now wouldn't be a problem.
That said, where is this name use : inside `openshift-pipelines` (then it's fine to not have tektoncd prefix I would guess), or in any annotated namespace (then, yes, we should prefix it because we are not pruning all resources, just tektoncd ones).

vdemeester (commented):
- Is `keep: 2` a default we will/would ship ? because if yes, it might be a tidy bit too small.
- `keep` vs `keepSince` : I tend to agree that initially both "should" not be used together, but it could make sense though : I want to keep 100 pipelineruns and those since 4 weeks, and clean those who do not fit – If I have less than 100 in 4 weeks, I clean those after 4 weeks, if I have more than 100 in 4 weeks, I clean the 100 ; or the opposite, we would have to decide there.

pkg/reconciler/shared/tektonconfig/controller.go
```diff
@@ -18,6 +18,8 @@ package tektonconfig
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/types"
+	namespaceinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/namespace"
```
vdemeester, line 22: Those imports are at the wrong place 😝

vdemeester (commented):
I think we should remove the `operator.tekton.dev/prune.resources.pipelineNames` and `operator.tekton.dev/prune.resources.taskNames` from this feature / PR for now. 

- The feature is not present in the cluster-wide CR
- It doesn't scale well, and we could probably give the user a better experience. On top of my head, we go with an annotation or a label, that if present on a TaskRun/PipelineRun, we would skip it — something like `operator.tekton.dev/prune.ignore` or something similar, and because we propagate annotation from pipeline to pipelinerun or task to taskrun, a Task or Pipeline could have this annotation too, to make all execution of it not be pruned.

pkg/apis/operator/v1alpha1/tektonconfig_validation.go
nikhil-thomas: if both are specified, it will return a validation error. 
or we can split this into
```
p.strategy = keep/keepSince
p.keep
p.keepSince
```

either one of p.keep and p.keepSince, could be used depending on p.strategy.
we are following a similar approach in overides for the same using namespace annotations.

vdemeester: In the code, I don't see that check — that it will fail if both are present in `Tektonconfig` CR.

vdemeester: On the "strategy", what if, someday, we want to allow both ? we would add a new strategy ?

pradeepitm12: >In the code, I don't see that check — that it will fail if both are present in Tektonconfig CR.
I am addressing the review comments and it will be taken care in the updated pr.

pradeepitm12: > On the "strategy", what if, someday, we want to allow both? we would add a new strategy ?

Let's not introduce strategy at the tektonconfig level.

pkg/reconciler/common/prune.go
pradeepitm12: earlier label/annotation was not there. 
We will be adding a label this time. https://github.com/tektoncd/operator/pull/410#discussion_r712104343

pkg/reconciler/common/prune.go
pradeepitm12: Good catch, changes will appear ones pr is updated.

pkg/reconciler/common/prune.go
pradeepitm12: Sure

pkg/reconciler/common/prune.go
pradeepitm12: Sure, I can handle it in this pr.

pkg/reconciler/shared/tektonconfig/controller.go
```diff
@@ -95,3 +101,14 @@ func NewExtensibleController(generator common.ExtensionGenerator) injection.Cont
...
 	}
 }
+
+// enqueueCustomName adds an event with name `config` in work queue so that
+// whenever a namespace event occurs, the TektonConfig reconciler get triggered.
+// This is required because we want to get our TektonConfig reconciler triggered
+// for already existing and new namespaces, without manual intervention like adding
+// a label/annotation on namespace to make it manageable by Tekton controller.
+func enqueueCustomName(impl *controller.Impl, name string) func(obj interface{}) {
+	return func(obj interface{}) {
```
sm43, line 111: previously we were filtering out namespace kube/openshift-
```
if err == nil && !nsRegex.MatchString(object.GetName()) {
			impl.EnqueueKey(types.NamespacedName{Namespace: "", Name: name})
		}
```
do we don't need to now?

pkg/reconciler/common/prune.go
```diff
@@ -122,9 +246,9 @@ func createCronJob(k kubernetes.Interface, ctx context.Context, pru v1alpha1.Pru
 		},
 	}
 
-	if _, err := k.BatchV1beta1().CronJobs(targetNs).Create(ctx, cj, v1.CreateOptions{}); err != nil {
+	if _, err := pruner.kc.BatchV1beta1().CronJobs(targetNs).Create(pruner.context, cj, v1.CreateOptions{}); err != nil {
```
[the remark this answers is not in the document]
pradeepitm12: Thanks for the reference.

pradeepitm12:
@vdemeester, I have removed the related resources part from this pr. This is good to review now.

pkg/reconciler/shared/tektonconfig/controller.go
sm43: previously, it was as below. we use to filter out namespaces `kube|openshift-` and wouldn't trigger on events on those namespaces
but now in the pr it the filter is removed so just wanna know do we need to watch event on all namespaces? 
```
func enqueueCustomName(impl *controller.Impl, name string) func(obj interface{}) {
	return func(obj interface{}) {
		object, err := kmeta.DeletionHandlingAccessor(obj)
		if err == nil && !nsRegex.MatchString(object.GetName()) {
			impl.EnqueueKey(types.NamespacedName{Namespace: "", Name: name})
		}
	}
}
```

pkg/reconciler/common/prune.go
```diff
@@ -133,35 +279,25 @@ func createCronJob(k kubernetes.Interface, ctx context.Context, pru v1alpha1.Pru
...
 		cmdArgs = cmdArgs + cmd
 	}
 	return cmdArgs
 }
 
-func checkAndDelete(k kubernetes.Interface, ctx context.Context, targetNamespace string) error {
-	if _, err := k.BatchV1beta1().CronJobs(targetNamespace).Get(ctx, CronName, v1.GetOptions{}); err != nil {
+func (pruner *Pruner) checkAndDeleteCron(ctx context.Context, ns string) error {
+	if _, err := pruner.kc.BatchV1beta1().CronJobs(ns).Get(ctx, CronName+"-"+ns, v1.GetOptions{}); err != nil {
 		if strings.Contains(err.Error(), "not found") {
```
sm43, line 301: should we use `apierrs.IsNotFound` to check not found err here?

pkg/reconciler/common/prune.go
```diff
@@ -122,9 +268,9 @@ func createCronJob(k kubernetes.Interface, ctx context.Context, pru v1alpha1.Pru
 		},
 	}
 
-	if _, err := k.BatchV1beta1().CronJobs(targetNs).Create(ctx, cj, v1.CreateOptions{}); err != nil {
+	if _, err := pruner.kc.BatchV1beta1().CronJobs(targetNs).Create(ctx, cj, v1.CreateOptions{}); err != nil {
 		if strings.Contains(err.Error(), "already exists") {
-			if _, err := k.BatchV1beta1().CronJobs(targetNs).Update(ctx, cj, v1.UpdateOptions{}); err != nil {
+			if _, err := pruner.kc.BatchV1beta1().CronJobs(targetNs).Update(ctx, cj, v1.UpdateOptions{}); err != nil {
```
sm43, line 273: cron job are in v1 from `k8s v1.21` should we switch the api to v1?

pkg/reconciler/common/prune.go
```diff
@@ -38,58 +41,200 @@ const (
...
+	}
+
+	tknImageFromEnv := os.Getenv(JobsTKNImageName)
+	if tknImageFromEnv == "" {
 		return fmt.Errorf("%s environment variable not found", JobsTKNImageName)
 	}
-	pru := tC.Spec.Pruner
+	pruner.tknImage = tknImageFromEnv
 	logger := logging.FromContext(ctx)
 	ownerRef := v1.OwnerReference{
```
sm43, line 94: here we can use 
```
ownerRef := *metav1.NewControllerRef(config, config.GetGroupVersionKind())
```

pkg/reconciler/shared/tektonconfig/controller.go
pradeepitm12: @sm43 Thanks for this catch. We need it. In multiple updates this got changed :+1:

pkg/reconciler/common/prune.go
pradeepitm12: Yes, I am on a separate pr for this.

pkg/reconciler/common/prune.go
pradeepitm12: Yes, we discussed to add it in iterative pr but I think we can update it here itself.

vdemeester (approved):

pradeepitm12:
> /hold cancel @pradeepitm12 needs a rebase

rebase Done

PR #412, opened 2021-09-20
Option to disable the default creation of RBAC resources

savitaashture:
# Changes

This PR addresses providing way to disable auto creation of RBAC resources at cluster as well as namespace level
**At cluster level**
* Set `disable-rbac-clusterwide: true` in **TektonConfig** CR

**At namespace level**
* Set label on namespace `oc label ns <ns-name> openshift-pipelines.tekton.dev/disable-rbac-injection=true`

/cc @nikhil-thomas @vdemeester @sm43 

# Submitter Checklist

review them:

# Release Notes

```release-note
Admin can disable auto creation of RBAC resources (ServiceAccount, RoleBinding, SCCRoleBinding, CABundlesConfigMap and add Subjects to openshift-pipelines-clusterinterceptors ClusterRoleBinding).
```


pkg/apis/operator/v1alpha1/tektonconfig_types.go
```diff
@@ -89,6 +89,9 @@ type TektonConfigSpec struct {
 	// Dashboard holds the customizable options for dashboards component
 	// +optional
 	Dashboard Dashboard `json:"dashboard,omitempty"`
+	// DisableRbacClusterWide holds the option to disable/enable creation of RBAC resources
+	// +optional
+	DisableRbacClusterWide bool `json:"disable-rbac-clusterwide,omitempty"`
```
nikhil-thomas, line 94: how about

```suggestion
	AutoConfigDefaultSA bool `json:"disable-rbac-clusterwide,omitempty"`
```
true/false

cluster wide is known as all other options here are clusterwide. And `RbacClusterwide` sounds a bit developer oriented.

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -42,23 +42,25 @@ const (
 	trustedCABundleConfigMap = "config-trusted-cabundle"
 	clusterInterceptors      = "openshift-pipelines-clusterinterceptors"
 	namespaceVersionLabel    = "openshift-pipelines.tekton.dev/namespace-reconcile-version"
+	disableRbacInjection     = "openshift-pipelines.tekton.dev/disable-rbac-injection"
```
nikhil-thomas, line 45: usually, double negatives could be confusing eg: `someting-disable: false` 
ie, disableRbac could be read as "disable doing something to do with RBAC" or there is an already existing entity which is disabling RBAC and this flag disbleRBAC will disable that.

if we can make this a neutral staement which toggles between true and false then it will be easier for our end user to think about it.

eg: <funciton-of-feature>: true/false

pkg/apis/operator/v1alpha1/tektonconfig_types.go
savitaashture: if we use `ConfigurePipelineServiceAccount` it indicates only for ServiceAccount but we are disabling other resources also along with `ServiceAccount` 

So how about using this 
`ConfigureRbacResources` :thinking:

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -42,23 +42,25 @@ const (
...
-	version           string
+	kubeClientSet          kubernetes.Interface
+	operatorClientSet      clientset.Interface
+	manifest               mf.Manifest
+	ownerRef               metav1.OwnerReference
+	version                string
+	disableRbacClusterWide bool
 }
 
 func (r *rbac) cleanUp(ctx context.Context) error {
```
sm43, line 60: should we add cleanup for injection label too? like too remove the label when uninstalling from ns

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -81,25 +83,84 @@ func (r *rbac) createResources(ctx context.Context) error {
...
+				} else {
+					rbacNamespaces = append(rbacNamespaces, n)
+					continue
+				}
+			} else {
+				rbacNamespaces = append(rbacNamespaces, n)
+				continue
+			}
+		}
+	}
```
nikhil-thomas, line 131: instead of nesting the checks, if we can make it 2 different checks, then it might make this easy to maitain in the future

how about

```
if disableRbacInjection explicitly set:
    if true
        add to disable list
    else
        add to enable list
    continue # as the label has precedence we don't have to check the value of config spec if label is set
    
if  r.disableRbacClusterWide = true
    add namespace to disablelist
else
    add namespace to enable list
```

pkg/apis/operator/v1alpha1/tektonconfig_types.go
vdemeester: If we change the field, we might want to change the "json" field as well.

vdemeester: Also, we should probably discuss the naming in the [enhancement review](https://github.com/openshift-pipelines/enhancements/pull/3/files)

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -81,25 +83,84 @@ func (r *rbac) createResources(ctx context.Context) error {
...
+	for _, n := range allNamespaces.Items {
 		if ignore := nsRegex.MatchString(n.GetName()); ignore {
 			continue
 		}
-		rbacNamespaces = append(rbacNamespaces, n)
+		v, ok := n.GetLabels()[disableRbacInjection]
+		if r.disableRbacClusterWide {
+			if ok {
+				if v == "false" || v == "" {
+					rbacNamespaces = append(rbacNamespaces, n)
```
sm43, line 105: previously, after creating the rbac resources in a `ns` we use to add `openshift-pipelines.tekton.dev/namespace-reconcile-version = <version>` label and in next reconcilation if ns has label then we use to ignore it
unless if it is an upgrade then the check will fail and we recreate rbac 

Currently, it is going through all ns each time right? do we not need to check the label
if the label exist then we can ignore going though that ns for rbac creation

pkg/reconciler/openshift/tektonconfig/rbac.go
savitaashture: Not sure whether we can clean the label as part of Operator uninstall as label is injected by user and not by Operator so i am thinking let user remove it based on their use case
But i may be wrong in considering end user scenario

/cc @ppitonak @vdemeester

pkg/apis/operator/v1alpha1/tektonconfig_types.go
nikhil-thomas: > we should probably discuss the naming in the enhancement review

absolutely. i missed that. 
i shall add a comment on this thread, if it doesn't resolve here, shortly, then we shall take it up there 🧑‍💻

pkg/reconciler/openshift/tektonconfig/rbac.go
sm43: ah sry I meant while deleting `TektonConfig` 😅 .. while uninstalling it would be kinda difficult to do this

pkg/apis/operator/v1alpha1/tektonconfig_types.go
nikhil-thomas: > So how about using this
ConfigureRbacResources 🤔

i agree with you, about the `SA` part. 

one comment on `ConfigureRbacResources` is that it could be read as, configuring RBAC by operator for the fucntioning of pipelines, triggers etc (rbac that comes with the components).

`ConfigureDefaultSA` could have been another name, but the `default` sa itself could make this confusing. 😄

nikhil-thomas: I think `ConfigureRBACResources` is good enough 👍 
with proper documentation, this name should work fine.

pkg/reconciler/openshift/tektonconfig/rbac.go
savitaashture: >and in next reconcilation if ns has label then we use to ignore it
unless if it is an upgrade then the check will fail and we recreate rbac

right but for this scenario everytime user update namespace label and in that case if we list namespace based on `openshift-pipelines.tekton.dev/namespace-reconcile-version = <version>` then while updating label on namespace it was never coming to this logic

pkg/reconciler/openshift/tektonconfig/rbac.go
savitaashture: even while deleting `TektonConfig` it won't do anything with labels on namespace because labels are set by User

nikhil-thomas: > so i am thinking let user remove it based on their use case

i think, this will be the expected UX from a user side.
as this label is not added by the controller, let us not touch it from the controller.

usecase: if a user wants to uninstall reinstall, the user (admin) will not want to relabel all the namespaces, again. The user might expect the re-installation to come back to the state based on already existing user supplied labels.

pkg/reconciler/openshift/tektonconfig/rbac.go
savitaashture: Yup updated :+1:

pkg/reconciler/openshift/tektonconfig/rbac.go
nikhil-thomas: i think if we split the namespace list query based on teh clusterWide spec, the list of namespaces will be much shorter.
This could fasten the process a lot as practically we will be interested in on only a short subset of all namespaces in any cluster. Let me try to come up with an example.

pkg/apis/operator/v1alpha1/tektonconfig_types.go
vdemeester: > i shall add a comment on this thread, if it doesn't resolve here, shortly, then we shall take it up there technologist

What do you mean ?

nikhil-thomas: if we agree on the `ConfigureRBACResources` then let us update that in the enhancement and we shall resolve this. :+1:

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -65,10 +65,11 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
 	}
```
sm43, line 66: should we add the param in TektonConfig CR with default value? similar to other defaults

vdemeester (approved):

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -75,35 +83,94 @@ func (r *rbac) cleanUp(ctx context.Context) error {
...
-	// `openshift-pipelines.tekton.dev/namespace-ready: <release-version>`
+	// `openshift-pipelines.tekton.dev/namespace-reconcile-version: <release-version>`
 	namespaces, err := r.kubeClientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{
 		LabelSelector: fmt.Sprintf("%s != %s", namespaceVersionLabel, r.version),
 	})
 	if err != nil {
 		return err
 	}
 
+	namespacesWithVersionLabel, err := r.kubeClientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{
```
sm43, line 122: why do we require this call? 🤔

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -65,10 +65,11 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
...
-		operatorClientSet: oe.operatorClientSet,
-		ownerRef:          configOwnerRef(tc),
-		version:           os.Getenv(versionKey),
+		kubeClientSet:                    oe.kubeClientSet,
+		operatorClientSet:                oe.operatorClientSet,
+		version:                          os.Getenv(versionKey),
+		disableRbacResourcesForOpenshift: config.Spec.Params,
+		tektonConfig:                     config,
 	}
 	return r.createResources(ctx)
```
sm43, line 74: what if here we check if the `createRbacResources` param is true or not?
if false 
  we get the name of installer set and delete it
if true
  then create an installer set and pass the owner ref of installer set to the create func
  
 ---- 
 inside it as we just need to update all the func 
 if resource exist then update the resource the owner ref  
 
 as it is an upgrade we will get all namespace in the existing `version != 1.6` call

pkg/apis/operator/v1alpha1/tektonconfig_types.go
```diff
@@ -97,6 +100,10 @@ type TektonConfigStatus struct {
 	// The version of the installed release
 	// +optional
 	Version string `json:"version,omitempty"`
+
+	// The current installer set name
+	// +optional
+	TektonInstallerSet map[string]string `json:"tektonInstallerSet,omitempty"`
```
sm43, line 106:
```suggestion
	TektonInstallerSet map[string]string `json:"tektonInstallerSets,omitempty"`
```

pkg/reconciler/openshift/tektonconfig/rbac.go
savitaashture: For the first time when user install OSP it adds `"openshift-pipelines.tekton.dev/namespace-reconcile-version"` to all namespace except `^(openshift|kube)-` 

and later when user update TektonConfig CR for `createRbacResources` its comes to rbac.go and don't go to any namespaces because this check https://github.com/tektoncd/operator/blob/main/pkg/reconciler/openshift/tektonconfig/rbac.go#L84-L85 

so to avoid that added another call to fetch all those namespaces which have `openshift-pipelines.tekton.dev/namespace-reconcile-version` to avoid looping over the **all Namespaces**

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -103,6 +131,117 @@ func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonC
...
+			Annotations: map[string]string{
+				releaseVersionKey:  releaseVersion,
+				targetNamespaceKey: tc.Spec.TargetNamespace,
+			},
+			OwnerReferences: []metav1.OwnerReference{ownerRef},
+		},
+	}
+}
+
+func (r *rbac) deleteInstallerSet(ctx context.Context, tc *v1alpha1.TektonConfig, component string) error {
```
nikhil-thomas, line 180: let us move this function belongs in the rbac.go, as it is a method on the rbac struct

pkg/apis/operator/v1alpha1/tektonconfig_types.go
```diff
@@ -97,6 +100,10 @@ type TektonConfigStatus struct {
 	// The version of the installed release
 	// +optional
 	Version string `json:"version,omitempty"`
+
+	// The current installer set name
+	// +optional
+	TektonInstallerSet map[string]string `json:"tektonInstallerSets,omitempty"`
```
nikhil-thomas, line 106: In my opinion, a list of names of TektonInstallerSets directly managed by TektonConfig is of low value. 
In addition, as  rbac is a platform specific operation let us keep the effects away from this common area.

As there will be only one InstallerSet for RBAC, we can keep the name constant and we do not have to record that as a status field.

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -67,10 +68,37 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
 	r := rbac{
 		kubeClientSet:     oe.kubeClientSet,
 		operatorClientSet: oe.operatorClientSet,
-		ownerRef:          configOwnerRef(tc),
 		version:           os.Getenv(versionKey),
+		tektonConfig:      config,
 	}
-	return r.createResources(ctx)
+
+	var createResourceError error
```
nikhil-thomas, line 75: could we move this to the top or to the rbac package.
let us make this more descriptive/meaningful as well.

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -67,10 +68,37 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
...
+					return err
+				}
+
+			} else {
+				createResourceError = r.createResources(ctx)
+			}
+		} else {
+			createResourceError = r.createResources(ctx)
+		}
+		continue
```
nikhil-thomas, line 98: let us remove this `else` and `coninue` is , 
this means that for each item in the params list which is not `createResources`, `createResources` will be called.

nikhil-thomas: the absence of the param is already handled above in 
```
if len(config.Spec.Params) == 0 {
		createResourceError = r.createResources(ctx)
	}
```

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -67,10 +68,37 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
...
+		// then disable auto creation of RBAC resources by deleting installerSet
+		if v.Name == "createRbacResource" {
+			if v.Value == "false" {
+				if err := r.deleteInstallerSet(ctx, r.tektonConfig, componentName); err != nil {
+					return err
+				}
+				// remove openshift-pipelines.tekton.dev/namespace-reconcile-version label from namespaces while deleting RBAC resources.
+				if err := r.cleanUp(ctx); err != nil {
+					return err
+				}
```
nikhil-thomas, line 90: do we need 2 separate calls here.
shall we wrap both of them in a `cleanupResources` method in rbac.go (like the `createResources` method)

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -67,10 +68,37 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
...
+
+			} else {
+				createResourceError = r.createResources(ctx)
```
nikhil-thomas, line 93: we could directly return the error and return a `nil` at the end after the for loop

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -103,6 +131,117 @@ func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonC
...
+func configOwnerRef(tc v1alpha1.TektonInstallerSet) metav1.OwnerReference {
+	return *metav1.NewControllerRef(&tc, tc.GetGroupVersionKind())
+}
+
+func createInstallerSet(ctx context.Context, oc versioned.Interface, ta *v1alpha1.TektonConfig,
+	releaseVersion, component, installerSetName string) error {
+
+	is := makeInstallerSet(ta, installerSetName, releaseVersion)
+
+	createdIs, err := oc.OperatorV1alpha1().TektonInstallerSets().
```
nikhil-thomas, line 143: let us move all `installerSet` related functionality into a separate file within this package (or to  the common package if that makes more sense). 

At present, we are calling methods from rbac.go from this file then later calling functions in this file from rbac.go.

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -41,6 +42,11 @@ const (
 	trustedCABundleConfigMap = "config-trusted-cabundle"
 	clusterInterceptors      = "openshift-pipelines-clusterinterceptors"
 	namespaceVersionLabel    = "openshift-pipelines.tekton.dev/namespace-reconcile-version"
+	createdByKey             = "operator.tekton.dev/created-by"
+	createdByValue           = "RBAC"
+	releaseVersionKey        = "operator.tekton.dev/release-version"
+	targetNamespaceKey       = "operator.tekton.dev/target-namespace"
+	componentName            = "rbacTaskInstallerSet"
```
nikhil-thomas, line 49: As the `kind` will be `TektonInstallerSet`, can we shorten this name to `rbac` or `rhosp-rbac` or `tekton-rbac`

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -103,6 +131,117 @@ func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonC
...
+	ownerRef := *metav1.NewControllerRef(tc, tc.GetGroupVersionKind())
+	return &v1alpha1.TektonInstallerSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Labels: map[string]string{
+				createdByKey: createdByValue,
+			},
+			Annotations: map[string]string{
+				releaseVersionKey:  releaseVersion,
+				targetNamespaceKey: tc.Spec.TargetNamespace,
```
nikhil-thomas, line 173: `targetNamespaceKey` could be avoided. This was there in the other InstallerSet without much functional usage at present.

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -104,6 +111,25 @@ func (r *rbac) createResources(ctx context.Context) error {
 		return nil
 	}
 
+	exist, err := checkIfInstallerSetExist(ctx, r.operatorClientSet, r.version, r.tektonConfig, componentName)
+	if err != nil {
+		return err
+	}
+	if !exist {
+		if err := createInstallerSet(ctx, r.operatorClientSet, r.tektonConfig,
+			r.version, componentName, "rbac-resources"); err != nil {
```
nikhil-thomas, line 120: why is 2 names necesary 1. the constant `componentName` and then `rbac-resources`.

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -104,6 +111,25 @@ func (r *rbac) createResources(ctx context.Context) error {
 		return nil
 	}
 
+	exist, err := checkIfInstallerSetExist(ctx, r.operatorClientSet, r.version, r.tektonConfig, componentName)
```
nikhil-thomas, line 114: can we do a direct TektonInstallerSets.Get() using the name `rbac-resources` to check the expected InstallerSet exists.

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -103,6 +131,117 @@ func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonC
...
+		}
+
+		// clear the name of installer set from TektonAddon status
+		delete(tc.Status.TektonInstallerSet, component)
+		_, err = r.operatorClientSet.OperatorV1alpha1().TektonConfigs().
+			UpdateStatus(ctx, tc, metav1.UpdateOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	}
```
nikhil-thomas, line 202: shall we use a simple TektonInstallerSet.Get() call with name `rbac-resources` to find the InstallerSet

pkg/apis/operator/v1alpha1/tektonconfig_types.go
savitaashture: Added status field to indicate creation of rbac resources are enabled or disabled

/cc @sm43

pkg/reconciler/openshift/tektonconfig/extension.go
savitaashture: As discussed with you handling this part based on flag

pkg/reconciler/openshift/tektonconfig/extension.go
savitaashture: kept `deleteInstallerSet` separate so that can be used later when someone add new installerSet for TektonConfig

/cc @pradeepitm12

pkg/apis/operator/v1alpha1/tektonconfig_types.go
sm43: for TektonPipelines, there were multiple installerSet created and keeping hardcoded name wasn't a good option
and status also works as an check whether we have enabled that part or not
like if the name of `installerSet` in status doesn't exist and `createRbacResource` is `false` then we can just save the delete api call which we would be doing on each reconcile

sm43: and if pruner is going to use installerSet as ownerRef then this would be visible in both platform
otherwise the field will not be visible on k8s

savitaashture:
@nikhil-thomas as per the discussion I have updated PR
PTAL

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -75,12 +83,39 @@ func (r *rbac) cleanUp(ctx context.Context) error {
 	return nil
 }
 
+func (r *rbac) setDefault() bool {
```
savitaashture, line 86: @sm43 added defaulting function to set default key and value for param
/cc @nikhil-thomas @vdemeester

pkg/reconciler/openshift/tektonconfig/common.go
```diff
@@ -0,0 +1,119 @@
+package tektonconfig
```
sm43, line 1: nit: license

pkg/reconciler/openshift/tektonconfig/common.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if compInstallerSet != "" {
+		// delete the installer set
+		err := oc.OperatorV1alpha1().TektonInstallerSets().
+			Delete(ctx, tc.Status.TektonInstallerSet[component], metav1.DeleteOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+
+		// clear the name of installer set from TektonAddon status
```
sm43, line 67:
```suggestion
		// clear the name of installer set from TektonConfig status
```

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -56,22 +56,44 @@ func (oe openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Tr
...
+		kubeClientSet:     oe.kubeClientSet,
+		operatorClientSet: oe.operatorClientSet,
+		version:           os.Getenv(versionKey),
+		tektonConfig:      config,
+	}
+
 	pipelineUpdated := openshiftPipeline.SetDefault(&config.Spec.Pipeline)
 	triggerUpdated := openshiftTrigger.SetDefault(&config.Spec.Trigger.TriggersProperties)
-	if pipelineUpdated || triggerUpdated {
+	if pipelineUpdated || triggerUpdated || r.setDefault() {
```
sm43, line 68: 👍🏽

savitaashture:
> /hold to address some minor comments sweat_smile also @nikhil-thomas ptal :)

@sm43 fixed review comments

vdemeester (approved):

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -103,6 +125,6 @@ func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonC
 }
 
 // configOwnerRef returns owner reference pointing to passed instance
-func configOwnerRef(tc v1alpha1.TektonComponent) metav1.OwnerReference {
-	return *metav1.NewControllerRef(tc, tc.GroupVersionKind())
+func configOwnerRef(tc v1alpha1.TektonInstallerSet) metav1.OwnerReference {
+	return *metav1.NewControllerRef(&tc, tc.GetGroupVersionKind())
```
nikhil-thomas, line 129: could you move this also to `pkg/reconciler/openshift/tektonconfig/common.go` as all other installerSet related functions are there.

PR #415, opened 2021-09-21
Set default value to TektonConfig object

### LINKED ISSUE #368: TektonConfig object gets created without any default values
Sometimes, webhook takes time to start and if the controller is started before it
it will create a `TektonConfig` object
but as webhook is not started the object will not have any default values.

```
  kind: TektonConfig
  metadata:
    name: config
  spec:
    addon: {}
    pipeline: {}
    profile: all
    pruner: {}

```

vinamra28:
# Changes

Currently when running the Tekton operator if the controller pod is up
and webhook pod is not ready the controller will create the TektonConfig
object and that TektonConfig object will not have any default values.

This PR add the default values to TektonConfig object whenever the CR
get's created by the Controller so that even if the webhook pod is not
ready the TektonConfig CR can still have the default values.

Closes #368

# Submitter Checklist

review them:

# Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/instance.go
+++ b/pkg/reconciler/shared/tektonconfig/instance.go
@@ -91,7 +91,7 @@ func (tc tektonConfig) ensureInstance(ctx context.Context) {
 			logger.Errorf("error getting Tektonconfig, Name: ", instance.GetName())
 			return false, nil
 		}
-		err = tc.createInstance()
+		err = tc.createInstance(ctx)
 		if err != nil {
 			//log error and retry
 			logger.Errorf("error creating Tektonconfig instance, Name: ", instance.GetName())
@@ -108,7 +108,7 @@ func (tc tektonConfig) ensureInstance(ctx context.Context) {
 	}
 }
 
-func (tc tektonConfig) createInstance() error {
+func (tc tektonConfig) createInstance(ctx context.Context) error {
 	tcCR := &v1alpha1.TektonConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: common.ConfigResourceName,
@@ -120,7 +120,8 @@ func (tc tektonConfig) createInstance() error {
 			},
 		},
 	}
+	tcCR.SetDefaults(ctx)
 	_, err := tc.operatorClientSet.OperatorV1alpha1().
-		TektonConfigs().Create(context.TODO(), tcCR, metav1.CreateOptions{})
+		TektonConfigs().Create(ctx, tcCR, metav1.CreateOptions{})
 	return err
 }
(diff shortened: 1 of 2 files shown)


pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -108,6 +108,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 		return nil
 	}
 
+	tc.SetDefaults(ctx)
```
sm43, line 111: @vdemeester when we pass the object through defaulting in reconciler, do we also update the object in etcd?

nikhil-thomas: i think the defaulter won't do it. however, the next update action down in the code/reconciler will get it into etcd. 
so your concern is right.

The change in this patch should be moved to [here](https://github.com/tektoncd/operator/blob/main/pkg/reconciler/shared/tektonconfig/instance.go#L122)

nikhil-thomas: in fact, we can do it in both places, it should have the same effect.

nikhil-thomas: im fine with merging this as it is.

vinamra28: addressed the comment btw

pkg/reconciler/shared/tektonconfig/instance.go
```diff
@@ -120,6 +120,7 @@ func (tc tektonConfig) createInstance() error {
 			},
 		},
 	}
+	tcCR.SetDefaults(context.TODO())
```
vdemeester, line 123: I think this should use `ctx`, meaning `createInstance` should take a `context.Context` as argument, otherwise we colud miss actual configured defaults

pkg/reconciler/shared/tektonconfig/tektonconfig.go
vdemeester: We have the same in `tektoncd/pipeline`, as @nikhil-thomas said, it won't be stored in `etcd` but ain't such a big deal.

pkg/reconciler/shared/tektonconfig/instance.go
vinamra28: also there is [line](https://github.com/tektoncd/operator/pull/415/files/9a80ec5b834597476f58b4c01477d0e3fac304b1#diff-5559ff56481d7f8604560c6ceb814063d2b6873416abaadb45e63bc07d1f9db8R125) where context.TODO is used. Shall I change that also?

vdemeester: I don't see the line but.. maybe, yeah ?

vinamra28: maybe it was moved with my commit. I have changed that with the `ctx` [here](https://github.com/tektoncd/operator/pull/415/files#diff-5559ff56481d7f8604560c6ceb814063d2b6873416abaadb45e63bc07d1f9db8L124)

vdemeester (approved):
/lgtm

PR #419, opened 2021-09-23
Add properties to configure Pipelines Metrics

khrm:
# Changes

# Submitter Checklist

review them:

# Release Notes


vdemeester (approved):

pkg/reconciler/openshift/tektonpipeline/extension.go
```diff
@@ -191,6 +191,23 @@ func SetDefault(pipeline *v1alpha1.Pipeline) bool {
 		updated = true
 	}
 
+	if pipeline.MetricsPipelinerunDurationType == "" {
```
sm43, line 194: are this openshift specific values? I see we are setting default values in webhook
do we need to again set in extension ?
are they different from k8s platform?

khrm: They were OpenShift specific values but I enabled them for k8s also. Now, I have removed it from here and values in both OpenShift and k8s will be the same. It will change the behaviour from the one we get in release.yaml but that's fine.

PR #421, opened 2021-09-24
Refactor TektonConfig Reconciler and Cleanup

sm43:
Fixes addon component deletion 

If clustertask and pipelines templates are disabled then
due to return statement it used to return early without
executing remaning code.

-----

 Updates TektonConfig Conditions and cleanup

This updates the tektonconfig conditions and removes
unnecessary code

----

 Removes deletion timestamp check in Autocreation of TektonConfig

If TektonConfig has a deletion stamp, it won't be deleted
unless the operator is up and running as the object has
finalizer. waiting for it to get deleted will stop operator
from running and object still won't be deleted.
hence removing the condition for checking deletion timestamp
and letting user to create a new instance of tektonconfig
after deletion.

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonaddon_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonaddon_types.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"reflect"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
@@ -63,6 +65,17 @@ type TektonAddonStatus struct {
 	// TektonInstallerSet created to install addons
 	// +optional
 	AddonsInstallerSet map[string]string `json:"installerSets,omitempty"`
+}
+
+// Addon defines the field to customize Addon component
+type Addon struct {
+	// Params is the list of params passed for Addon customization
+	// +optional
+	Params []Param `json:"params,omitempty"`
+}
+
+func (a Addon) IsEmpty() bool {
+	return reflect.DeepEqual(a, Addon{})
 }
 
 // TektonAddonsList contains a list of TektonAddon
--- a/pkg/apis/operator/v1alpha1/tektonaddon_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonaddon_types.go
@@ -17,8 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"reflect"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
@@ -75,7 +73,7 @@ type Addon struct {
 }
 
 func (a Addon) IsEmpty() bool {
-	return reflect.DeepEqual(a, Addon{})
+	return len(a.Params) == 0
 }
 
 // TektonAddonsList contains a list of TektonAddon
--- a/pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
+++ b/pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
@@ -21,107 +21,130 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var (
-	_ TektonComponentStatus = (*TektonConfigStatus)(nil)
+const (
+	PreInstall      apis.ConditionType = "PreInstall"
+	ComponentsReady apis.ConditionType = "ComponentsReady"
+	PostInstall     apis.ConditionType = "PostInstall"
+)
 
+var (
 	configCondSet = apis.NewLivingConditionSet(
-		DependenciesInstalled,
-		DeploymentsAvailable,
-		InstallSucceeded,
+		PreInstall,
+		ComponentsReady,
+		PostInstall,
 	)
 )
 
-// GroupVersionKind returns SchemeGroupVersion of a TektonConfig
-func (tp *TektonConfig) GroupVersionKind() schema.GroupVersionKind {
+func (tc *TektonConfig) GroupVersionKind() schema.GroupVersionKind {
 	return SchemeGroupVersion.WithKind(KindTektonConfig)
 }
 
-// GetCondition returns the current condition of a given condition type
-func (tps *TektonConfigStatus) GetCondition(t apis.ConditionType) *apis.Condition {
-	return configCondSet.Manage(tps).GetCondition(t)
+func (tc *TektonConfig) GetGroupVersionKind() schema.GroupVersionKind {
+	return SchemeGroupVersion.WithKind(KindTektonConfig)
 }
 
-// InitializeConditions initializes conditions of an TektonConfigStatus
-func (tps *TektonConfigStatus) InitializeConditions() {
-	configCondSet.Manage(tps).InitializeConditions()
+func (tcs *TektonConfigStatus) GetCondition(t apis.ConditionType) *apis.Condition {
+	return configCondSet.Manage(tcs).GetCondition(t)
 }
 
-// IsReady looks at the conditions returns true if they are all true.
-func (tps *TektonConfigStatus) IsReady() bool {
-	return configCondSet.Manage(tps).IsHappy()
+func (tcs *TektonConfigStatus) InitializeConditions() {
+	configCondSet.Manage(tcs).InitializeConditions()
 }
 
-// MarkInstallSucceeded marks the InstallationSucceeded status as true.
-func (tps *TektonConfigStatus) MarkInstallSucceeded() {
-	configCondSet.Manage(tps).MarkTrue(InstallSucceeded)
-	if tps.GetCondition(DependenciesInstalled).IsUnknown() {
-		// Assume deps are installed if we're not sure
-		tps.MarkDependenciesInstalled()
-	}
+func (tcs *TektonConfigStatus) IsReady() bool {
+	return configCondSet.Manage(tcs).IsHappy()
 }
 
-// MarkInstallFailed marks the InstallationSucceeded status as false with the given
-// message.
-func (tps *TektonConfigStatus) MarkInstallFailed(msg string) {
-	configCondSet.Manage(tps).MarkFalse(
-		InstallSucceeded,
+func (tcs *TektonConfigStatus) MarkPreInstallComplete() {
+	configCondSet.Manage(tcs).MarkTrue(PreInstall)
+}
+
+func (tcs *TektonConfigStatus) MarkComponentsReady() {
+	configCondSet.Manage(tcs).MarkTrue(ComponentsReady)
+}
+
+func (tcs *TektonConfigStatus) MarkPostInstallComplete() {
+	configCondSet.Manage(tcs).MarkTrue(PostInstall)
+}
+
+func (tcs *TektonConfigStatus) MarkNotReady(msg string) {
+	configCondSet.Manage(tcs).MarkFalse(
+		apis.ConditionReady,
 		"Error",
-		"Install failed with message: %s", msg)
+		"Ready: %s", msg)
 }
 
-// MarkDeploymentsAvailable marks the DeploymentsAvailable status as true.
-func (tps *TektonConfigStatus) MarkDeploymentsAvailable() {
-	configCondSet.Manage(tps).MarkTrue(DeploymentsAvailable)
-}
-
-// MarkDeploymentsNotReady marks the DeploymentsAvailable status as false and calls out
-// it's waiting for deployments.
-func (tps *TektonConfigStatus) MarkDeploymentsNotReady() {
-	configCondSet.Manage(tps).MarkFalse(
-		DeploymentsAvailable,
-		"NotReady",
-		"Waiting on deployments")
-}
-
-// MarkDependenciesInstalled marks the DependenciesInstalled status as true.
-func (tps *TektonConfigStatus) MarkDependenciesInstalled() {
-	configCondSet.Manage(tps).MarkTrue(DependenciesInstalled)
-}
-
-// MarkDependencyInstalling marks the DependenciesInstalled status as false with the
-// given message.
-func (tps *TektonConfigStatus) MarkDependencyInstalling(msg string) {
-	configCondSet.Manage(tps).MarkFalse(
-		DependenciesInstalled,
-		"Installing",
-		"Dependency installing: %s", msg)
-}
-
-// MarkDependencyMissing marks the DependenciesInstalled status as false with the
-// given message.
-func (tps *TektonConfigStatus) MarkDependencyMissing(msg string) {
-	configCondSet.Manage(tps).MarkFalse(
-		DependenciesInstalled,
+func (tcs *TektonConfigStatus) MarkPreInstallFailed(msg string) {
+	tcs.MarkNotReady("PreReconciliation failed")
+	configCondSet.Manage(tcs).MarkFalse(
+		PreInstall,
 		"Error",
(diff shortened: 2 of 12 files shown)


pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
```diff
@@ -21,107 +21,128 @@ import (
...
+)
 
+var (
 	configCondSet = apis.NewLivingConditionSet(
-		DependenciesInstalled,
-		DeploymentsAvailable,
-		InstallSucceeded,
+		PreReconciler,
+		ComponentsReady,
+		PostReconciler,
```
savitaashture, line 32: I assume these conditions are displayed on status and i think `PreReconciler` and `PostReconciler` words are more towards technical

May i know the reason behind removing old conditions

nikhil-thomas: > PreReconciler and PostReconciler words are more towards technical

I agree with this. It would be great if we could rename it with something that indicates the objective/end result of the action. It doesn't has to be precise. Even `PreInstall` `PostInstall` will be better.

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
```diff
@@ -21,107 +21,128 @@ import (
...
-	_ TektonComponentStatus = (*TektonConfigStatus)(nil)
+const (
+	ComponentsReady apis.ConditionType = "ComponentsReady"
+)
 
+var (
 	configCondSet = apis.NewLivingConditionSet(
-		DependenciesInstalled,
-		DeploymentsAvailable,
-		InstallSucceeded,
```
nikhil-thomas, line 30: shall we retain `InstallSucceeded` as the final condition. Then we could share that as the standard way to check whether all aspects of TektonConfig installation was completed successfully.

pkg/apis/operator/v1alpha1/tektonconfig_types.go
```diff
@@ -24,11 +24,6 @@ import (
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
 
-var (
-	_ TektonComponent     = (*TektonConfig)(nil)
-	_ TektonComponentSpec = (*TektonConfigSpec)(nil)
-)
-
```
nikhil-thomas, line 31: what makes TektonConfig not implement the TektonComponent Interface.
If TetkonConfig does not implement TektonComponent Interface, will that restrict us in some way in tests or in extensions.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -185,7 +185,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ta *v1alpha1.TektonAddon
 		}
 	} else {
 		// if disabled then delete the installer Set if exist
-		return r.deleteInstallerSet(ctx, ta, clusterTaskInstallerSet)
+		if err := r.deleteInstallerSet(ctx, ta, clusterTaskInstallerSet); err != nil {
+			return err
+		}
```
nikhil-thomas, line 190: nit: is this change really necessary.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -207,7 +209,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ta *v1alpha1.TektonAddon
 		}
 	} else {
 		// if disabled then delete the installer Set if exist
-		return r.deleteInstallerSet(ctx, ta, pipelinesTemplateInstallerSet)
+		if err := r.deleteInstallerSet(ctx, ta, pipelinesTemplateInstallerSet); err != nil {
+			return err
+		}
```
nikhil-thomas, line 214: same 🧑‍💻

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -52,13 +51,7 @@ func (oe kubernetesExtension) PostReconcile(ctx context.Context, comp v1alpha1.T
...
 
 	if configInstance.Spec.Profile == common.ProfileLite || configInstance.Spec.Profile == common.ProfileBasic {
-		err := extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
-		if err != nil {
-			if errors.IsNotFound(err) {
-				return nil
-			}
-			return err
-		}
+		return extension.TektonDashboardCRDelete(oe.operatorClientSet.OperatorV1alpha1().TektonDashboards(), common.DashboardResourceName)
```
nikhil-thomas, line 54: i think, here the code should ignore notFound error.

if initially the profile was `LITE` and then it was changed to `BASIC`, then we will get a `NotFound` error for delete Dashboard, as Dashboard is not provided in either of the profiles.

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
sm43: yeah `PreInstall` `PostInstall` will also be nice
also we do want to break down the component status in some way currently the components cond will show any error
in further interation we do want to find a better way for it

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
sm43: currently i haven't added any cond for pruner .,. so kept a common `componentready` for all ..
we can look into it more
what would `InstallSucceeded` say?

pkg/apis/operator/v1alpha1/tektonconfig_types.go
sm43: we can enable it back but with some changes I have added a TODO in lifecycle for it
once we refactor al components (dashboard, result ) we can find common things and update TektonComponent
otherwise they would break

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
sm43: yeah it is required, if clustertasks are disabled then rest of the code was not executed 
so we only return if there is some err while deleting

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
sm43: same as above 😄

pkg/reconciler/kubernetes/tektonconfig/extension.go
sm43: actually the not found case is already handled inside `TektonDashboardCRDelete` so we don't need to explicitly handle it :)

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
savitaashture: `InstallSucceeded` shows the success of Operator installation basically success of TektonConfig installation

pkg/reconciler/kubernetes/tektonconfig/extension.go
nikhil-thomas: oh that is wonderful 🧑‍💻 👍

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
nikhil-thomas: oh.. i see. it is about `not` returning nil. and not about returning error. 👍

pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -30,45 +28,26 @@ import (
...
-	mflogger := zapr.NewLogger(logger.Named("manifestival").Desugar())
-	manifest, err := mf.ManifestFrom(mf.Slice{}, mf.UseClient(mfclient), mf.UseLogger(mflogger))
-	if err != nil {
-		logger.Fatalw("error creating initial manifest", zap.Error(err))
-	}
-	ext := openshiftExtension{
+	return openshiftExtension{
 		operatorClientSet: operatorclient.Get(ctx),
 		kubeClientSet:     kubeclient.Get(ctx),
-		manifest:          manifest,
```
nikhil-thomas, line 62: 👍

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
nikhil-thomas: > currently i haven't added any cond for pruner .,. so kept a common componentready for all ..

ok.. copy that.
i recommend adding a final `InstallSucceed` so that QE and admins can use scripts for monitoring this one condition, no matter what how we add other intermediate conditions (or add more checks before setting `InstallSucceed: True`)

sm43: by default `Ready` condition is added by knative/pkg which gets true only when all other cond get true
so `InstallSucceeded` is diff from `Ready` ?

nikhil-thomas: Oh.. great 👍 forgot about that.
That is enough.

pkg/apis/operator/v1alpha1/tektonconfig_types.go
nikhil-thomas: could you add it as an issue in https://github.com/tektoncd/operator/issues

nikhil-thomas: my concern is that, if we don't have this checks, we might end up loosing uniformity in methods of components in the future. It is not a deal breaker, but i would like a reminder to check this later.

sm43:
/hold
to rename the conditions

pkg/apis/operator/v1alpha1/tektonaddon_types.go
```diff
@@ -65,6 +67,17 @@ type TektonAddonStatus struct {
...
 
+// Addon defines the field to customize Addon component
+type Addon struct {
+	// Params is the list of params passed for Addon customization
+	// +optional
+	Params []Param `json:"params,omitempty"`
+}
+
+func (a Addon) IsEmpty() bool {
+	return reflect.DeepEqual(a, Addon{})
```
vdemeester, line 78: Any reason to use `reflect` here instead of just `len(a.Params) == 0` ?
We should avoid using `reflect` when we can, it's kind-of costly for.. not always a valid reason to be.

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
vdemeester: Note that if we change what is written in the Status (the name of the condition, …), we need to validate other tools that are depending on this won't be broken (e.g. openshift console, …)

vdemeester: (because we are changing the API)

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle_test.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+    httc://www.apache.org/licenses/LICENSE-2.0
```
vdemeester, line 8: http

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
```diff
@@ -21,107 +21,128 @@ import (
...
-func (tp *TektonConfig) GroupVersionKind() schema.GroupVersionKind {
+func (tc *TektonConfig) GroupVersionKind() schema.GroupVersionKind {
 	return SchemeGroupVersion.WithKind(KindTektonConfig)
 }
 
-// GetCondition returns the current condition of a given condition type
-func (tps *TektonConfigStatus) GetCondition(t apis.ConditionType) *apis.Condition {
-	return configCondSet.Manage(tps).GetCondition(t)
+func (tc *TektonConfig) GetGroupVersionKind() schema.GroupVersionKind {
+	return SchemeGroupVersion.WithKind(KindTektonConfig)
```
vdemeester, line 41: Any reason to have both ?

pkg/apis/operator/v1alpha1/tektonconfig_types.go
```diff
@@ -95,13 +90,13 @@ type TektonConfigSpec struct {
...
-	// The version of the installed release
+	// The profile installed
 	// +optional
-	Version string `json:"version,omitempty"`
+	Profile string `json:"profile,omitempty"`
 
-	// The url links of the manifests, separated by comma
+	// The version of the installed release
 	// +optional
-	Manifests []string `json:"manifests,omitempty"`
```
vdemeester, line 104: Are we breaking the status API by removing this ?

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
sm43: yeah in this case we are chaning the api
we have removed the manifest field from status
and updated conditions
what tools do we have to validate? 🤔

pkg/apis/operator/v1alpha1/tektonaddon_types.go
sm43: yeah len of param would work... 
@vdemeester how about using `equality.Semantic.DeepEqual` not in this case but [here](https://github.com/tektoncd/operator/blob/039009587fe6bd89f975455a98c39113530f097e/pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go#L61) is that okay compared to reflect?

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
sm43: `GroupVersionKind` can be removed  now.. will update !

pkg/apis/operator/v1alpha1/tektonconfig_types.go
sm43: yes

sm43: previously it was nil though.. we didn't use to add anything to that field in TektonConfig

pkg/apis/operator/v1alpha1/tektonconfig_lifecycle.go
sm43: in generated code and internally in knative/pkg `` is used 
`GetGroupVersionKind` is defined so that we can do
```
tektonTriggerinformer.Get(ctx).Informer().AddEventHandler(cache.FilteringResourceEventHandler{
			FilterFunc: controller.FilterController(&v1alpha1.TektonConfig{}),
			Handler:    controller.HandleAll(impl.EnqueueControllerOf),
		})
```
so lets keep `GroupVersionKind` also 
i will take a deep look and then remove

pkg/apis/operator/v1alpha1/tektonaddon_types.go
vdemeester: Maybe yes ? My point is just that we should aim to use as less as possible `reflect` and those things.. but when we don't have the choice .. 😛

pkg/apis/operator/v1alpha1/tektonconfig_types.go
sm43: https://github.com/tektoncd/operator/issues/424
created an issue

PR #426, opened 2021-09-24
Read component version from ConfigMap

### LINKED ISSUE #413: Read component version from configmap instead of CRD label
In the recent refactoring while starting the controller we read the version of component from CRD label
but we also have a config map added to all components which has the version of that component.
The configmap was added so that client like tkn can read the version of component.

In Operator we can do the same instead of reading from CRD we should read from configmap

from controller we make a call https://github.com/tektoncd/operator/blob/a3d280298cf08d090da741e2f6d0ee17b09e9d5f/pkg/reconciler/kubernetes/tektonpipeline/controller.go#L70

and the func is implemented 

https://github.com/tektoncd/operator/blob/a3d280298cf08d090da741e2f6d0ee17b09e9d5f/pkg/reconciler/common/utils.go#L30

Update the func `FetchVersionFromCRD` to have param for configmap name
and the func would find the version from that configmap and return it

for pipelines the configmap is `pipelines-info` https://github.com/tektoncd/pipeline/blob/main/config/config-info.yaml
similarly they re present for other components too

/kind feature
/good-first-issue
/help-wanted

vinamra28:
# Changes

Currently we were reading the component version from CRD but with
Pipelines version 0.25.x release and Triggers 0.15.x release the version
information is available in a ConfigMap and anyone who is
system:authenticated can read the version from ConfigMap. Also making it
backward compatible so that if someone tries to install the older
version of pipelines or triggers then version information should be
fetched from the CRD as ConfigMap will not be there.

Closes #413 

# Submitter Checklist

review them:

# Release Notes
```release-note
Fetch release version from ConfigMap for Pipelines and Triggers and if ConfigMap not available then fetch from CRD.
```


pkg/reconciler/kubernetes/tektonpipeline/controller.go
```diff
@@ -39,6 +40,7 @@ import (
 )
 
 const releaseLabel = "pipeline.tekton.dev/release"
```
sm43, line 42: we can remove this now :)

sm43: ah ok you are using both of them, but we can just skip reading from crd
as configmap was added a few releases back and now if configmap is missing then something is wrong with the release yaml

sm43: wdyt?

vinamra28: what is somebody tries to install Pipelines version 0.24.x and Triggers version 0.14.x ?

vinamra28: also user has complete freedom to delete the configmap

sm43: yeah we don't need to support older versiosn
this is possible only when user has cloned the repo and doing make apply with required version

sm43: when we create a release we use specific versions of components

sm43: > also user has complete freedom to delete the configmap

this is not allowed.. operator will create it again in next reconcilation 😄

vinamra28: I am fine removing the code for fetching the version from CRDs. :smile: 
my main objective was to make it backward compatible

pkg/reconciler/common/utils.go
```diff
@@ -22,24 +22,28 @@ import (
...
+	dataObj, _, _ := unstructured.NestedStringMap(versionConfigMap.Object, "data")
+	version := dataObj["version"]
+
+	if version != "" {
+		return version, nil
 	}
 
-	return version, nil
+	return "", fmt.Errorf("failed to find version data field on ConfigMap %s", configMapName)
 }
```
nikhil-thomas, line 47: @vinamra28 when version is unknown (configmap not exist, or version field not exist) let us return `unknown`, instead of returning error. 

Let us not return error and block/break installation for this reason. 🧑‍💻

vinamra28: @nikhil-thomas maybe we can return error from here and handle that in TektonPipeline and TektonTriggers controller. This way our error would be logged and we can set version to unknown over there. WDYT?

vinamra28: like :arrow_down: 
```go
                var releaseVersion string = "unknown"

		// Read the release version of pipelines
		releaseVersion, err = common.FetchVersionFromConfigMap(manifest, versionConfigMap)
		if err != nil {
			logger.Infow("failed to read release version from ConfigMap", err)
		}
```

nikhil-thomas: this could work, 
but i would make it an internal implementation detail of `releaseVersion()`, because this error handling doesn't add more claritiy to the reconciler flow. 🧑‍💻

sm43: why do we wanna show unknown? in what case that would happen?
nightly, release version, dev release yaml 
all have version set with some value.
then if version is not set then doesn't mean something is wrong?

nikhil-thomas: > then if version is not set then doesn't it mean something is wrong?
yes, but as this won't affect any critical functionality. 
can we just do a warning. is the error and break necessary.

nikhil-thomas: in addition, we will see the `Unknown` value in the CR status, i think that could be the necessary indicator that something is wrong 😇

sm43: > this won't affect any critical functionality

if the release yaml is missing something then it can
i don't mind keeping it this way 😄  ... but if we assume version is missing i.e. something is wrong with release yaml and still we allow everything to install then we might break an existing working pipelines, triggers
if we stop then only operator would fail !

vdemeester: we can "type" the error and have a `IsUnknow` error check. From this part of the code perspective, it make sense to return an error. The consumer of this is the one with the responsability to handle it.

nikhil-thomas:
adding a hold as @vinamra28 is making a minor change.

@vinamra28 please feel free to remove this hold at any point (with or without changes)

vinamra28:
> adding a hold as @vinamra28 is making a minor change.
> 
> @vinamra28 please feel free to remove this hold at any point (with or without changes)

thanks @nikhil-thomas for this :+1: 
did the changes.
/hold cancel

PR #428, opened 2021-09-27
Fixes AppendManifest used to read files from kodata

### LINKED ISSUE #425: TektonInstallerSet created have each resource twice
When we read resources from ko data and create installer set, the installer set has 2 copies of each resource.

for eg,
In addons, for clustertasks, when we read clustertasks from kodata [here](https://github.com/tektoncd/operator/blob/3c77c775a7464cb8fd42e4addace86298d5d39df/pkg/reconciler/openshift/tektonaddon/tektonaddon.go#L329) -> it calls [`applyAddons`](https://github.com/tektoncd/operator/blob/3c77c775a7464cb8fd42e4addace86298d5d39df/pkg/reconciler/openshift/tektonaddon/tektonaddon.go#L481)
in apply addon we pass the path of the directory to ` common.AppendManifest(manifest, addonLocation)` and it
reads all the resources at the location and returns inside the manifest obj
but the manifest has 2 copies of each resource so manifest is getting bigger.

It is probably issue with [AppendManifest](https://github.com/tektoncd/operator/blob/3c77c775a7464cb8fd42e4addace86298d5d39df/pkg/reconciler/common/releases.go#L176)

To reproduce the issue
1. install operator using `make TARGET=openshfit` apply` on Openshift
2. let `all` profile install all components
3. for addons there would 4 installler set created
4. check any installer set if there are 2 instances of each resource eg. in clustertask `addon-clustertask- ` installerset there would be 2 buildah clustertask which are same.

sm43:
Previously, if we pass a path to AppendManifest
for ex. `/kodata/tekton/abc.yaml` and later
        `/kodata/tekton/`
both use to read the file and add in the manifest, which
would make manifest to have the same resource twice.

This patch updates the AppendManifest func to use mf.ManifestFrom
    which will recursively read all files from a location by passing
    the path to it.

Closes #425 

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/common/releases.go
+++ b/pkg/reconciler/common/releases.go
@@ -174,19 +174,10 @@ func latestRelease(instance v1alpha1.TektonComponent) string {
 }
 
 func AppendManifest(manifest *mf.Manifest, yamlLocation string) error {
-	var files []string
-	if err := filepath.Walk(yamlLocation, func(path string, info os.FileInfo, err error) error {
-		files = append(files, path)
-		return nil
-	}); err != nil {
+	m, err := mf.ManifestFrom(mf.Recursive(yamlLocation))
+	if err != nil {
 		return err
 	}
-	for i := range files {
-		m, err := Fetch(files[i])
-		if err != nil {
-			return err
-		}
-		*manifest = manifest.Append(m)
-	}
+	*manifest = manifest.Append(m)
 	return nil
 }
--- a/pkg/reconciler/common/releases_test.go
+++ b/pkg/reconciler/common/releases_test.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"testing"
 
+	mf "github.com/manifestival/manifestival"
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
 	util "github.com/tektoncd/operator/pkg/reconciler/common/testing"
 )
@@ -46,4 +47,29 @@ func TestListReleases(t *testing.T) {
 	version, err := allReleases(&v1alpha1.TektonTrigger{})
 	util.AssertEqual(t, err, nil)
 	util.AssertDeepEqual(t, version, expectedVersionList)
+}
+
+func TestAppendManifest(t *testing.T) {
+
+	// Case 1
+	var manifest mf.Manifest
+	err := AppendManifest(&manifest, "testdata/kodata/tekton-addon")
+	if err != nil {
+		t.Fatal("failed to read yaml: ", err)
+	}
+
+	if len(manifest.Resources()) != 3 {
+		t.Fatalf("failed to find expected number of resource: %d found, expected 3", len(manifest.Resources()))
+	}
+
+	// Case 2
+	var newManifest mf.Manifest
+	err = AppendManifest(&newManifest, "testdata/kodata/tekton-addon/0.0.1")
+	if err != nil {
+		t.Fatal("failed to read yaml: ", err)
+	}
+
+	if len(newManifest.Resources()) != 1 {
+		t.Fatalf("failed to find expected number of resource: %d found, expected 3", len(newManifest.Resources()))
+	}
 }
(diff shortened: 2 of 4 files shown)


pkg/reconciler/common/testdata/kodata/tekton-addon/0.0.2/01-clustertriggerbindings/bitbucket-1.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: ClusterTriggerBinding
+metadata:
+  name: bitbucket-pullreq-0.0.2-copy
```
savitaashture, line 4: why duplicated file ?

sm43: I was testing having multiple files in the directory

pkg/reconciler/common/releases.go
```diff
@@ -181,7 +181,16 @@ func AppendManifest(manifest *mf.Manifest, yamlLocation string) error {
 	}); err != nil {
 		return err
 	}
+
 	for i := range files {
+		folderInfo, err := os.Stat(files[i])
+		if os.IsNotExist(err) {
+			return fmt.Errorf("failed to find file/dir at path: %s: %s ", files[i], err.Error())
+		}
```
nikhil-thomas, line 189: to me it looks like`IsNotExist` is the one error which is not guranteed to happen here as the file list is populated by the above walk operation.

so shall we either return for all error here. It looks like if the error is of any kind it will be ignored.

if this is to add whether the a file exists when a single file path is provided, then shall we move this check to the beginning of the AppendManifest function. 

In addition, wouldn't the `filepath.Walk` catch that error if the `yamlLocation` param is a non-existent file.

nikhil-thomas:
> Previously, if we pass a path to AppendManifest
for ex. /kodata/tekton/abc.yaml and later
/kodata/tekton/
both use to read the file and add in the manifest, which
would make manifest to have the same resource twice.

I think this should be solved by restructuring the file tree. This patch seems to be a bit too restrictive, because current way of doing will be the general expectation.

so if a data root has 2 separate things which are handled by 2 entities.
eg: 
```
/kodata/tekton/abc.yaml 
/kodata/tekton/`
```
let us change that to

```
/kodata/tekton/abc-resources/abc.yaml and later
/kodata/tekton/main-resources/
```
note: main-resources or any other significant name.

nikhil-thomas:
> > Previously, if we pass a path to AppendManifest
> > for ex. /kodata/tekton/abc.yaml and later
> > /kodata/tekton/
> > both use to read the file and add in the manifest, which
> > would make manifest to have the same resource twice.
> 
> I think this should be solved by restructuring the file tree. This patch seems to be a bit too restrictive, because current way of doing will be the general expectation.
> 
> so if a data root has 2 separate things which are handled by 2 entities.
> eg:
> 
> ```
> /kodata/tekton/abc.yaml 
> /kodata/tekton/`
> ```
> 
> let us change that to
> 
> ```
> /kodata/tekton/abc-resources/abc.yaml and later
> /kodata/tekton/main-resources/
> ```
> 
> note: main-resources or any other significant name.

it turns out my assumption is wrong here. This pr is actually trying to deduplicate Appending of manifest as [manifestival doesnot check for duplicates](https://github.com/manifestival/manifestival/blob/bd449da8687a4dfb81777f9eebf8e3ca40bc9798/manifestival.go#L78). 

so let us either move the changes in this patch into a function like (skipDir) and make the need for this change visible through a comment. The flow of logic in the for loop body will be simplified.

pkg/reconciler/common/releases.go
```diff
@@ -174,19 +174,10 @@ func latestRelease(instance v1alpha1.TektonComponent) string {
...
 		return err
 	}
-	for i := range files {
-		m, err := Fetch(files[i])
-		if err != nil {
-			return err
-		}
-		*manifest = manifest.Append(m)
-	}
+	*manifest = manifest.Append(m)
```
nikhil-thomas, line 181: 👍

PR #432, opened 2021-10-04
Modify olm.skipRange handling in bundle generator

nikhil-thomas:
Add a `--olm-skip-range` flag to capture explicit value for
`olm.skipRange` annotation

Note: Earlier the tool used to compute the upper and lower bounds for
skipRange from `operator-version` and `previous-operator-version`
arguments. This lack the flexibility to set a skipRange which is defined
by a development teams release process.

Note: The already existing flag `--addn-annotations key1=val1,key2=val2` couldn't
 be used as the parsing fails on the special characters in skipRanges.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/operatorhub/openshift/release-artifacts/bundle/manifests/openshift-pipelines-operator-rh.clusterserviceversion.yaml
+++ b/operatorhub/openshift/release-artifacts/bundle/manifests/openshift-pipelines-operator-rh.clusterserviceversion.yaml
@@ -28,7 +28,7 @@ metadata:
     description: Red Hat OpenShift Pipelines is a cloud-native CI/CD solution for
       building pipelines using Tekton concepts which run natively on OpenShift and
       Kubernetes.
-    olm.skipRange: '>=1.5.2 <1.6.0'
+    olm.skipRange: '>=1.5.0 <1.6.0'
     operators.openshift.io/infrastructure-features: '["disconnected","proxy-aware"]'
     operators.operatorframework.io/builder: operator-sdk-v1.7.1+git
     operators.operatorframework.io/internal-objects: '["tektoninstallersets.operator.tekton.dev",
--- a/operatorhub/openshift/README.md
+++ b/operatorhub/openshift/README.md
@@ -7,24 +7,25 @@ bundle for using `local` strategy.
 
 **Note:** The input release.yaml could be a github release or the result of `ko resolve config`
 
-1. From the project root (tektoncd/operator) run
+ From the project root (tektoncd/operator) run
 
     ```bash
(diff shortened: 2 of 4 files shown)


vdemeester (approved):

operatorhub/openshift/release-artifacts/bundle/manifests/openshift-pipelines-operator-rh.clusterserviceversion.yaml
```diff
@@ -28,7 +28,7 @@ metadata:
     description: Red Hat OpenShift Pipelines is a cloud-native CI/CD solution for
       building pipelines using Tekton concepts which run natively on OpenShift and
       Kubernetes.
-    olm.skipRange: '>=1.5.2 <1.6.0'
+    olm.skipRange: '>=1.5.0 <1.6.0'
```
sm43, line 31: trying to understand what does this change does?
previously as it was `>=1.5.2` it was not allowing 1.5.1 to get installed so we are including that in the range now right?

nikhil-thomas: this is just an example.. the values in this annotation will depend on how a team want the update graph to look like.

so theoretically, when we make a 1.6.0 release, this annotation will have value `>=1.5.0 <1.6.0`

This annotation will not prevent any version from being installed if it is the latest. The need for this annotation is to state that update can directly jump from minor to minor instead of rolling through patch versions. I state again, this will totally upto the team to decide, depend upon state of update graph and release policy/state of a version.

nikhil-thomas: > it was not allowing 1.5.1 to get installed so we are including that in the range now right?

that is unrelated to this.
if we do an annotation like this in our 1.6 release then olm will allow direct upgrades from any `1.5.x` till `1.6.0`. if a user is on 1.5.2 and even though there were othere releases like 1.5.3, 1.5.4, the upgrade from 1.5.2 will directly jump to 1.6.0 for that user.

nikhil-thomas: ref: https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#skiprange

pradeepitm12 (approved):
/lgtm

PR #467, opened 2021-10-26
Ship Tekton Hub via Tekton Operator

vinamra28:
# Changes

This following patch contains controllers for deploying Tekton Hub DB
and API components. This also contains a post-reconciler based on
platform which will add an Ingress in case of kubernetes and Route in
case of OpenShift.

Signed-off-by: vinamra28 <vinjain@redhat.com>
# Submitter Checklist

review them:

# Release Notes
```release-note
Add Tekton Hub DB and API reconcilers to ship via Tekton Operator.
```


nikhil-thomas:
@vinamra28 i see that this patch adds hub resources into operator kodata. But we don't want that. 
All payloads are now pulled during just in time before a build/testing using https://github.com/tektoncd/operator/blob/main/hack/fetch-releases.sh

So remove all files from `cmd/kubernetes/operator/kodata/hub`. Then update the script so that hub releases are pulled `just in time`.

The reason for this is that, without this script a component manager will have to make a pr to operator each time you need to add a new version. By using the script, the operator logic and the payload will be more loosley coupled.

@sm43

vinamra28:
> @vinamra28 i see that this patch adds hub resources into operator kodata. But we don't want that. All payloads are now pulled during just in time before a build/testing using https://github.com/tektoncd/operator/blob/main/hack/fetch-releases.sh
> 
> So remove all files from `cmd/kubernetes/operator/kodata/hub`. Then update the script so that hub releases are pulled `just in time`.
> 
> The reason for this is that, without this script a component manager will have to make a pr to operator each time you need to add a new version. By using the script, the operator logic and the payload will be more loosley coupled.
> 
> @sm43

@nikhil-thomas we don't have any release.yaml for Hub as of now so for the time being we have made it like this 😅. We as in hub team are yet to decide how to generate the release YAMLs. Either we'll be having 4 release YAMLs each for separate component or a single release yaml.
cc @piyush-garg @PuneetPunamiya @sm43

nikhil-thomas:
> we don't have any release.yaml for Hub

we need that :technologist:  to integrate with operator.
adding these static files in operator code base is difficuly to manage.

you can use the standard steps from the release pipeline of all tekton projects for hub as well.

nikhil-thomas:
>  Either we'll be having 4 release YAMLs each for separate component or a single release yaml.

both options are fine.

for example, triggers has 2 separate yaml files which constitue their release artifact.

pkg/apis/operator/v1alpha1/tektonhub_default_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestSetDefault(t *testing.T) {
+
+	th := &TektonHub{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "hub",
+		},
+		Spec: TektonHubSpec{
+			Api: ApiSpec{
+				HubConfigUrl:  "https://hubconfig.url",
```
piyush-garg, line 34: Why do we have defaulted this value for hub config?

pkg/apis/operator/v1alpha1/tektonhub_defaults.go
```diff
@@ -0,0 +1,28 @@
...
+package v1alpha1
+
+import (
+	"context"
+)
+
+func (th *TektonHub) SetDefaults(ctx context.Context) {
+
+	if th.Spec.Db.DbSecretName == "" {
+		th.Spec.Db.DbSecretName = "db"
```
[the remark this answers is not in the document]
piyush-garg: +1 on having prefix, we need to refactor the yamls a bit so that they are specific to tekton hub

pkg/apis/operator/v1alpha1/tektonhub_default_test.go
vinamra28: We haven't defaulted the `HubConfigUrl` value. This is the required field hence we need to specify this in struct for the validation webhook to validate this.

nikhil-thomas:
> @vinamra28 i see that this patch adds hub resources into operator kodata. But we don't want that. All payloads are now pulled during just in time before a build/testing using https://github.com/tektoncd/operator/blob/main/hack/fetch-releases.sh
> 
> So remove all files from `cmd/kubernetes/operator/kodata/hub`. Then update the script so that hub releases are pulled `just in time`.
> 
> The reason for this is that, without this script a component manager will have to make a pr to operator each time you need to add a new version. By using the script, the operator logic and the payload will be more loosley coupled.
> 
> @sm43

/hold

holding till we have a mechanism to pull in hub payload (instead of adding the release.yaml in this codebase)

pkg/apis/operator/v1alpha1/tektoninstallerset_lifecycle.go
```diff
@@ -29,6 +29,7 @@ const (
 	WebhookReady         apis.ConditionType = "WebhooksReady"
 	ControllerReady      apis.ConditionType = "ControllersReady"
 	AllDeploymentsReady  apis.ConditionType = "AllDeploymentsReady"
+	JobCompleted         apis.ConditionType = "JobCompleted"
```
nikhil-thomas, line 32: im not sure whether this fits in here.
may be we should add this to TektonHub status. 

If the "Job"we are talking about here is the db-migration, then it will make more sense if it is a status in TektonHub with a proper "DatabaseMigration" (True)

Having a "JobCompleted" status in TektonInstallerSet Status is not much informational in this scenario.

Feel free to create an additional InstallerSet to handle the Job alone. ie, the Job alone could have a different lifecycle.

nikhil-thomas: May be PostReconcile might be the place where we can handle the "Job"better

pkg/reconciler/common/releases.go
```diff
@@ -92,6 +92,8 @@ func ComponentDir(instance v1alpha1.TektonComponent) string {
 		return filepath.Join(koDataDir, "tekton-config")
 	case *v1alpha1.TektonResult:
 		return filepath.Join(koDataDir, "tekton-results")
+	case *v1alpha1.TektonHub:
+		return filepath.Join(koDataDir, "hub")
```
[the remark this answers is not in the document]
nikhil-thomas: we want it to be deleted when we do `make clean`as we will be adding the mechanism to pull the resources to fetch components script.

pkg/reconciler/kubernetes/tektonhub/transformer.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func changeNamespace(targetNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetNamespace() != targetNamespace {
+			u.SetNamespace(targetNamespace)
+			return nil
+		}
+		return nil
+	}
```
nikhil-thomas, line 48: do we really need to redefine these 2 transformers as they already exist in manifestival and is used everywhere else.

- https://github.com/tektoncd/operator/blob/e37aa26ed3db9b58dd3f1777ff8ae775cfe604d8/pkg/reconciler/common/transformers.go#L100

- https://github.com/tektoncd/operator/blob/e37aa26ed3db9b58dd3f1777ff8ae775cfe604d8/pkg/reconciler/common/transformers.go#L65

cmd/kubernetes/operator/kodata/hub/v1.4.0/db-migration/db-migration.yaml
```diff
@@ -0,0 +1,52 @@
...
+    app: db
+spec:
+  template:
+    spec:
+      containers:
+        - name: db-migration
+          image: quay.io/tekton-hub/db-migration:v1.4.0
+          env:
+            - name: POSTGRES_HOST
+              value: tekton-hub-db
```
nikhil-thomas, line 30: im assuming that we will remove these yaml files from this patch and extend the [fetch releases script](https://github.com/tektoncd/operator/blob/main/hack/fetch-releases.sh)

pkg/reconciler/kubernetes/tektonhub/controller.go
```diff
@@ -0,0 +1,85 @@
...
+		impl := tektonHubReconciler.NewImpl(ctx, c)
+
+		// Add enqueue func in reconciler
+		c.enqueueAfter = impl.EnqueueAfter
+
+		logger.Info("Setting up event handlers")
+
+		tektonHubInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
+
+		tektonHubInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
```
sm43, line 79: both are `tektonHubInformer` ? 
one would be installerSet informer

docs/TektonHub.md
```diff
@@ -0,0 +1,96 @@
+# Tekton Hub
+
+TektonHub custom resource allows user to install and manage [Tekton Hub][hub].
+
+TektonHub is an optional component and currently cannot be installed through TektonConfig. It has to be installed seperately. It is available for both Kubernetes and OpenShift platform.
+
+To install Tekton Hub on your cluster follow steps as given below:
+
+1.  Make sure that `tekton-pipelines` namespace is present on your
```
sm43, line 9: we should follow `targetNamespace` similar to other component
by default in config dir CR examples we can have the defaults per platform 
but user should be allowed to install any namespace
in hub case it is straight forward .. instead injecting `tekton-pipelines` ns in resources
we inject spec.targetNamespace

pkg/apis/operator/v1alpha1/tektoninstallerset_lifecycle.go
sm43: +1 on not adding this as installerSet cond

pkg/reconciler/kubernetes/tektonhub/controller.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			logger.Fatalw("Error creating initial manifest", zap.Error(err))
+		}
+
+		c := &Reconciler{
+			kubeClientSet:     kubeClient,
+			operatorClientSet: operatorclient.Get(ctx),
+			extension:         generator(ctx),
+			manifest:          manifest,
+			namespace:         os.Getenv("DEFAULT_TARGET_NAMESPACE"),
```
sm43, line 68: this can be read from spec.targetNamespace

pkg/reconciler/kubernetes/tektonhub/transformer.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+
+		u.SetAnnotations(annotations)
+		return nil
+	}
+}
+
+func changeIngressHostValue(hostValue string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
```
sm43, line 66: both transformer can be merged 
they both are for same obj

cmd/kubernetes/operator/kodata/hub/v1.4.0/db-migration/db-migration.yaml
vinamra28: yes

pkg/reconciler/kubernetes/tektonhub/testdata/ingress.yaml
```diff
@@ -0,0 +1,35 @@
+# Copyright © 2021 The Tekton Authors.
```
PuneetPunamiya, line 1: Do we need this file ?

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,580 @@
...
+func (r *Reconciler) validateApiDependencies(ctx context.Context, th *v1alpha1.TektonHub) error {
+	logger := logging.FromContext(ctx)
+	apiSecretKeys := []string{"GH_CLIENT_ID", "GH_CLIENT_SECRET", "JWT_SIGNING_KEY", "ACCESS_JWT_EXPIRES_IN", "REFRESH_JWT_EXPIRES_IN", "GHE_URL"}
+	apiConfigMapKeys := []string{"CONFIG_FILE_URL"}
+
+	th.Status.MarkApiDependencyInstalling("checking for api secrets in the namespace and creating the ConfigMap")
+
+	_, err := r.getSecretForHub(ctx, th.Spec.Api.ApiSecretName, namespace, apiSecretKeys)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
PuneetPunamiya, line 295: I think we are not creating secrets anywhere 
we should create the secrets here

config/base/300-operator_v1alpha1_hub_crd.yaml
```diff
@@ -0,0 +1,54 @@
...
+          type: string
+        - jsonPath: .status.conditions[?(@.type=="Ready")].status
+          name: Ready
+          type: string
+        - jsonPath: '.status.conditions[?(@.type=="Ready")].reason'
+          name: Reason
+          type: string
+        - jsonPath: .status.apiUrl
+          name: ApiUrl
+          type: string
```
PuneetPunamiya, line 49: We are missing the auth url

hack/fetch-releases.sh
```diff
@@ -120,7 +120,48 @@ release_yaml_pac() {
...
 }
 
-#Args: <target-platform> <pipelines version> <triggers version> <dashboard version> <results version> <pac version>
+
+release_yaml_hub() {
+  echo fetching '|' component: ${1} '|' version: ${2}
+  local version=$2
+
+  ko_data=${SCRIPT_DIR}/cmd/${TARGET}/operator/kodata
+  dirPath=${ko_data}/tekton-hub/${version}
```
nikhil-thomas, line 129:
```suggestion
  dirPath=${ko_data}/tekton-hub
```
Currently the operator stories only 1 version of any payload. So let us make sure that kodata has only 1 set of payload for any component.

In addition, the name `dirPath` is as generic as naming a variable `x` 😛 . Make it a name that is meaningful.

hack/fetch-releases.sh
```diff
@@ -120,7 +120,48 @@ release_yaml_pac() {
...
+  mkdir -p ${dirPath} || true
+
+  url=""
+  components="db db-migration"
+
+  if [[ ${TARGET} == "openshift" ]]; then
+    components="$components api-openshift"
+  else
+    components="$components api-k8s"
+  fi
```
nikhil-thomas, line 140:
```suggestion
  components="db db-migration api"
```

is the `if` condition necessary, as this is happening already inside `cmd/<platform>/operator/kodata`

nikhil-thomas: oh i see.. 🧑‍💻 
resolving 🤓

hack/fetch-releases.sh
```diff
@@ -120,7 +120,48 @@ release_yaml_pac() {
...
+
+  dirNames=(db db-migration api)
+  arrayIndex=0
```
nikhil-thomas, line 143:
```suggestion
  components="db db-migration api"
```

and then inside the for loop we can add an if condition specifically for api.
for example

```bash
  fileName=${component}.yaml
  [[ ${component} == "api" ]] && fileName=${component}-${TARGET}.yaml
```

and dest can be `dest=${dirPath}/${component}`

we will not need `arrayIndex=$((arrayIndex+1))`

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -379,8 +379,17 @@ func (r *Reconciler) targetNamespaceCheck(ctx context.Context, tp *v1alpha1.Tekt
 	ns, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, tp.GetSpec().GetTargetNamespace(), metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			if err := common.CreateTargetNamespace(labels, tp, r.kubeClientSet); err != nil {
+				return err
```
nikhil-thomas, line 383: from my experience an additional check on this error `isAlreadyExist` and `return nil` will keep the recociler running.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	createdByKey       = "operator.tekton.dev/created-by"
+	targetNamespaceKey = "operator.tekton.dev/target-namespace"
+	createdByValue     = "TektonHub"
+)
+
+// FinalizeKind removes all resources after deletion of a TektonHub.
+func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.TektonHub) pkgreconciler.Event {
+	logger := logging.FromContext(ctx)
+
+	installerSets := original.Status.HubInstallerSet
```
nikhil-thomas, line 83: This should be handled using a List query. We don't use name(s) to query installerSets anymore. It can cause rogue un-garbage collected installerSets. 

We need to set labels (created-by, type and verision). then in finalize use the `created-by` and may be `type` label to make a List query and delete all that is returned

examples:
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L100
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L360

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	if len(installerSets) == 0 {
+		return nil
+	}
+
+	for _, value := range installerSets {
+		err := r.operatorClientSet.OperatorV1alpha1().TektonInstallerSets().Delete(ctx, value, metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
```
nikhil-thomas, line 93: This should be handled using a List query. We don't use name(s) to query installerSets anymore. It can cause rogue un-garbage collected installerSets. 

We need to set labels (created-by, type and verision). then in finalize use the `created-by` and may be `type` label to make a List query and delete all that is returned

examples:
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L100
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L360

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+
+	if th.GetName() != v1alpha1.HubResourceName {
+		msg := fmt.Sprintf("Resource ignored, Expected Name: %s, Got Name: %s",
+			v1alpha1.HubResourceName,
+			th.GetName(),
+		)
+		logger.Error(msg)
+		th.Status.MarkNotReady(msg)
+		return nil
+	}
```
nikhil-thomas, line 158: move this check to the beggining of the reconcile function. 
if this doesn't checkout the operator doesn't need to do anything further.

eg: https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L121

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	if err := common.Uninstall(ctx, manifest, nil); err != nil {
+		logger.Error("Failed to finalize platform resources", err)
+	}
+	return nil
+}
+
+// ReconcileKind compares the actual state with the desired, and attempts to
+// converge the two.
+func (r *Reconciler) ReconcileKind(ctx context.Context, th *v1alpha1.TektonHub) pkgreconciler.Event {
```
nikhil-thomas, line 125: first we need the CR name check (as k8s doesn't support singleton CRDs, yet).

Then we need some logic to check if the current loops is an upgrade. 
eg: [here](https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L135) without this in upgrades TektonHub reconciler:
1. will not wait for conditions/checks on other resources (even if it can it will be on resources of previous versions)
2. will not reset conditions, without it the conditions will stay true during an upgrade.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	logger := logging.FromContext(ctx)
+	th.Status.InitializeConditions()
+	th.Status.ObservedGeneration = th.Generation
+	namespace = th.Spec.GetTargetNamespace()
+
+	if namespace == "" {
```
nikhil-thomas, line 131: handle this in defaulting webhook not in reconciler.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+
+	_, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			if err := common.CreateTargetNamespace(map[string]string{}, th, r.kubeClientSet); err != nil {
+				return err
+			}
+		}
+		return err
+	}
```
nikhil-thomas, line 144: extract this into a function, these checks are a bit low level from the reconciler loop context.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
nikhil-thomas: move all the manfest fetching logic into the controler.go
eg: https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/controller.go#L54

or else move the logic for creating defferent types of installerSets into individual functions.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+}
+
+// Validate DB are present on the cluster. If DB secrets are present and all the keys don't
+// have values then update the remaining values with default values. If the DB secret
+// is not present then create a new secret with default values.
+func (r *Reconciler) validateOrCreateDBSecrets(ctx context.Context, th *v1alpha1.TektonHub) error {
+	logger := logging.FromContext(ctx)
+	dbKeys := []string{"POSTGRES_DB", "POSTGRES_USER", "POSTGRES_PASSWORD", "POSTGRES_PORT"}
+
+	th.Status.MarkDbDependencyInstalling("db secrets are being added into the namespace")
```
nikhil-thomas, line 270: this is an overkill, this will not be updated anyway in status before the end of this reconcile call. So we can get rid of this.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+
+	dbSecret, err := r.getSecretForHub(ctx, th.Spec.Db.DbSecretName, namespace, dbKeys)
```
nikhil-thomas, line 272: if err == nil  then we should still do validation for sanity.
without it the validation happens only once, ie, during creation.

pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
```diff
@@ -177,5 +177,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, installerSet *v1alpha1.T
 	// Mark all deployments ready
 	installerSet.Status.MarkAllDeploymentsReady()
 
+	// job
+	err = installer.IsJobCompleted()
+	if err != nil {
+		return err
+	}
+
```
nikhil-thomas, line 185: this should be something handled in a reconciler hub reconciler.

At present let us use tektoninstallerset to create resources and check if they are created properly. ie, clusterOperations related things.

 Checking for completion of a job seems to be something lifcycle related. Let the installerset create it, then the lifecycle aware tektonhub reconciler can check the job completion status and proceed.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	// manifest is empty, but with a valid client and logger. all
+	// manifests are immutable, and any created during reconcile are
+	// expected to be appended to this one, obviating the passing of
+	// client & logger
+	manifest mf.Manifest
+	// Platform-specific behavior to affect the transform
+	extension common.Extension
+	// enqueueAfter enqueues a obj after a duration
+	enqueueAfter func(obj interface{}, after time.Duration)
+}
```
nikhil-thomas, line 57: let us add operator version also in here. The reconciler(s) have to decide whether a pod restart is an upgrade based on that.

The reason for this is that, even when the code in hub hasn't changed the changes from base images in builds (eg: cve fixes) has to be delivered to hub deployments.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	// Check if installer set is already created
+	compInstallerSet, ok := th.Status.HubInstallerSet[component]
+	if !ok {
+		return false, nil
+	}
+
+	if compInstallerSet != "" {
+		// if already created then check which version it is
+		ctIs, err := oc.OperatorV1alpha1().TektonInstallerSets().
+			Get(ctx, compInstallerSet, metav1.GetOptions{})
```
nikhil-thomas, line 491: We need to use List() instead of Get to verify this. ie, use a set of labels to query for Installersets instead of using a constant name.

If more than one InstallerSet is returned for a type delete all and create a new one.

As installersets are being create using `GenerateName: fmt.Sprintf("%s-", prefix),` using names and Get queries is not enough to catch duplicates and obsolete installersets. 

eg: 
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L160
- 
https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L359

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+func makeInstallerSet(th *v1alpha1.TektonHub, manifest mf.Manifest, prefix, releaseVersion, namespace string) *v1alpha1.TektonInstallerSet {
+	ownerRef := *metav1.NewControllerRef(th, th.GetGroupVersionKind())
+	return &v1alpha1.TektonInstallerSet{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-", prefix),
+			Labels: map[string]string{
+				createdByKey: createdByValue,
+			},
+			Annotations: map[string]string{
+				releaseVersionKey:  releaseVersion,
```
nikhil-thomas, line 556: we need additional labels here:
- type
- version (operator version)
- createdBy

eg: https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L359

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+		r.enqueueAfter(th, 10*time.Second)
+		return err
+	}
+
+	th.Status.MarkApiDependenciesInstalled()
+
+	// Create the API route based on platform
+	if err := r.extension.PreReconcile(ctx, th); err != nil {
+		return err
+	}
```
nikhil-thomas, line 226: move this next to `if th.GetName() != v1alpha1.HubResourceName {}`

"PreReconciling" should be done before the main reconciling. This is "idomatic" in this operator. ie, that is the mental model everyone has and that is what everyone will assume during development and debugging.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+
+	if s.Data["POSTGRES_PORT"] == nil || len(s.Data["POSTGRES_PORT"]) == 0 {
+		s.StringData["POSTGRES_PORT"] = "5432"
+	}
+
+	return s
+}
+
+func (r *Reconciler) applyManifest(ctx context.Context, manifestLocation string, th *v1alpha1.TektonHub, installerSetName, version, prefixName string) error {
```
nikhil-thomas, line 441: misleading name `applyManifest`. 
i was expecting to see a `m.Apply()`. I understand that createInstallerSet might be already in use. May be make this name `setupAndCreateInstallerSet` or some other name that reveals the intent(behavior) of this method. 

Explicit and boring is good. 🧑‍💻

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+			// Not deleting all TektonHubs. Nothing to do here.
+			return nil
+		}
+	}
+
+	if err := r.extension.Finalize(ctx, original); err != nil {
+		logger.Error("Failed to finalize platform resources", err)
+	}
+	logger.Info("Deleting cluster-scoped resources")
+	manifest, err := r.installed(ctx, original)
```
nikhil-thomas, line 112: is this necessary, as we are creating all resources throught installersets, deleting installersets should delete the resources. In addition, leave as much as clusterOperations to TektonInstallerSet Reconciler.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
vinamra28: @nikhil-thomas I guess this comment is related to PR https://github.com/tektoncd/operator/pull/600 as I have rebased over it for now

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+		if err == errKeyMissing {
+			th.Status.MarkApiDependencyMissing(fmt.Sprintf("%s secret is missing the keys", th.Spec.Api.ApiSecretName))
+			return err
+		} else {
+			logger.Error(err)
+			return err
+		}
+	}
+
+	_, err = r.getConfigMapForHub(ctx, apiConfigMapName, namespace, apiConfigMapKeys)
```
nikhil-thomas, line 323: style suggestion
```suggestion
	_, err = r.getConfigMap(ctx, apiConfigMapName, namespace, apiConfigMapKeys)
```

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -0,0 +1,595 @@
...
+	return nil
+}
+
+func (r *Reconciler) installed(ctx context.Context, instance v1alpha1.TektonComponent) (*mf.Manifest, error) {
+	// Create new, empty manifest with valid client and logger
+	installed := r.manifest.Append()
+	stages := common.Stages{common.AppendInstalled}
+	err := stages.Execute(ctx, &installed, instance)
+	return &installed, err
+}
```
nikhil-thomas, line 474: imo, we do not need this. we should just delete the installerset and let installerSet reconciler worry about getting items and deleting them.

In other words, the logic in this function is clusterOperation related not component lifecycle related. So let us not handle it here.

pkg/reconciler/kubernetes/tektoninstallerset/install.go
```diff
@@ -320,6 +326,24 @@ func (i *installer) AllDeploymentsReady() error {
...
+			return err
+		}
+		if !isJobCompleted(job) {
+			return fmt.Errorf("Job not successful")
+		}
+	}
+
+	return nil
+}
+
```
nikhil-thomas, line 346: From installerSet stand point isJobRunninig is something that "might" come under the responsibility of installerset. The success and failure of a jobs is upto the parent reconciler to decode and interpret. 

If if makes sense, from TektonHub reconciler, create an additional installerset specifically for this job.

However, i don't have a strong opinion here. If it makes more sense to you to add this generic job completion check to installerSet reconciler, then we can keep this here.

pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -0,0 +1,173 @@
...
+	)
+	if err != nil {
+		logger.Error("failed to transform manifest")
+		return err
+	}
+
+	// Just apply the routes and nothing else
+	if err := manifest.Filter(mf.ByKind("Route")).Apply(); err != nil {
+		return err
+	}
```
nikhil-thomas, line 102: replace this with installerSet creation.
1. check for existing installerSet of this type (eg: 'hub-routes') 
1. if there are more than 1 delete it reconcile again
1. create/update instaler set
1. create and installerSet with createdBy, version, type labels.
1. also in extension finalizer add logic to query and delete the installerSets created in this extension.

eg: 
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/openshift/tektonpipeline/extension.go#L125
- https://github.com/tektoncd/operator/blob/a0cd67e5c428babf1f76316a14cf9c4526055cab/pkg/reconciler/openshift/tektonpipeline/extension.go#L204

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -219,6 +194,10 @@ func (r *Reconciler) ensureTargetNamespaceExists(ctx context.Context, tc *v1alph
 			}
 		}
 	} else {
+		labels := map[string]string{"operator.tekton.dev/targetNamespace": "true"}
+		if err := common.CreateTargetNamespace(labels, tc, r.kubeClientSet); err != nil {
+			return err
+		}
```
nikhil-thomas, line 200: are the lines following this logic 201-214 redundant. 201-214 is doing exactly what `common.CreateTargetNamespace()` is doing, isn't it.

docs/TektonHub.md
```diff
@@ -0,0 +1,91 @@
...
+
+    ```sh
+    kubectl apply -f <name>.yaml
+    ```
+
+1.  Check the status of installation using following command
+
+    ```sh
+    kubectl get tektonhub.operator.tekton.dev
+    ```
```
nikhil-thomas, line 89: Adding a sample output of a successfully created CR instance would be nice.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
vinamra28: the following changes were the part of #600 hence resolved there

hack/fetch-releases.sh
vinamra28: >In addition, the name dirPath is as generic as naming a variable x stuck_out_tongue . Make it a name that is meaningful.

@nikhil-thomas I also felt the same but then just saw other functions and kept it in consistent with others.

vinamra28: @nikhil-thomas we would be need the ${version} since operator is going to read this version dir and display that in the TektonHub CR status
```
$ kubectl get tektonhub hub
NAME   VERSION   READY   REASON   APIURL
hub    v1.6.0    True             https://api.url

nikhil-thomas:
/hold cancel

> holding till we have a mechanism to pull in hub payload (instead of adding the release.yaml in this codebase)

as this is resolved. 
the release.yaml is not embedded in the codebase anymore.

Issue #497, opened 2021-11-08
Validate Component CR name in webhook 


sm43:
all Operator CRDs are clusterscoped and we allow only one instance per component to be created.
for example. to install TektonPipeline, the CR has to be created by name `pipeline` and if it is something different then
we set the status to not ready with error msg here
https://github.com/tektoncd/operator/blob/c2e1245b3da591e7c0bbb2e2c892ce88bd169e09/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L109

now that we have webhook, we can do this validation in there and user would get instant error while applying the CR
we can add the name validation here
https://github.com/tektoncd/operator/blob/c2e1245b3da591e7c0bbb2e2c892ce88bd169e09/pkg/apis/operator/v1alpha1/tektonpipeline_validation.go#L25

this name validation can be added to all components

also we should move all name constants in a single place in `pkg/apis/...` and use them in reconciler whereever needed

/good-first-issue
/help-wanted

PR #501, opened 2021-11-11
Validates component CR name in webhook

PuneetPunamiya:
- This patch adds a check where it validates the name
    of the component CR before applying it on the cluster

  - This patch also moves all the constants to `pkg/apis/const.go`

Fixes: #497

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/apis/operator/v1alpha1/const.go
+++ b/pkg/apis/operator/v1alpha1/const.go
@@ -16,7 +16,9 @@ limitations under the License.
 
 package v1alpha1
 
-import "fmt"
+import (
+	"fmt"
+)
 
 var (
 	// RECONCILE_AGAIN_ERR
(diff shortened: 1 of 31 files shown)


pkg/apis/operator/v1alpha1/tektonaddon_validation.go
```diff
@@ -28,6 +28,10 @@ func (ta *TektonAddon) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return nil
 	}
 
+	if ta.GetName() != AddonResourceName {
+		errs = errs.Also(apis.ErrInvalidValue(ta.GetName(), "metadata.name,  Only one instance of TektonAddon is allowed by name `addon`"))
```
nikhil-thomas, line 32:
```suggestion
		errs = errs.Also(apis.ErrInvalidValue(ta.GetName(), "metadata.name,  Only one instance of TektonAddon is allowed by name",  AddonResourceName))
```

pkg/apis/operator/v1alpha1/const.go
```diff
@@ -62,3 +65,24 @@ var (
...
+	DashboardResourceName = "dashboard"
+	AddonResourceName     = "addon"
+	ConfigResourceName    = "config"
+	ResultResourceName    = "result"
+	Interval              = 10 * time.Second
+	Timeout               = 1 * time.Minute
+)
+
+const (
+	PipelineNotReady       = "tekton-pipelines not ready"
```
sm43, line 83: can we only move the const names ? 
those are used by webhook and controller both so we can keep at one place

`Interval`, `Timeout` and rest of them can be kept where it was before

PR #503, opened 2021-11-12
fix resources not getting deleted for unique ns cron

pradeepitm12:
Issue: add unique value for schedule as annotation to a namespace.
It creates a cronjob in that ns.
But when the job is getting actually scheduled it does not find
tekton-pipeline-controller sa. Which gives error

Fix: Instead of creating cronjob in other ns create it in the
installation target ns. There the sa is already available.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/common/prune.go
+++ b/pkg/reconciler/common/prune.go
@@ -44,6 +44,7 @@ const (
 	pruneKeepSince   = "operator.tekton.dev/prune.keep-since"
 	pruneSchedule    = "operator.tekton.dev/prune.schedule"
 	pruneCronLabel   = "tektonconfig.operator.tekton.dev/pruner"
+	pruneCronNsLabel = "tektonconfig.operator.tekton.dev/pruner.ns"
 	keep             = "keep"
 	keepSince        = "keep-since"
 )
@@ -67,7 +68,7 @@ func Prune(ctx context.Context, k kubernetes.Interface, tC *v1alpha1.TektonConfi
 	pruner := &Pruner{kc: k}
 
 	//to remove cronjob created by older version of operator.
-	oldCronName := CronName[7:] + "-" + tC.Spec.TargetNamespace
+	oldCronName := CronName[7:]
 	if err := pruner.checkAndDeleteCron(ctx, oldCronName, tC.Spec.TargetNamespace); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pradeepitm12:
cc: @vdemeester

pkg/reconciler/common/prune_test.go
```diff
@@ -122,6 +123,16 @@ func TestCompleteFlowPrune(t *testing.T) {
 	if len(cronjobs.Items) != 2 {
 		assert.Error(t, err, "number of cronjobs not correct")
 	}
+	client.CoreV1().Namespaces().Update(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-two"}}, metav1.UpdateOptions{})
```
vdemeester, line 126: According to the build error, we need to check this error 😉

vdemeester (approved):

pkg/reconciler/common/prune.go
```diff
@@ -249,7 +250,24 @@ func (pruner *Pruner) createJobContainers(nsConfig *pruneConfigPerNS, ns string)
...
+func (pruner *Pruner) checkAndCreate(ctx context.Context, cronName, targetNs, uniquePruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference, listOpt string) error {
+	cronList, err := pruner.listCronJobs(ctx, targetNs, listOpt)
+	if err != nil {
+		return err
+	}
+	if len(cronList.Items) == 1 {
+		if err := pruner.deleteCronJob(ctx, cronList.Items[0].Name, targetNs); err != nil {
+			return err
+		}
+		return createCronJob(ctx, pruner.kc, cronName, targetNs, uniquePruneNs, schedule, pruneContainers, oR)
```
nikhil-thomas, line 262:
```suggestion
```
this line could be safely removed, as the next line will handle the create action for both cases

pkg/reconciler/common/prune.go
```diff
@@ -249,7 +250,24 @@ func (pruner *Pruner) createJobContainers(nsConfig *pruneConfigPerNS, ns string)
...
+	}
+	if err := createCronJob(ctx, pruner.kc, cronName, targetNs, uniquePruneNs, schedule, pruneContainers, oR); err != nil {
+		return err
+	}
+	return nil
```
nikhil-thomas, line 267:
```suggestion
	if len(cronList.Items) == 1 {
		if err := pruner.deleteCronJob(ctx, cronList.Items[0].Name, targetNs); err != nil {
			return err
		}
	}
	return createCronJob(ctx, pruner.kc, cronName, targetNs, uniquePruneNs, schedule, pruneContainers, oR)
```
as the eventual create actions is same from the if block and the normal flow, let us keep only one statement for create.

nikhil-thomas: not critical, :technologist: :+1:

PR #510, opened 2021-11-18
fix pruning enabled by default

pradeepitm12:
adds the default value to instance of tektonconfig

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/instance.go
+++ b/pkg/reconciler/shared/tektonconfig/instance.go
@@ -99,6 +99,7 @@ func (tc tektonConfig) ensureInstance(ctx context.Context) {
 }
 
 func (tc tektonConfig) createInstance(ctx context.Context) error {
+	pruneKeep := uint(100)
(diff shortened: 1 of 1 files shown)


pradeepitm12:
@nikhil-thomas

savitaashture (commented):
:+1: 
/lgtm

vdemeester (commented):
I guess the default Pruner is not set in the `SetDefaults` because it's only for OpenShift so far, right ?

pradeepitm12:
> I guess the default Pruner is not set in the SetDefaults because it's only for OpenShift so far, right ?
This is a shared controller for both platforms.
And it is not added in the SetDefault as it will try to set the default values even if the user intentionally removes the pruner spec

vdemeester:
> And it is not added in the SetDefault as it will try to set the default values even if the user intentionally removes the pruner spec

Ah, good point 👍🏼

PR #541, opened 2021-12-02
[cherry-pick] Updates TektonInstallerSet to allow replica changes to a deployment

vdemeester:
# Changes

This now updates the installer set to allow replica changes, it does
it as follow
- when creating a deployment it compute hash of spec and add it as
annotation
- while computing hash it set replicas to a const so that next time we
can set replicas to same const and then calculate hash
- if hash is different which would mean something is changed in spec
other than hash
- so it revert the spec to orignal and updates the replicas to what was
on the cluster deployment
- it will allow replica change but revert any other change to spec as the
hash would change

(cherry picked from commit e20a2c9eb5015a1fc7456f460269d7b38dac8074)

/cc @tektoncd/operator-maintainers 

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/kubernetes/tektoninstallerset/install.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/install.go
@@ -21,10 +21,18 @@ import (
 	"strings"
 
 	mf "github.com/manifestival/manifestival"
+	"github.com/tektoncd/operator/pkg/reconciler/common"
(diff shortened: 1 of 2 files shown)


concaf (approved):

savitaashture (commented):
/lgtm

savitaashture (approved):

PR #545, opened 2021-12-07
operatorhub: call Makefile instead of importing

vdemeester:
# Changes

Having a `Makefile` in operatorhub means it should be able to run on
its own (without the root `Makefile`). This mean we shouldn't need to
import it, but we can rather call it using `make -C …`.

/cc @nikhil-thomas 

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/operatorhub/Makefile
+++ b/operatorhub/Makefile
@@ -7,6 +7,9 @@ $(BIN)/operator-sdk: | $(BIN) ; $(info $(M) getting operator-sdk)
 	@./operatorhub/tools/install_operator-sdk.sh $(BIN)
 
 .PHONY: operator-bundle
-operator-bundle: | $(OPERATOR_SDK)
+operator-bundle: | $(OPERATOR_SDK) openshift/release-artifacts/bundle/manifests
 	@$(info BUNDLE_ARGS: $(BUNDLE_ARGS))
-	@OPERATOR_SDK=$(OPERATOR_SDK) operatorhub/tools/bundle.py ${BUNDLE_ARGS}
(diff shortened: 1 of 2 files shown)


operatorhub/Makefile
```diff
@@ -7,6 +7,9 @@ $(BIN)/operator-sdk: | $(BIN) ; $(info $(M) getting operator-sdk)
...
 
 .PHONY: operator-bundle
-operator-bundle: | $(OPERATOR_SDK)
+operator-bundle: | $(OPERATOR_SDK) openshift/release-artifacts/bundle/manifests
 	@$(info BUNDLE_ARGS: $(BUNDLE_ARGS))
-	@OPERATOR_SDK=$(OPERATOR_SDK) operatorhub/tools/bundle.py ${BUNDLE_ARGS}
+	@OPERATOR_SDK=$(OPERATOR_SDK) ./tools/bundle.py ${BUNDLE_ARGS}
+
+openshift/release-artifacts/bundle/manifests:
+	mkdir -p openshift/release-artifacts/bundle/manifests
```
nikhil-thomas, line 15: is this target necessary. 

the bundle generation will create the `bundle/manifests` directory.
so i think here `mkdir -p openshift/release-artifacts/` should be sufficient.

In addtion, the bundle generation script should be the one creating the release-artifacts directory if it doesn't exist

vdemeester: Ah right, I took the latest folder, but most likely we only need to create `openshift/release-artifacts`

concaf (approved):

PR #549, opened 2021-12-08
Stop overriding system defaults in SSL_CERT_DIR

concaf:
# Changes

Prior to this commit, we used to set default value of SSL_CERT_DIR in
our TaskRun, etc pods to `/tekton-custom-certs`; while this worked for
certs mounted via the bundle config maps, this broke use cases where users
relied on default locations (like `/etc/ssl/certs`,
`/etc/pki/tls/certs`, etc) for already existing or user-mounted certificates.

This commit fixes that by setting SSL_CERT_DIR as a list of directories
which contains `/tekton-custom-certs` and the system default locations
as well.

# Release Notes

```release-note
System default locations for certificates are no longer overridden by SSL_CERT_DIR
```

### DIFF
--- a/pkg/reconciler/openshift/common/cabundle.go
+++ b/pkg/reconciler/openshift/common/cabundle.go
@@ -114,7 +114,7 @@ func ApplyCABundles(u *unstructured.Unstructured) error {
 			}
 		}
 
-		// we wiil mount the certs at this location so we don't override the existing certs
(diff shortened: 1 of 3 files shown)


nikhil-thomas:
/approve
:heavy_check_mark:  lgtm

vdemeester (approved):
/lgtm

PR #550, opened 2021-12-08
Adds PipelinesAsCode as an Addon for OpenShift Platform

sm43:
# Changes

This adds pac which from coming from openshift-pipelines/pipleines-as-code
as a component in TektonAddon.
It can be enabled/disabled through TektonConfig.

# Submitter Checklist

review them:

# Release Notes

```
Users will be now able to install PipelinesAsCode through TektonConfig's `spec.addon.enablePipelinesAsCode` field only on OpenShift Platform and by default its enabled 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ TEKTON_PIPELINE_VERSION ?= latest
 TEKTON_TRIGGERS_VERSION ?= latest
 TEKTON_DASHBOARD_VERSION ?= latest
 TEKTON_RESULTS_VERSION ?= v0.3.1 # latest returns an older version hence hard coding to v0.3.1 for now (tektoncd/results#138)
+PAC_VERSION ?= 0.5.2
 
 $(BIN)/ko: PACKAGE=github.com/google/ko/cmd/ko
 
@@ -87,7 +88,7 @@ bin/%: cmd/% FORCE
 
 .PHONY: get-releases
 get-releases: |
-	$Q ./hack/fetch-releases.sh $(TARGET) $(TEKTON_PIPELINE_VERSION) $(TEKTON_TRIGGERS_VERSION) $(TEKTON_DASHBOARD_VERSION) $(TEKTON_RESULTS_VERSION) || exit ;
+	$Q ./hack/fetch-releases.sh $(TARGET) $(TEKTON_PIPELINE_VERSION) $(TEKTON_TRIGGERS_VERSION) $(TEKTON_DASHBOARD_VERSION) $(TEKTON_RESULTS_VERSION) $(PAC_VERSION) || exit ;
 
 .PHONY: apply
 apply: | $(KO) $(KUSTOMIZE) get-releases ; $(info $(M) ko apply on $(TARGET)) @ ## Apply config to the current cluster
--- a/pkg/apis/operator/v1alpha1/tektonaddon_defaults.go
+++ b/pkg/apis/operator/v1alpha1/tektonaddon_defaults.go
@@ -18,33 +18,35 @@ package v1alpha1
 
 import (
 	"context"
+
+	"knative.dev/pkg/ptr"
 )
 
 func (ta *TektonAddon) SetDefaults(ctx context.Context) {
-	setAddonDefaults(&ta.Spec.Params)
+	setAddonDefaults(&ta.Spec.Addon)
 }
 
-func setAddonDefaults(params *[]Param) {
+func setAddonDefaults(addon *Addon) {
 
-	paramsMap := ParseParams(*params)
+	paramsMap := ParseParams(addon.Params)
 	_, ptOk := paramsMap[PipelineTemplatesParam]
 	ct, ctOk := paramsMap[ClusterTasksParam]
 
 	// If clusterTasks is false and pipelineTemplate is not set, then set it as false
 	// as pipelines templates are created using clusterTasks
 	if ctOk && (ct == "false" && !ptOk) {
-		*params = append(*params, Param{
+		addon.Params = append(addon.Params, Param{
 			Name:  PipelineTemplatesParam,
 			Value: "false",
 		})
-		paramsMap = ParseParams(*params)
+		paramsMap = ParseParams(addon.Params)
 	}
 
 	// set the params with default values if not set in cr
 	for d := range AddonParams {
 		_, ok := paramsMap[d]
 		if !ok {
-			*params = append(*params,
+			addon.Params = append(addon.Params,
 				Param{
 					Name:  d,
 					Value: AddonParams[d].Default,
@@ -52,4 +54,8 @@ func setAddonDefaults(params *[]Param) {
 		}
 	}
 
+	// by default enable pac
+	if addon.EnablePAC == nil {
+		addon.EnablePAC = ptr.Bool(true)
+	}
 }
--- a/pkg/apis/operator/v1alpha1/tektonaddon_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonaddon_types.go
@@ -71,7 +71,7 @@ type Addon struct {
 	// Params is the list of params passed for Addon customization
 	// +optional
 	Params []Param `json:"params,omitempty"`
-	// InstallPipelinesAsCode field defines whether to install PAC
+	// EnablePAC field defines whether to install PAC
 	// +optional
 	EnablePAC *bool `json:"enablePipelinesAsCode,omitempty"`
 }
--- a/pkg/apis/operator/v1alpha1/tektonaddon_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonaddon_types.go
@@ -47,9 +47,7 @@ func (tp *TektonAddon) GetStatus() TektonComponentStatus {
 // TektonAddonSpec defines the desired state of TektonAddon
 type TektonAddonSpec struct {
 	CommonSpec `json:",inline"`
-	// The params to customize different components of Addon
-	// +optional
-	Params []Param `json:"params,omitempty"`
+	Addon      `json:",inline"`
 	// Config holds the configuration for resources created by Addon
 	// +optional
 	Config Config `json:"config,omitempty"`
@@ -73,6 +71,9 @@ type Addon struct {
 	// Params is the list of params passed for Addon customization
 	// +optional
 	Params []Param `json:"params,omitempty"`
+	// InstallPipelinesAsCode field defines whether to install PAC
+	// +optional
+	EnablePAC *bool `json:"enablePipelinesAsCode,omitempty"`
 }
 
 func (a Addon) IsEmpty() bool {
--- a/pkg/reconciler/openshift/tektonaddon/tektonaddon.go
+++ b/pkg/reconciler/openshift/tektonaddon/tektonaddon.go
@@ -322,6 +322,27 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ta *v1alpha1.TektonAddon
 		return nil
 	}
 
+	// Check if PAC is enabled
+
+	if *ta.Spec.EnablePAC {
+
+		// make sure pac is installed
+		exist, err := checkIfInstallerSetExist(ctx, r.operatorClientSet, r.version,
+			fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, PACInstallerSet))
+		if err != nil {
+			return err
+		}
+		if !exist {
+			return r.ensurePAC(ctx, ta)
+		}
+
+	} else {
+		// if disabled then delete the installer Set if exist
+		if err := r.deleteInstallerSet(ctx, fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, PACInstallerSet)); err != nil {
+			return err
+		}
+	}
+
 	ta.Status.MarkInstallerSetReady()
 
 	if err := r.extension.PostReconcile(ctx, ta); err != nil {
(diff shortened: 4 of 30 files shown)


pkg/reconciler/openshift/tektonpipelinesascode/extension.go
```diff
@@ -0,0 +1,43 @@
...
+//func (oe openshiftExtension) PreReconcile(context.Context, v1alpha1.TektonComponent) error {
+//	return nil
+//}
+//func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.TektonComponent) error {
+//
+//	return nil
+//}
+//func (oe openshiftExtension) Finalize(context.Context, v1alpha1.TektonComponent) error {
+//	return nil
+//}
```
nikhil-thomas, line 43: shall we keep the `return nil` implementation for all these and return an instance of openshiftExtension instead of common.NoExtension. 

I think it makes it clear that where to add OpenShift specific changes in the future.

pkg/reconciler/kubernetes/tektonpipelinesascode/controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+func fetchVersion(manifest mf.Manifest) (string, error) {
+	crds := manifest.Filter(mf.CRDs)
+
+	if len(crds.Resources()) == 0 {
+		return "", fmt.Errorf("failed to find CRD in source manifest")
+	}
+
+	labels := crds.Resources()[0].GetLabels()
+	version, ok := labels["app.kubernetes.io/version"]
```
nikhil-thomas, line 111: may be we should create an issue in PAC to create an "info" configMap which holds version information, like other components.

https://github.com/tektoncd/triggers/blob/d360e8f7e51ca97d4cc1fc072c955a5365a865f6/config/config-info.yaml#L29

Makefile
```diff
@@ -87,7 +88,7 @@ bin/%: cmd/% FORCE
 
 .PHONY: get-releases
 get-releases: |
-	$Q ./hack/fetch-releases.sh $(TARGET) $(PIPELINES_VERSION) $(TRIGGERS_VERSION) $(DASHBOARD_VERSION) $(RESULTS_VERSION) || exit ;
+	$Q ./hack/fetch-releases.sh $(TARGET) $(PIPELINES_VERSION) $(TRIGGERS_VERSION) $(DASHBOARD_VERSION) $(RESULTS_VERSION) $(PAC_VERSION) || exit ;
```
nikhil-thomas, line 91: may be it is time to think about a config file (eg: config.sh) to capture the version arguments for this target/script.
note: not a blocker for this patch 🧑‍💻

nikhil-thomas:
shall we add the controlelr to `cmd/kubernetes/operator/main.go` also.

pkg/reconciler/kubernetes/tektonpipelinesascode/tektonpipelinesascode.go
```diff
@@ -0,0 +1,274 @@
...
+
+		manifest := r.manifest
+		if err := r.transform(ctx, &manifest, tpac); err != nil {
+			logger.Error("manifest transformation failed:  ", err)
+			return err
+		}
+
+		return createInstallerSet(ctx, r.operatorClientSet, tpac, r.manifest, r.version, "pac")
+	}
+
```
nikhil-thomas, line 144: don't we need the upgrade scenario as well. ie, `installerset` exist but the releaseVersion has changed

pkg/reconciler/kubernetes/tektonpipelinesascode/tektonpipelinesascode.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+	}
+
+	return nil
+}
+
+// checkIfInstallerSetExist checks if installer set exists for a component and return true/false based on it
+// and if installer set which already exist is of older version then it deletes and return false to create a new
+// installer set
+func checkIfInstallerSetExist(ctx context.Context, oc clientset.Interface, relVersion string, tpac *v1alpha1.TektonPipelinesAsCode) (bool, error) {
```
nikhil-thomas, line 192:
```suggestion
func checkIfValidInstallerSetExist(ctx context.Context, oc clientset.Interface, relVersion string, tpac *v1alpha1.TektonPipelinesAsCode) (bool, error) {
```

as this function is doing both "exists" check and "latest version" check, let us make the name proclaim it.

pkg/reconciler/kubernetes/tektonpipelinesascode/tektonpipelinesascode.go
nikhil-thomas: oh, i see, https://github.com/tektoncd/operator/pull/550/files#r766263482

nikhil-thomas: but, how will the targetNamespace changed (or any spec) scenario be handled.

pkg/reconciler/kubernetes/tektonpipelinesascode/controller.go
```diff
@@ -0,0 +1,105 @@
...
+func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	return NewExtendedController(common.NoExtension)(ctx, cmw)
+}
+
+// NewExtendedController returns a controller extended to a specific platform
+func NewExtendedController(generator common.ExtensionGenerator) injection.ControllerConstructor {
+	return func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+		logger := logging.FromContext(ctx)
+
+		mfclient, err := mfc.NewClient(injection.GetConfig(ctx))
```
PuneetPunamiya, line 53: Should we use the init controller to fetch the manifest ?

pkg/reconciler/kubernetes/tektonpipelinesascode/controller.go
```diff
@@ -0,0 +1,105 @@
...
+		manifest, err := mf.ManifestFrom(mf.Slice{}, mf.UseClient(mfclient), mf.UseLogger(mflogger))
+		if err != nil {
+			logger.Fatalw("Error creating initial manifest", zap.Error(err))
+		}
+
+		if err := fetchSourceManifests(ctx, &manifest); err != nil {
+			logger.Fatalw("failed to read manifest", err)
+		}
+
+		version := os.Getenv(versionKey)
```
PuneetPunamiya: Any reason we are reading the version from env, I mean doesn't pac has a config map which has version-info in it ?

PuneetPunamiya: If not can't it get the value from common.TargetVersion ?

sm43: pac doesn't have info configmap currently, it is reading from crd label. 
this is an older commit

pkg/reconciler/kubernetes/tektonpipelinesascode/controller.go
sm43: once pac has info configmap we will switch to init

sm43:
> .

yes but due to https://github.com/openshift-pipelines/pipelines-as-code/issues/385
I kept it on hold
or we will have to give the operator sa on k8s side some extra permission to create this clusterrole

pkg/reconciler/kubernetes/tektonpipelinesascode/tektonpipelinesascode.go
sm43: updated !

nikhil-thomas:
✅ lgtm

config/base/kustomization.yaml
```diff
@@ -18,6 +18,7 @@ resources:
 - 300-operator_v1alpha1_pipeline_crd.yaml
 - 300-operator_v1alpha1_trigger_crd.yaml
 - 300-operator_v1alpha1_config_crd.yaml
+- 300-operator_v1alpha1_pac_crd.yaml
```
savitaashture, line 21: @sm43 as this PR focuses on adding PAC for openshift shall we move this to config/openshift folder only

config/crs/kubernetes/pipelines-as-code/operator_v1alpha1_addon_cr.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: operator.tekton.dev/v1alpha1
+kind: TektonPipelinesAsCode
+metadata:
+  name: pac
+spec:
+  targetNamespace: pipelines-as-code
```
savitaashture, line 20: We have same yaml under `crs/kubernetes` and `crs/openshift` if nothing is different should we keep in common place?

config/crs/openshift/pipelines-as-code/operator_v1alpha1_addon_cr.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: operator.tekton.dev/v1alpha1
+kind: TektonPipelinesAsCode
+metadata:
+  name: pac
+spec:
+  targetNamespace: pipelines-as-code
```
savitaashture, line 20: same here

config/base/kustomization.yaml
sm43: yep make sense. updated

config/crs/kubernetes/pipelines-as-code/operator_v1alpha1_addon_cr.yaml
sm43: currently for all components we do that way 
lets address it separately

pradeepitm12:
@sm43 :wave: 
It seems the pr title is a bit confusing to others 
can we change `Adds TektonPipelinesAsCode Api` --> `Adds PipelinesAsCode Api`
as there are other efforts also going on in the same direction.

dibyom:
> It seems the pr title is a bit confusing to others


To clarify, the confusion here is that pipelines as code today is an openshift extension while the config options here make it seem like its part of the Tekton API surface. I think generally Tekton projects should only contain things that are part of Tekton and not vendor specific extensions. Operator is interesting in that it installs other projects - I think its fine for the operator to be able to install vendor specific extensions but it should be clear that they are not part of the core tekton API surface. I think we should ensue that the PipelinesAsCode is only part of the openshift extension (until PAC is part of tekton itself) and that it is clear to the user that it is not part of Tekton API (e.g. not using the tekton.dev group for the CRDs).

dibyom:
Whoops...did not mean to close it 😅

nikhil-thomas:
> > It seems the pr title is a bit confusing to others
> 
> To clarify, the confusion here is that pipelines as code today is an openshift extension while the config options here make it seem like its part of the Tekton API surface. I think generally Tekton projects should only contain things that are part of Tekton and not vendor specific extensions. Operator is interesting in that it installs other projects - I think its fine for the operator to be able to install vendor specific extensions but it should be clear that they are not part of the core tekton API surface. I think we should ensue that the PipelinesAsCode is only part of the openshift extension (until PAC is part of tekton itself) and that it is clear to the user that it is not part of Tekton API (e.g. not using the tekton.dev group for the CRDs).
> 
> Update: @vdemeester corrected me - PAC also works on vanilla k8s. I think it should still be clear to the user whether they are installing something that is part of Tekton itself vs an addon/extension to Tekton :)

I agree 🧑‍💻 👍 

At present, this patch adds the new Pipeline-as-a-Code only to OpenShift builds. So in our github releases, [operatorhub.io releases](https://operatorhub.io/operator/tektoncd-operator), the PAC Api and reconciler will not be included. 

In the future, if we add something similar to the builds for Kubernetes, then we shall definitely make it clear that the API is a vendor specific extension. 

In the future, if we have similar API from Tektoncd we can keep that as the main API and move the the differences to vendor specific reconciler extension mechanism that the operator currently supports.

config/crs/openshift/pipelines-as-code/operator_v1alpha1_addon_cr.yaml
savitaashture: @sm43 I think need to change name of the file as well

sm43: updated !

sm43:
hi @dibyom, we understand your concern and based on feedback
we have decided to move PipelinesAsCode in TektonAddon.
So, it would be installed as a part of TektonAddon. Currently, TektonAddon is enabled only for OpenShift platform.
TektonAddon is not available in Kubernetes platform build and won't be there in the Operator release.
let us know if you have any other concern. 🙂

dibyom:
@sm43 Thank you! Appreciate you all taking the feedback into account!

pkg/apis/operator/v1alpha1/tektonaddon_defaults.go
```diff
@@ -18,38 +18,44 @@ package v1alpha1
...
-			*params = append(*params,
+			addon.Params = append(addon.Params,
 				Param{
 					Name:  d,
 					Value: AddonParams[d].Default,
 				})
 		}
 	}
 
+	// by default enable pac
```
savitaashture, line 57: Could you add this info to release note as well

savitaashture:
@sm43 could you add e2e test also to verify pac installation

pkg/apis/operator/v1alpha1/tektonaddon_types.go
```diff
@@ -73,6 +71,9 @@ type Addon struct {
 	// Params is the list of params passed for Addon customization
 	// +optional
 	Params []Param `json:"params,omitempty"`
+	// InstallPipelinesAsCode field defines whether to install PAC
```
savitaashture, line 74:
```suggestion
	// EnablePAC field defines whether to install PAC
```

pkg/apis/operator/v1alpha1/tektonaddon_types.go
```diff
@@ -73,6 +71,9 @@ type Addon struct {
...
+	// +optional
+	EnablePAC *bool `json:"enablePipelinesAsCode,omitempty"`
```
savitaashture, line 76: could you add information about this field in the [TektonAddon.md](https://github.com/tektoncd/operator/blob/main/docs/TektonAddon.md#params) doc file

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -280,6 +301,28 @@ func (r *Reconciler) checkComponentStatus(ctx context.Context, labelSelector str
 	return nil
 }
 
+func (r *Reconciler) ensurePAC(ctx context.Context, ta *v1alpha1.TektonAddon) error {
+	pacManifest := r.manifest
```
nikhil-thomas, line 305:
```suggestion
	pacManifest := mf.Manifest{}
```

at present `r.manifest`is empty. But if that changes it could break this function

savitaashture:
LGTM

PR #554, opened 2021-12-08
[cherry-pick] Treat SSL_CERT_DIR as a colon separated list

concaf:
Prior to this commit, we used to treat SSL_CERT_DIR as a string
containing one directory but it is/can be a colon separated list of
directories. This commit fixes that assumption.


vdemeester (approved):

savitaashture (commented):
:+1: 

/lgtm

Issue #539, opened 2021-12-01
Add an `operator-info` config map in "targetNamespace"


nikhil-thomas:
### Feature request

Operator can add an `operator-info` config map in the namespace where Tekton components are installed (targetNamespace). So that we can try to support printing Operator version in the output of `tkn version` command.

### Use case

While working from CLI there is no easy way to find version of operator. This is important especially on OpenShift as the operator version indicates whether an upgrade has happened and whether an awaited bugfix/feature is availabe on a cluster.

### Reference

- https://github.com/tektoncd/pipeline/pull/3971

nikhil-thomas:
Related issue in Tekton CLI: Add Operator version in the output of tkn version command #1473

PR #563, opened 2021-12-15
Adds config map for operator version

PuneetPunamiya:
- This patch adds a config map which has the operator verison.
    The config map is created in the tektonConfig pre-reconciler

  - First it checks in the tektonConfig reconciler if the targetNs
    exists, if not it creates the targetNs and then in the pipeline
    reconciler if the tektonConfig instance exists then just add the
    labels from the release yamls if the targetNs is `tekton-pipelines`

Fixes: https://github.com/tektoncd/operator/issues/539

# Changes

# Submitter Checklist

review them:

# Release Notes
 ```release-note
 - Adds a config map which has the operator verison
 - This config map can be used by `tkn cli` which would give
   the operator version in the output of `tkn version` command
```

### DIFF
--- a/cmd/openshift/operator/kodata/info/operator-config.yaml
+++ b/cmd/openshift/operator/kodata/info/operator-config.yaml
@@ -1,0 +1,24 @@
+# Copyright 2022 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operators-info
+  labels:
+    app.kubernetes.io/instance: default
+data:
+  # Contains operator version which can be queried by external
+  # tools such as CLI.
+  version: 'devel'
--- a/pkg/reconciler/kubernetes/tektonpipeline/controller.go
+++ b/pkg/reconciler/kubernetes/tektonpipeline/controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"github.com/tektoncd/operator/pkg/reconciler/kubernetes/initcontroller"
 	"k8s.io/client-go/tools/cache"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection"
@@ -60,6 +61,7 @@ func NewExtendedController(generator common.ExtensionGenerator) injection.Contro
 
 		c := &Reconciler{
 			operatorClientSet: operatorclient.Get(ctx),
+			kubeClientSet:     kubeclient.Get(ctx),
 			extension:         generator(ctx),
 			manifest:          manifest,
 			releaseVersion:    releaseVersion,
--- a/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
+++ b/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
@@ -34,6 +34,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/logging"
 	pkgreconciler "knative.dev/pkg/reconciler"
@@ -63,7 +64,8 @@ type Reconciler struct {
 	// releaseVersion describes the current pipelines version
 	releaseVersion string
 	// metrics handles metrics for pipeline install
-	metrics *Recorder
+	metrics       *Recorder
+	kubeClientSet kubernetes.Interface
 }
 
 // Check that our Reconciler implements controller.Reconciler
@@ -115,6 +117,10 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
 
 	// Pass the object through defaulting
 	tp.SetDefaults(ctx)
+
+	if err := r.targetNamespaceCheck(ctx, tp); err != nil {
+		return err
+	}
 
 	if err := r.extension.PreReconcile(ctx, tp); err != nil {
 		tp.Status.MarkPreReconcilerFailed(fmt.Sprintf("PreReconciliation failed: %s", err.Error()))
@@ -329,6 +335,22 @@ func makeInstallerSet(tp *v1alpha1.TektonPipeline, manifest mf.Manifest, tpSpecH
 			Manifests: manifest.Resources(),
 		},
 	}
+}
+
+func (r *Reconciler) targetNamespaceCheck(ctx context.Context, tp *v1alpha1.TektonPipeline) error {
+	labels := r.manifest.Filter(mf.ByKind("Namespace")).Resources()[0].GetLabels()
+
+	ns, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, tp.GetSpec().GetTargetNamespace(), metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+	for key, value := range labels {
+		ns.Labels[key] = value
+	}
+	_, err = r.kubeClientSet.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
+	return err
 }
 
 // transform mutates the passed manifest to one with common, component
--- a/pkg/reconciler/shared/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/shared/tektonconfig/tektonconfig.go
@@ -19,13 +19,19 @@ package tektonconfig
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 
+	corev1 "k8s.io/api/core/v1"
+
+	mf "github.com/manifestival/manifestival"
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
 	clientset "github.com/tektoncd/operator/pkg/client/clientset/versioned"
 	tektonConfigreconciler "github.com/tektoncd/operator/pkg/client/injection/reconciler/operator/v1alpha1/tektonconfig"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"github.com/tektoncd/operator/pkg/reconciler/shared/tektonconfig/pipeline"
 	"github.com/tektoncd/operator/pkg/reconciler/shared/tektonconfig/trigger"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/logging"
@@ -40,6 +46,7 @@ type Reconciler struct {
 	operatorClientSet clientset.Interface
 	// Platform-specific behavior to affect the transform
 	extension common.Extension
+	manifest  mf.Manifest
 }
 
 // Check that our Reconciler implements controller.Reconciler
@@ -88,6 +95,14 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 
 	tc.SetDefaults(ctx)
 
+	if err := r.ensureTargetNamespaceExists(ctx, tc); err != nil {
+		return err
+	}
+
+	if err := r.createOperatorVersionConfigMap(tc); err != nil {
+		return err
+	}
+
 	if err := r.extension.PreReconcile(ctx, tc); err != nil {
 		if err == v1alpha1.RECONCILE_AGAIN_ERR {
 			return err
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

**Comment by pradeepitm12** [comment]
@PuneetPunamiya Thanks for the pr. Can you please add a commit message explaining what this pr is doing!
This not only helps in the review but also to know what things got in by just looking at the commit message.

**Comment by PuneetPunamiya** [comment]
> @PuneetPunamiya Thanks for the pr. Can you please add a commit message explaining what this pr is doing! This not only helps in the review but also to know what things got in by just looking at the commit message.

Yeah sure, since the implementation was not complete yet and I have been working on it hence did just created a wip pr, once the implementation is complete will add a proper commit message which describe what the patch does

File: pkg/reconciler/kubernetes/tektonconfig/extension.go [hunk 1]
```diff
@@ -18,29 +18,72 @@ package tektonconfig
...
+func (oe kubernetesExtension) PreReconcile(ctx context.Context, tc v1alpha1.TektonComponent) error {
+
+	koDataDir := os.Getenv(common.KoEnvKey)
+	operatorDir := filepath.Join(koDataDir, "operator")
+
+	if err := common.AppendManifest(oe.manifest, operatorDir); err != nil {
+		return err
+	}
+
+	manifest := oe.manifest.Append()
```
sm43, line 73: we are not passing anything to `Append()` ... we can skip this

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -18,29 +18,72 @@ package tektonconfig
...
 type kubernetesExtension struct {
 	operatorClientSet versioned.Interface
+	manifest          *mf.Manifest
 }
 
 func (oe kubernetesExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Transformer {
 	return []mf.Transformer{}
 }
-func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonComponent) error {
+func (oe kubernetesExtension) PreReconcile(ctx context.Context, tc v1alpha1.TektonComponent) error {
```
sm43, line 64: I see configmap for the openshift and k8s 
but this code is in k8s extension?
do we not need to add in openshift part?

pkg/reconciler/common/common.go
```diff
@@ -39,6 +39,8 @@ const (
 	TriggerNotReady        = "tekton-triggers not ready"
 	TriggerNotFound        = "tekton-triggers not installed"
 	NamespaceIgnorePattern = "^(openshift|kube)-"
+	ConfigNotFound         = "tekton config not found"
```
nikhil-thomas, line 42:
```suggestion
	ConfigNotFound         = "tekton-config not found"
```

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -63,3 +106,13 @@ func (oe kubernetesExtension) Finalize(ctx context.Context, comp v1alpha1.Tekton
...
+
+func changeNamespace(targetNamespace string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetNamespace() != targetNamespace {
+			u.SetNamespace(targetNamespace)
+			return nil
+		}
+		return nil
+	}
+}
```
nikhil-thomas, line 118: we do have an already exiting transformer to do this, don't we .
why don't we reuse it.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -88,6 +91,23 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+	}
+	_, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, tc.GetSpec().GetTargetNamespace(), metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			if _, err = r.kubeClientSet.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{}); err != nil {
+				return err
+			}
+		}
+	}
+
```
nikhil-thomas, line 110:
```suggestion
	r.createTargetNamespace()
```

please move this into a function.

There are 2 ways how "functions" help us in coding:

1. the obvious, code reuse.
2. greatly improved readability. ie, when some one reads the ReconcileKind() what need to be seen is something like `ensureTargetNamespace()`. Keeping the full implementation here is distracting.

Whenever we write code, if a sequence of code lines explain more than the intention of the current function, we are going too low level. That is, the implementaiton of a sub mechanisms could be written as functions even if they might not be reused else where.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -124,6 +128,26 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
...
+			}
+			_, err = r.kubeClientSet.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		r.manifest = r.manifest.Filter(mf.Not(mf.ByKind("Namespace")))
+	}
+
```
nikhil-thomas, line 150: 1. move all this into a function (ref: https://github.com/tektoncd/operator/pull/563/files#r778478046)
2. this logic should run before the `r.extension.Prereconcile()`

cmd/kubernetes/operator/kodata/operator/operator-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operators-info
+  labels:
+    app.kubernetes.io/instance: default
+data:
+  # Contains operator version which can be queried by external
+  # tools such as CLI.
+  version: 'devel'
```
nikhil-thomas, line 24: please change the name of the sub directory from `operator` to something more informational.
eg:
~cmd/kubernetes/operator/kodata/operator/~
`cmd/kubernetes/operator/kodata/info`

cmd/openshift/operator/kodata/operator/operator-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operators-info
+  labels:
+    app.kubernetes.io/instance: default
+data:
+  # Contains operator version which can be queried by external
+  # tools such as CLI.
+  version: 'devel'
```
nikhil-thomas, line 24: https://github.com/tektoncd/operator/pull/563/files#r778479787

pkg/reconciler/kubernetes/tektonconfig/extension.go
```diff
@@ -18,29 +18,72 @@ package tektonconfig
...
 }
 
 func (oe kubernetesExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Transformer {
 	return []mf.Transformer{}
 }
-func (oe kubernetesExtension) PreReconcile(context.Context, v1alpha1.TektonComponent) error {
+func (oe kubernetesExtension) PreReconcile(ctx context.Context, tc v1alpha1.TektonComponent) error {
+
+	koDataDir := os.Getenv(common.KoEnvKey)
+	operatorDir := filepath.Join(koDataDir, "operator")
```
nikhil-thomas, line 67: let us handle this (creation of info configMap) in the main reconciler it self as it is common to all platforms.
let us add this logic in the TektonConfig main reconciler, after creating the TargetNamespace, but before calling the  prereconcile.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -116,6 +120,10 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
 	// Pass the object through defaulting
 	tp.SetDefaults(ctx)
 
+	if err := r.ensureTektonConfigExists(ctx, tp); err != nil {
```
sm43, line 123: what are we trying to do here using this func?

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -138,3 +154,45 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+	return nil
+}
+
+func (r *Reconciler) ensureTargetNamespaceExists(ctx context.Context, tc *v1alpha1.TektonConfig) error {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: tc.GetSpec().GetTargetNamespace(),
+			Labels: map[string]string{
+				"operator.tekton.dev/targetNamespace": tc.GetSpec().GetTargetNamespace(),
```
nikhil-thomas, line 185: what is the purpose of this label.

🧑‍💻 if it is to query kubernetes namespaces api and find the current tekton targetnamespace, then the `value` part should be something constant irrespective of the value of targetNamespace field in the CR.

eg: `"operator.tekton.dev/targetNamespace": true`

if the `value` part of this label could change, then this label becomes useless.

In addition, if we already know the name then why don't we just `get` the namespace using it as the metadata.name will already have the name.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -138,3 +154,45 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+			},
+		},
+	}
+	_, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, tc.GetSpec().GetTargetNamespace(), metav1.GetOptions{})
```
nikhil-thomas, line 189: here we have to handle the secanario where the `spec.targetNamespace` being changed by a user.

so ideally, the flow should be, 
1. query (list) namespaces using a label (eg: operator.tekton.dev/targetNamespace: true)
2. if list is non-empty, check if any of the namespaces has name == spec.targetNamespace 
     2.a. if for any namespace in the list name != spec.targetNamespace, then delete it
3. if spec.targetNamespace exists (in the list) then return nil
4. else create spec.targetNamespace

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -138,3 +154,45 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+	if err != nil {
+		return err
+	}
+
+	if err = manifest.Apply(); err != nil {
+		return err
+	}
+
+	return nil
+}
```
nikhil-thomas, line 178: 👍

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -331,6 +339,29 @@ func makeInstallerSet(tp *v1alpha1.TektonPipeline, manifest mf.Manifest, tpSpecH
 	}
 }
 
+func (r *Reconciler) ensureTektonConfigExists(ctx context.Context, tp *v1alpha1.TektonPipeline) error {
```
nikhil-thomas, line 342: rename this function to something like (`targetNamespaceCheck` )
as our intention is to so 2 things:
1. if targetNamespace exists add extra labels (and remove it from manifest list)
2. keep it in manifest.


so now i think we can completey skip checking the status of TektonConfig as we are only interested in the namespace.

so let us use a `get` api call to check if the TektonPipeline spec.targetNamespace exists.
if yes then apply new labels then filter out namespaces from manifest
if no then do nothing and continue.

note: it is a good practice to use names to communicate the purpose of a function or variable

nikhil-thomas:
@PuneetPunamiya we need one modification in one of the Tasks in Operator release pipeline.

A sed command below the already existing ones here: https://github.com/tektoncd/operator/blob/2217ee6cdf0fb31842459a7b339d88fcbf8d5049/tekton/build-publish-images-manifests.yaml#L122

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -138,3 +152,64 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+func (r *Reconciler) createOperatorVersionConfigMap(tc *v1alpha1.TektonConfig) error {
+	koDataDir := os.Getenv(common.KoEnvKey)
+	operatorDir := filepath.Join(koDataDir, "info")
+
+	if err := common.AppendManifest(&r.manifest, operatorDir); err != nil {
+		return err
+	}
+
+	manifest, err := r.manifest.Transform(
+		mf.InjectNamespace(tc.GetSpec().GetTargetNamespace()),
```
nikhil-thomas, line 165: could you make sure that he ownerReference is also set to TektonConfig instance. So that we can make sure this config map is cascade-deleted when we delete the instance (name:config) of TektonConfig

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -138,3 +152,64 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+	if err != nil {
+		return err
+	}
+
+	if len(nps.Items) > 0 {
+		for _, j := range nps.Items {
+			if j.Name != tc.GetSpec().GetTargetNamespace() {
+				if err := r.kubeClientSet.CoreV1().Namespaces().Delete(ctx, tc.GetSpec().GetTargetNamespace(), metav1.DeleteOptions{}); err != nil {
+					return err
+				}
```
nikhil-thomas, line 193: Functionaly the logic here is correct. However, this looks a bit agressive. I mean, here we are deleting previous targetNamespace(s) before components get a chance to properly finalize/move items in their respective reconcilers.

My suggestion is to split what this function does into 2 funcitons.

1. In this `ensureTargetNamespaceExists` function instead of deleting the previous targetNamespaces add an additional label to them `"operator.tekton.dev/targetNamespace/mark-for-deletion": "true"`
2. Write a second function `deleteObsoleteTargetNamespaces()`
     Then query namespaces with `"operator.tekton.dev/targetNamespace/mark-for-deletion": "true"` and then loop and delete them. 
    👉 call `deleteObsoleteTargetNamespaces()` at the end of TektonConfig reconciler just before the final update call ([here](https://github.com/tektoncd/operator/blob/2217ee6cdf0fb31842459a7b339d88fcbf8d5049/pkg/reconciler/shared/tektonconfig/tektonconfig.go#L133)) . 
Note: if there is an error in deleting ohsolete namespaces, then let us log it and continue, let us not add a status field for that now.

nikhil-thomas:
@PuneetPunamiya overall this patch looks good. 

Could you make 🧑‍💻 3 more changes
1. https://github.com/tektoncd/operator/pull/563#issuecomment-1015924367
2. https://github.com/tektoncd/operator/pull/563#discussion_r787252939
3. https://github.com/tektoncd/operator/pull/563#discussion_r787246096

tekton/build-publish-images-manifests.yaml
```diff
@@ -123,6 +123,7 @@ spec:
       sed -i -e 's/\(operator.tekton.dev\/release\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\(app.kubernetes.io\/version\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\(version\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\("-version"\), "devel"/\1, "$(params.versionTag)"/g' ${PROJECT_ROOT}/config/base/*.yaml
       sed -i -e 's/\(operator.tekton.dev\/release\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\(app.kubernetes.io\/version\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\(version\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\("-version"\), "devel"/\1, "$(params.versionTag)"/g' ${PROJECT_ROOT}/config/${KUBE_DISTRO}/base/*.yaml
       sed -i -e 's/\(operator.tekton.dev\/release\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\(app.kubernetes.io\/version\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\(version\): "devel"/\1: "$(params.versionTag)"/g' -e 's/\("-version"\), "devel"/\1, "$(params.versionTag)"/g' ${PROJECT_ROOT}/config/${KUBE_DISTRO}/overlays/default/*.yaml
+      sed -i -e 's/\(version\): "devel"/\1: "$(params.versionTag)"/g' ${PROJECT_ROOT}/cmd/${KUBE_DISTRO}/operator/kodata/info/operator-config.yaml
```
nikhil-thomas, line 126: 👍

nikhil-thomas:
✅ lgtm

nikhil-thomas:
@PuneetPunamiya could you add a release note in the PR description

format

```
```release-note
release note line(s)
```
```

File: cmd/openshift/operator/kodata/info/operator-config.yaml [hunk 18]
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operators-info
```
savitaashture, line 18: @PuneetPunamiya I see `operator-config.yaml` is same for both Openshift and kubernetes 
So can you keep it in common place as both are same

pkg/reconciler/kubernetes/tektonpipeline/controller.go
```diff
@@ -18,6 +18,7 @@ package tektonpipeline
 
 import (
 	"context"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
```
savitaashture, line 21:
```suggestion
	"context"

	kubeclient "knative.dev/pkg/client/injection/kube/client"
```

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	stdError "errors"
 	"fmt"
+	"k8s.io/client-go/kubernetes"
 	"time"
```
savitaashture, line 24:
```suggestion
	"context"
	stdError "errors"
	"fmt"
	"time"

	"k8s.io/client-go/kubernetes"
```
How `go fmt` dint catch this :thinking:

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -331,6 +337,25 @@ func makeInstallerSet(tp *v1alpha1.TektonPipeline, manifest mf.Manifest, tpSpecH
...
+		}
+	}
+	for key, value := range labels {
+		ns.Labels[key] = value
+	}
+	_, err = r.kubeClientSet.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
```
savitaashture, line 356:
```suggestion
	_, err = r.kubeClientSet.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{})
	return err
```

cmd/openshift/operator/kodata/info/operator-config.yaml
nikhil-thomas: as this is in `cmd/<platform>/.../kodata` (not `config/<platform>/) let us keep this as separate files. 

otherwise, we will have to setup kustomize hierarchy for `kodata` also like we do for `config/`. I think that will be too much effort for just 1 file. ie, setuping kustomize and related changes in release pipeline to set this file in kodata.

🧑‍💻  In the future, if we have more common files like this, we can think about a script/kustomize to handle this.

vdemeester (approved):

cmd/openshift/operator/kodata/info/operator-config.yaml
savitaashture: Okay
Ya agree i mean require some effort to do it

savitaashture (commented):
Thank you :+1: :tada: 

/lgtm

PR #564, opened 2021-12-15
fix cronjob creation with irrelevant config change

pradeepitm12:
observed behaviour :
  crojob get recreated with additon/update in the prune config
  annotaions, but it also gets recreated even if irrelevant change
  happen in any namespace.
fix:
  add config hash which is applied, and check it next reconcile it
  checks the last applied hash with the current config hash and
  only create cronjobs if oldhash not equal to current hash

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/common/prune.go
+++ b/pkg/reconciler/common/prune.go
@@ -54,8 +54,10 @@ const (
 )
 
 type Pruner struct {
-	kc       kubernetes.Interface
-	tknImage string
+	kc              kubernetes.Interface
+	tknImage        string
+	targetNamespace string
+	ownerRef        v1.OwnerReference
 }
 
 type pruneConfigPerNS struct {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/common/prune.go
```diff
@@ -120,7 +124,7 @@ func Prune(ctx context.Context, k kubernetes.Interface, tC *v1alpha1.TektonConfi
 	if pruningNamespaces.commonScheduleNs != nil && len(pruningNamespaces.commonScheduleNs) > 0 {
 		jobs := pruner.createAllJobContainers(pruningNamespaces.commonScheduleNs)
 		cronNameWithRandomSuffix := SimpleNameGenerator.RestrictLengthWithRandomSuffix(CronName)
-		if err := pruner.checkAndCreate(ctx, cronNameWithRandomSuffix, tC.Spec.TargetNamespace, tC.Spec.TargetNamespace, tC.Spec.Pruner.Schedule, jobs, ownerRef, pruneCronLabel); err != nil {
+		if err := pruner.checkAndCreate(ctx, cronNameWithRandomSuffix, tC.Spec.TargetNamespace, "", tC.Spec.Pruner.Schedule, jobs, ownerRef, pruneCronLabel, pruningNamespaces.configChanged); err != nil {
```
nikhil-thomas, line 127: the function signature has become too long. Can we capture some of these things in the `Pruner`struct.

In addition, can we break the `checkAndCreate`method into sub methods for improved redability

nikhil-thomas, line 127: `pruningNamespaces.configChanged`will this ever become `true`in this call.

how does `pruningNamespaces, err := prunableNamespaces(ctx, k, tC.Spec.Pruner)`change `configChanged`

nikhil-thomas:
/approve
✅ lgtm

pkg/reconciler/common/prune.go
pradeepitm12: I think you are right. Let me try this and update the pr.

pradeepitm12: Updated, reduced the func signature.
Also, update a bit of the function behavior.

pradeepitm12:
cc: @savitaashture @sm43

PR #565, opened 2021-12-20
Skips Namespace deletion on installlerSet deletion

sm43:
Previoulsy, if we delete an installerSet with Namespace eg pipelines
it would delete ns too which means it would delete resources from other
installerSet and the other installerSet will try to recreate it. so it
might corrupt the resources.

this update the installerSet to skip the deletion and update the owner of namespace to
the owner of installerSet which means namespace will be deleted when TektonPipeline CR
is deleted and not when its installerSet is deleted.

this also updates the deletion flow in finalizer of TektonConfig.
delete the resources in the reverse way of deletion.
so delete platform specific first then triggers and then pipelines.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/shared/tektonconfig/tektonconfig.go
@@ -50,20 +50,20 @@ var _ tektonConfigreconciler.Finalizer = (*Reconciler)(nil)
 func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.TektonConfig) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
 
+	if err := r.extension.Finalize(ctx, original); err != nil {
+		logger.Error("Failed to finalize platform resources", err)
+	}
+
 	if original.Spec.Profile == v1alpha1.ProfileLite {
 		return pipeline.TektonPipelineCRDelete(ctx, r.operatorClientSet.OperatorV1alpha1().TektonPipelines(), v1alpha1.PipelineResourceName)
 	} else {
 		// TektonPipeline and TektonTrigger is common for profile type basic and all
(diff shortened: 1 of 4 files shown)


pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -50,22 +50,22 @@ var _ tektonConfigreconciler.Finalizer = (*Reconciler)(nil)
...
+	if err := r.extension.Finalize(ctx, original); err != nil {
+		logger.Error("Failed to finalize platform resources", err)
+	}
+
 	if original.Spec.Profile == v1alpha1.ProfileLite {
 		return pipeline.TektonPipelineCRDelete(ctx, r.operatorClientSet.OperatorV1alpha1().TektonPipelines(), v1alpha1.PipelineResourceName)
 	} else {
 		// TektonPipeline and TektonTrigger is common for profile type basic and all
-		if err := pipeline.TektonPipelineCRDelete(ctx, r.operatorClientSet.OperatorV1alpha1().TektonPipelines(), v1alpha1.PipelineResourceName); err != nil {
+		if err := trigger.TektonTriggerCRDelete(ctx, r.operatorClientSet.OperatorV1alpha1().TektonTriggers(), v1alpha1.TriggerResourceName); err != nil {
```
PuneetPunamiya, line 61: Why do we need this change 🤔

sm43: we need to delete in the reverse order of resource creation
as deleting TektonPipeline will delete namespace
we delete dashboard, triggers and then pipelines 
just to make all resource get deleted in a cleaner way

pradeepitm12 (approved):
/lgtm

PR #569, opened 2021-12-22
Add field in config to disable hub devconsole integration

piyush-garg:
This will add a struct in config by named hub, where user can add flags and field
to set properties related to hub

Right now, only one field is available to enable/disable devconsole pipeline
builder and hub resources integration on OpenShift, by default the integration is
enabled but it can be disabled by adding the field with value
false in tektonconfig CR. By default the field will not be there in CR and
the integration is enabled.

The field getting added is with possible values as true and false.

if user provides invalid key and invalid value it will be handled by webhook

Also, if the field is not there in CR, webhook will not add it.

This patch also add the config CRD view role for all system authenticated users.

# Submitter Checklist

review them:

# Release Notes

```release-note
Add field in config to disable hub devconsole integration
```


pkg/apis/operator/v1alpha1/const.go
```diff
@@ -37,6 +37,9 @@ const (
 	ClusterTasksParam      = "clusterTasks"
 	PipelineTemplatesParam = "pipelineTemplates"
 
+	// Hub Params
+	DevconsoleIntegrationParam = "enable-devconsole-integration"
```
sm43, line 41: should we follow camel case similar to other param?
`EnableDevConsoleIntegration`

karthikjeeyar:
We may need a new clusterRoleBinding addon for the developer (non-admin) users to read the tekton config CRs.
Right now, only kube:admin users are able to read the tektonconfig CR, but with below CRB on the cluster, even the developer users will get access to read the config.
@vdemeester  @piyush-garg What do you think?
```
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: tekton-config-view-auth
subjects:
  - kind: Group
    apiGroup: rbac.authorization.k8s.io
    name: 'system:authenticated'
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: tektonconfigs.operator.tekton.dev-v1alpha1-view
  ```

**Comment by nikhil-thomas** [comment]
> Right now, only kube:admin users are able to read the tektonconfig CR,

@karthikjeeyar this is the intented, as CRDs like TektonConfig, TektonPipeline are internal. Non-admin users generally deal with CRDs like Task, TaskRun etc.
However, we can cosider adding such a rolebinding. Could you give us more details on how it will help a non-admin user.

**Comment by nikhil-thomas** [comment]
Ideally, the operator should be reading the new param and set a value in a configMap for the Devconsole to read. In the current scope of this patch, the operator does not do anything in response to a change in the param added by this patch. 

So we are hosting a spec field in TektonConfig for Devconsole code to read. 

👉 this is alright functionaly. However, this is a tight coupling. ie, we will not be able to change the structure or name of this part in TektonConfig in the future without breaking Devconsole.

If there is a intermediary like a configMap where the operator can set configuration for Devconsole after reading spec in TektonConfig, then we will be able to change the api/behavior on operator side without breaking devconsole.

I understand that we agreed with @karthikjeeyar that devconsole side can proceed development based on the TektonConfig spec field name as a contract.

However, we all should be aware of the tight coupling here.

cc @sm43 @vdemeester

**Comment by vdemeester** [comment]
> Ideally, the operator should be reading the new param and set a value in a configMap for the Devconsole to read. In the current scope of this patch, the operator does not do anything in response to a change in the param added by this patch.
> 
> So we are hosting a spec field in TektonConfig for Devconsole code to read.
> 
> point_right this is alright functionaly. However, this is a tight coupling. ie, we will not be able to change the structure or name of this part in TektonConfig in the future without breaking Devconsole.
> 
> If there is a intermediary like a configMap where the operator can set configuration for Devconsole after reading spec in TektonConfig, then we will be able to change the api/behavior on operator side without breaking devconsole.
> 
> I understand that we agreed with @karthikjeeyar that devconsole side can proceed development based on the TektonConfig spec field name as a contract.
> 
> However, we all should be aware of the tight coupling here.
> 
> cc @sm43 @vdemeester

@nikhil-thomas I agree and disagree. For `TektonPipeline`, … I do agree those are internal, *but* `TektonConfig` is a different beast. It's the *main* entrypoint, it's the one that the user will interact in 99% of the time they might need to interact. So the "contract" there is both with user (actual person) as well as tools (devconsole, other tooling, …). Having a configmap instead of this would make sense but it would just move the "contract" of the devconsole in a specific object (here a configmap) — I don't think I really have strong opinion about that though.

**Comment by nikhil-thomas** [comment]
I think this patch will be fine. In case we need to change the TektonConfig field in the future, then we can do it without breaking DevConsole by using proper deprecation process like we handled `serviceAccount` to `serviceAccountName` in PipelineRuns. 🧑‍💻 👍

**Comment by nikhil-thomas** [comment]
we can merge this right after the @piyush-garg pushes the minor change in defaulting that we discussed. (Removing explicit defaulting if nothing is specified in the CR, so that users on kubernetes don't see this field in their CRs.

File: pkg/apis/operator/v1alpha1/tektonconfig_defaults.go [hunk 2]
```diff
@@ -30,6 +30,8 @@ func (tc *TektonConfig) SetDefaults(ctx context.Context) {
 
 	setAddonDefaults(&tc.Spec.Addon.Params)
 
+	setHubDefaults(&tc.Spec.Hub.Params)
```
sm43, line 33: defaulting through webhook would mean this param appears on k8s platform too
we did for addon through webhook as there is a plan to enable it on k8s
but having this on k8s too might confuse the upstream users

karthikjeeyar:
> However, we can cosider adding such a rolebinding. Could you give us more details on how it will help a non-admin user.

@nikhil-thomas Most of the time the developer console users will have limited permission (eg: create/edit permissions on one or two namespaces). So if the logged in user is a devleoper (non-admin) user, then the call to fetch TektonConfig CR will fail due to lack of permission as it a cluster wide CR. With the clusterRoleBinding I suggested previously, all the logged in users should be able to read the config and UI can behave based on the configuration.

pkg/apis/operator/v1alpha1/tektonconfig_defaults.go
piyush-garg: Yes @sm43 This is what discussed in the call and decided to not do defaulting, only validation is enough.

nikhil-thomas:
> > However, we can cosider adding such a rolebinding. Could you give us more details on how it will help a non-admin user.
> 
> @nikhil-thomas Most of the time the developer console users will have limited permission (eg: create/edit permissions on one or two namespaces). So if the logged in user is a devleoper (non-admin) user, then the call to fetch TektonConfig CR will fail due to lack of permission as it a cluster wide CR. With the clusterRoleBinding I suggested previously, all the logged in users should be able to read the config and UI can behave based on the configuration.

does this mean that devconsole uses the logged in users credentials to query the api.
if yes, then giving a non-admin user read permission on the entire TektonConfig instance specto `disable-hub-integration` feature is an overkill.

I highly recomend reconsidering how we are handling this.

@piyush-garg

vdemeester:
> does this mean that devconsole uses the logged in users credentials to query the api.

I think so yes. The Devconsole is querying k8s directly for that reason too, there is no "devconsole service" running anywhere.

karthikjeeyar:
Yes, Devconsole uses Logged-in user credentials to query any resource.

pkg/apis/operator/v1alpha1/const.go
piyush-garg: fixed

pkg/apis/operator/v1alpha1/tektonconfig_defaults.go
piyush-garg: Fixed

nikhil-thomas:
✅ lgtm

piyush-garg:
@karthikjeeyar @nikhil-thomas @sm43 

Have tested the PR. It is working as expected now.

Changes
 - Webhook will not add the param in config CR by default now. cc @karthikjeeyar 
 - All system authenticated user will have access to read, list and watch

piyush-garg:
I have a question - the tekton config CR read role we want to get added for both kubernetes and openshift platform or only openshift cc @vdemeester @nikhil-thomas

vdemeester:
> I have a question - the tekton config CR read role we want to get added for both kubernetes and openshift platform or only openshift cc @vdemeester @nikhil-thomas

It's only useful for OpenShift today, but I don't see any reason for not putting it in both.

piyush-garg:
> > I have a question - the tekton config CR read role we want to get added for both kubernetes and openshift platform or only openshift cc @vdemeester @nikhil-thomas
> 
> It's only useful for OpenShift today, but I don't see any reason for not putting it in both.

Yaa this just provides read permission, I think it will be fine to provide for both. On that note, this PR is good to review and merge.

vdemeester (approved):

PR #574, opened 2022-01-06
Fix multiple installersets for tektonaddon

savitaashture:
# Changes

This PR fixes creation of multiple installer set for clustertasks addon

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/openshift/tektonaddon/extension.go
+++ b/pkg/reconciler/openshift/tektonaddon/extension.go
@@ -87,31 +87,29 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 	logger := logging.FromContext(ctx)
 	addon := comp.(*v1alpha1.TektonAddon)
 
-	exist, err := checkIfInstallerSetExist(ctx, oe.operatorClientSet, oe.version, addon, miscellaneousResourcesInstallerSet)
+	exist, err := checkIfInstallerSetExist(ctx, oe.operatorClientSet, oe.version,
+		fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, MiscellaneousResourcesInstallerSet))
 	if err != nil {
 		return err
 	}
 	if !exist {
-
 		manifest, err := getMiscellaneousManifest(ctx, addon, oe.manifest, comp)
 		if err != nil {
 			return err
 		}
 
 		if err := createInstallerSet(ctx, oe.operatorClientSet, addon, manifest, oe.version,
-			miscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
+			MiscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
 			return err
 		}
-	}
-
-	// Check if installer set is already created
-	compInstallerSet, ok := addon.Status.AddonsInstallerSet[miscellaneousResourcesInstallerSet]
-	if !ok {
 		return v1alpha1.RECONCILE_AGAIN_ERR
 	}
 
+	// Check if installer set is already created
 	installedTIS, err := oe.operatorClientSet.OperatorV1alpha1().TektonInstallerSets().
-		Get(ctx, compInstallerSet, metav1.GetOptions{})
+		List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, MiscellaneousResourcesInstallerSet),
+		})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			manifest, err := getMiscellaneousManifest(ctx, addon, oe.manifest, comp)
@@ -120,9 +118,10 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 			}
 
 			if err := createInstallerSet(ctx, oe.operatorClientSet, addon, manifest, oe.version,
-				miscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
+				MiscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
 				return err
 			}
+			return v1alpha1.RECONCILE_AGAIN_ERR
 		}
 		logger.Error("failed to get InstallerSet: %s", err)
 		return err
@@ -134,7 +133,7 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 	}
 
 	// spec hash stored on installerSet
-	lastAppliedHash := installedTIS.GetAnnotations()[tektoninstallerset.LastAppliedHashKey]
+	lastAppliedHash := installedTIS.Items[0].GetAnnotations()[tektoninstallerset.LastAppliedHashKey]
 
 	if lastAppliedHash != expectedSpecHash {
 
@@ -144,33 +143,31 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 		}
 
 		// Update the spec hash
-		current := installedTIS.GetAnnotations()
+		current := installedTIS.Items[0].GetAnnotations()
 		current[tektoninstallerset.LastAppliedHashKey] = expectedSpecHash
-		installedTIS.SetAnnotations(current)
+		installedTIS.Items[0].SetAnnotations(current)
 
 		// Update the manifests
-		installedTIS.Spec.Manifests = manifest.Resources()
+		installedTIS.Items[0].Spec.Manifests = manifest.Resources()
 
 		if _, err = oe.operatorClientSet.OperatorV1alpha1().TektonInstallerSets().
-			Update(ctx, installedTIS, metav1.UpdateOptions{}); err != nil {
+			Update(ctx, &installedTIS.Items[0], metav1.UpdateOptions{}); err != nil {
 			return err
 		}
 
 		return v1alpha1.RECONCILE_AGAIN_ERR
 	}
 
-	existingInstallerSet, ok := addon.Status.AddonsInstallerSet[miscellaneousResourcesInstallerSet]
-	if !ok {
-		return v1alpha1.RECONCILE_AGAIN_ERR
-	}
 	installedAddonIS, err := oe.operatorClientSet.OperatorV1alpha1().TektonInstallerSets().
-		Get(ctx, existingInstallerSet, metav1.GetOptions{})
+		List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, MiscellaneousResourcesInstallerSet),
+		})
 	if err != nil {
 		logger.Error("failed to get InstallerSet: %s", err)
 		return err
 	}
 
-	ready := installedAddonIS.Status.GetCondition(apis.ConditionReady)
+	ready := installedAddonIS.Items[0].Status.GetCondition(apis.ConditionReady)
 	if ready == nil {
 		return v1alpha1.RECONCILE_AGAIN_ERR
 	}
@@ -180,7 +177,8 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 	}
 
 	consolecliManifest := oe.manifest
-	exist, err = checkIfInstallerSetExist(ctx, oe.operatorClientSet, oe.version, addon, consoleCLIInstallerSet)
+	exist, err = checkIfInstallerSetExist(ctx, oe.operatorClientSet, oe.version,
+		fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, ConsoleCLIInstallerSet))
 	if err != nil {
 		return err
 	}
@@ -203,7 +201,7 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 		}
 
 		if err := createInstallerSet(ctx, oe.operatorClientSet, addon, consolecliManifest, oe.version,
-			consoleCLIInstallerSet, "addon-consolecli"); err != nil {
+			ConsoleCLIInstallerSet, "addon-consolecli"); err != nil {
 			return err
 		}
 	}
--- a/pkg/reconciler/openshift/tektonaddon/tektonaddon.go
+++ b/pkg/reconciler/openshift/tektonaddon/tektonaddon.go
@@ -65,16 +65,6 @@ const (
 	providerTypeRedHat    = "redhat"
 )
 
-const (
-	clusterTaskInstallerSet            = "ClusterTaskInstallerSet"
-	pipelinesTemplateInstallerSet      = "PipelinesTemplateInstallerSet"
-	triggersResourcesInstallerSet      = "TriggersResourcesInstallerSet"
-	consoleCLIInstallerSet             = "ConsoleCLIInstallerSet"
-	miscellaneousResourcesInstallerSet = "MiscellaneousResourcesInstallerSet"
-
-	createdByValue = "TektonAddon"
-)
-
 // Check that our Reconciler implements controller.Reconciler
 var _ tektonaddonreconciler.Interface = (*Reconciler)(nil)
 var _ tektonaddonreconciler.Finalizer = (*Reconciler)(nil)
@@ -180,7 +170,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, ta *v1alpha1.TektonAddon
 	// with their manifest
 	if ctVal == "true" {
 
-		exist, err := checkIfInstallerSetExist(ctx, r.operatorClientSet, r.version, ta, clusterTaskInstallerSet)
+		exist, err := checkIfInstallerSetExist(ctx, r.operatorClientSet, r.version,
+			fmt.Sprintf("%s=%s", tektoninstallerset.InstallerSetType, ClusterTaskInstallerSet))
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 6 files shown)


pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -416,41 +420,27 @@ func createInstallerSet(ctx context.Context, oc clientset.Interface, ta *v1alpha
...
+func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, releaseVersion, component, specHash string) *v1alpha1.TektonInstallerSet {
 	ownerRef := *metav1.NewControllerRef(ta, ta.GetGroupVersionKind())
 	return &v1alpha1.TektonInstallerSet{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", prefix),
 			Labels: map[string]string{
-				tektoninstallerset.CreatedByKey: createdByValue,
+				tektoninstallerset.CreatedByKey:      createdByValue,
+				tektoninstallerset.InstallerSetType:  component,
+				tektoninstallerset.ReleaseVersionKey: releaseVersion,
```
pradeepitm12, line 441: :+1:

vdemeester (commented):
@savitaashture can we have some tests ? (not sure how easy it would be but still)

savitaashture:
> @savitaashture can we have some tests ? (not sure how easy it would be but still)

Updated e2e test to verify installerSets for addon based on label

pkg/reconciler/kubernetes/tektoninstallerset/const.go
```diff
@@ -21,4 +21,12 @@ const (
...
 	ReleaseVersionKey  = "operator.tekton.dev/release-version"
 	TargetNamespaceKey = "operator.tekton.dev/target-namespace"
+	InstallerSetType   = "operator.tekton.dev/type"
+
+	ClusterTaskInstallerSet            = "ClusterTask"
+	PipelinesTemplateInstallerSet      = "PipelinesTemplate"
+	TriggersResourcesInstallerSet      = "TriggersResources"
+	ConsoleCLIInstallerSet             = "ConsoleCLI"
+	MiscellaneousResourcesInstallerSet = "MiscellaneousResources"
+	CreatedByValue                     = "TektonAddon"
```
sm43, line 31: these are addon specific const right?
should we keep them in addon pkg? and not in tektoninstallerset?

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -369,26 +354,34 @@ func (r *Reconciler) ensureClusterTasks(ctx context.Context, ta *v1alpha1.Tekton
...
+		}
+		return false, err
 	}
 
-	if compInstallerSet != "" {
-		// if already created then check which version it is
-		ctIs, err := oc.OperatorV1alpha1().TektonInstallerSets().
-			Get(ctx, compInstallerSet, metav1.GetOptions{})
+	// Delete installer sets if exist more than one
+	if len(installerSets.Items) > 1 {
```
sm43, line 371: here if there are more than 1 installer set we are deleting all right?
then do we need to continue after deleting?

as after deleting
we are looping on items
```
for i := range installerSets.Items {
```
but we have already deleted all

sm43: we need a return statement I think after deleting

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -369,26 +354,34 @@ func (r *Reconciler) ensureClusterTasks(ctx context.Context, ta *v1alpha1.Tekton
 // and if installer set which already exist is of older version then it deletes and return false to create a new
 // installer set
 func checkIfInstallerSetExist(ctx context.Context, oc clientset.Interface, relVersion string,
```
sm43, line 356: as this is more generalised now, can we move this to installer Set pkg
so that other component can use the same?
something like `installerSet.GetByLabel()`

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -369,26 +354,34 @@ func (r *Reconciler) ensureClusterTasks(ctx context.Context, ta *v1alpha1.Tekton
...
-	ta *v1alpha1.TektonAddon, component string) (bool, error) {
+	component string) (bool, error) {
```
nikhil-thomas, line 357: the instead of sending the component, version etc separately the last paramerte should be `labelSelector` so that whoever using this function could format `1` or `n` labels and pass it to this function. 

Otherwise, we will have to change the the implementation of this function for each use case for other components in the future.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -416,41 +410,27 @@ func createInstallerSet(ctx context.Context, oc clientset.Interface, ta *v1alpha
...
 
-func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, releaseVersion, specHash string) *v1alpha1.TektonInstallerSet {
+func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, releaseVersion, component, specHash string) *v1alpha1.TektonInstallerSet {
 	ownerRef := *metav1.NewControllerRef(ta, ta.GetGroupVersionKind())
 	return &v1alpha1.TektonInstallerSet{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", prefix),
 			Labels: map[string]string{
-				tektoninstallerset.CreatedByKey: createdByValue,
+				tektoninstallerset.CreatedByKey:      tektoninstallerset.CreatedByValue,
```
nikhil-thomas, line 429: `tektoninstallerset.CreatedByValue` this definitely a reconciler specific item, which has to live in the reconciler which is using this (for example pipelines, addons, triggers..)(

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -416,41 +410,27 @@ func createInstallerSet(ctx context.Context, oc clientset.Interface, ta *v1alpha
...
+	if _, err := oc.OperatorV1alpha1().TektonInstallerSets().
+		Create(ctx, is, metav1.CreateOptions{}); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, releaseVersion, specHash string) *v1alpha1.TektonInstallerSet {
+func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, releaseVersion, component, specHash string) *v1alpha1.TektonInstallerSet {
```
nikhil-thomas, line 423: here also let us pass the set of labes as a single argument (label selector itself or map). 
so that formatiing the labels will be the responsibility of the caller.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -462,28 +442,14 @@ func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, re
 	}
 }
 
-func (r *Reconciler) deleteInstallerSet(ctx context.Context, ta *v1alpha1.TektonAddon, component string) error {
+func (r *Reconciler) deleteInstallerSet(ctx context.Context, component string) error {
```
nikhil-thomas, line 446: same here, instead of the `component string` let us pass label selector as the last param

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -462,28 +442,14 @@ func makeInstallerSet(ta *v1alpha1.TektonAddon, manifest mf.Manifest, prefix, re
...
-
-		// clear the name of installer set from TektonAddon status
-		delete(ta.Status.AddonsInstallerSet, component)
-		_, err = r.operatorClientSet.OperatorV1alpha1().TektonAddons().
-			UpdateStatus(ctx, ta, metav1.UpdateOptions{})
-		if err != nil && !errors.IsNotFound(err) {
-			return err
-		}
+	err := r.operatorClientSet.OperatorV1alpha1().TektonInstallerSets().
+		DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{
```
nikhil-thomas, line 448: 👍

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
nikhil-thomas, line 371: why is this `>` 1 we should delete all, shouldn't we.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -369,26 +354,34 @@ func (r *Reconciler) ensureClusterTasks(ctx context.Context, ta *v1alpha1.Tekton
...
-			if errors.IsNotFound(err) {
-				return false, nil
-			}
 			return false, err
 		}
+	}
 
-		version, ok := ctIs.Annotations[tektoninstallerset.ReleaseVersionKey]
+	// Check if installer set is already created
+	for i := range installerSets.Items {
```
nikhil-thomas, line 382: what is the purpose of this `for loop`. 
the above `DeleteCollection()` call will send a `delete `  api call so why do we try to process the same set with this `for loop`.

All the items in this  instalerset.Items is being deleted because of the above call, isn't it.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
nikhil-thomas: in other words, we as @sm43 said we should return if we are deleting all. 

I think this `> 1 ` check is misleading. We should stick to a single mechanism irrespective of the number of instalerSets returned (delete 1 or many).

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
nikhil-thomas: the DeleteCollection call only makes an Api call, it won't make `installerSets.Items` empty.

savitaashture:
@sm43 @nikhil-thomas I have addressed review comments except refactoring part 
Planning to do refactor  for `checkIfInstallerSetExist` `createInstallerSet` and  `makeInstallerSet` common for all 
as a separate PR once @pradeepitm12 changes are also in for this multiple installerset fix for Pipeline and Triggers

WDYT?

pkg/reconciler/openshift/tektonaddon/extension.go
```diff
@@ -134,7 +131,7 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 	}
 
 	// spec hash stored on installerSet
-	lastAppliedHash := installedTIS.GetAnnotations()[tektoninstallerset.LastAppliedHashKey]
+	lastAppliedHash := installedTIS.Items[0].GetAnnotations()[tektoninstallerset.LastAppliedHashKey]
```
sm43, line 134: should we have a check for `len` for Items ? just to be safe before accessing

pkg/reconciler/openshift/tektonaddon/extension.go
```diff
@@ -120,21 +117,25 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
...
 			}
 		}
 		logger.Error("failed to get InstallerSet: %s", err)
 		return err
 	}
 
+	if len(installedTIS.Items) == 0 {
+		return v1alpha1.RECONCILE_AGAIN_ERR
+	}
+
```
savitaashture, line 131: @sm43 if len is 0 shall we return Reconcile again and wait for sometime or shall we return error itself

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
nikhil-thomas: what i meant is `tektoninstallerset.CreatedByValue` sounds odd, i was expecting just `CreatedByValue` (as CreatedByValue is defined in addons. 

In other words, the value part of this annotation comes from the entity who creates the installerset. So referencing it as `tektoninstallerset.` sounds odd.

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -369,40 +358,36 @@ func (r *Reconciler) ensureClusterTasks(ctx context.Context, ta *v1alpha1.Tekton
...
-		// release version doesn't exist or is different from expected
-		// deleted existing InstallerSet and create a new one
-
-		err = oc.OperatorV1alpha1().TektonInstallerSets().
-			Delete(ctx, compInstallerSet, metav1.DeleteOptions{})
-		if err != nil {
-			return false, err
-		}
+	// release version doesn't exist or is different from expected
+	// deleted existing InstallerSet and create a new one
```
nikhil-thomas, line 385:
```suggestion
	// release version doesn't exist or is different from expected
	// deleted existing InstallerSet and create a new one
	// or there is more than one installerset (unexpected)
```

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
```diff
@@ -369,40 +358,36 @@ func (r *Reconciler) ensureClusterTasks(ctx context.Context, ta *v1alpha1.Tekton
...
+	if err = oc.OperatorV1alpha1().TektonInstallerSets().
+		DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{
+			LabelSelector: labelSelector,
+		}); err != nil {
+		return false, err
```
nikhil-thomas, line 390: this should work but ideally, we should loop through the list and call individual delete() after ensuring `DeletionTimestamp` is not set in metadata of each installerset. 

 We could be here even after a previous reconcile loop has called `DeleteCollection` as the installerset might take few seconds to clean up (finalize) what it has created.

nikhil-thomas: ie, the DeletionCollection() is just an API call, so we cannot gurantee when the installerSet will be completely removed.

Mentioning this here, to make sure that we are all on the same page.
cc @sm43

pkg/reconciler/openshift/tektonaddon/extension.go
```diff
@@ -120,21 +117,25 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 			}
 
 			if err := createInstallerSet(ctx, oe.operatorClientSet, addon, manifest, oe.version,
-				miscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
+				MiscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
```
nikhil-thomas, line 120: retrun `RCONCILE_AGAIN_ERR` here to ensure that this API call has created the installerSet.

pkg/reconciler/openshift/tektonaddon/extension.go
```diff
@@ -120,21 +117,25 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
...
 				return err
 			}
 		}
 		logger.Error("failed to get InstallerSet: %s", err)
 		return err
 	}
 
+	if len(installedTIS.Items) == 0 {
+		return v1alpha1.RECONCILE_AGAIN_ERR
+	}
```
nikhil-thomas, line 130: as creating a new installerset will not have any effect on the length of current installedTIS.items, let us move this return to https://github.com/tektoncd/operator/pull/574/files#r782647382

pkg/reconciler/openshift/tektonaddon/extension.go
nikhil-thomas: https://github.com/tektoncd/operator/pull/574/files#r782647681

pkg/reconciler/openshift/tektonaddon/tektonaddon.go
nikhil-thomas: separate pr sound fine. 
could you create and issue (or jira) to track this.

nikhil-thomas:
@savitaashture could you make 2 changes
1. https://github.com/tektoncd/operator/pull/574/files#r782647681
2. (minor comment which will help future readers 🧑‍💻 ) https://github.com/tektoncd/operator/pull/574#discussion_r782641306

savitaashture:
> @savitaashture could you make 2 changes
> 
> 1. https://github.com/tektoncd/operator/pull/574/files#r782647681
> 2. (minor comment which will help future readers technologist ) [Fix multiple installersets for tektonaddon #574 (comment)](https://github.com/tektoncd/operator/pull/574#discussion_r782641306)

@nikhil-thomas Addressed :+1:

pkg/reconciler/openshift/tektonaddon/extension.go
```diff
@@ -87,31 +87,28 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
...
 		manifest, err := getMiscellaneousManifest(ctx, addon, oe.manifest, comp)
 		if err != nil {
 			return err
 		}
 
 		if err := createInstallerSet(ctx, oe.operatorClientSet, addon, manifest, oe.version,
-			miscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
-			return err
+			MiscellaneousResourcesInstallerSet, "addon-openshift"); err != nil {
+			return v1alpha1.RECONCILE_AGAIN_ERR
```
nikhil-thomas, line 103: if err is nil then return `v1alpha1.RECONCILE_AGAIN_ERR` 
else return err.. 

otherwise we will never catch the error

pkg/reconciler/openshift/tektonaddon/extension.go
nikhil-thomas: adding a explicit wait in situations like these is a kubernetes anti-pattern.

if the cluster state is not " yet " the desired state, we exit out of the reconcile loop (return RECONCILE_AGAIN) and let the future loops proceed past the check.

pkg/reconciler/openshift/tektonaddon/extension.go
savitaashture: :woman_facepalming:  :see_no_evil:

nikhil-thomas:
✅ lgtm

sm43:
LGTM

PR #588, opened 2022-01-25
Fix duplicate InstallerSets in Pipeline Reconciler

nikhil-thomas:
Fix occurrence of duplicate installerSets in pre, post and main
recociler logic of TektonPipeline reconciler

Adds a label based querying to identify InstallerSets instead of
querying by there name

Use Operator version instead of Component versions while deciding whether to recreate deployments and other resources during upgrades

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
fixes duplicate pipeline (pre, post, main)  installerSets occuring in in TektoncdPipeline reconciler. 
```

### DIFF
--- a/pkg/reconciler/kubernetes/initcontroller/initcontroller.go
+++ b/pkg/reconciler/kubernetes/initcontroller/initcontroller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"go.uber.org/zap"
 	"knative.dev/pkg/injection"
+	"knative.dev/pkg/logging"
 )
 
 type Controller struct {
@@ -39,6 +40,16 @@ type Controller struct {
 
 type PayloadOptions struct {
 	ReadOnly bool
+}
+
+func OperatorVersion(ctx context.Context) (string, error) {
+	logger := logging.FromContext(ctx)
+	operatorVersion, ok := os.LookupEnv(v1alpha1.VersionEnvKey)
+	if !ok || operatorVersion == "" {
+		logger.Errorf(v1alpha1.VERSION_ENV_NOT_SET_ERR.Error())
+		return "", v1alpha1.VERSION_ENV_NOT_SET_ERR
+	}
+	return operatorVersion, nil
 }
 
 func (ctrl Controller) InitController(ctx context.Context, opts PayloadOptions) (mf.Manifest, string) {
--- a/pkg/reconciler/kubernetes/tektoninstallerset/query.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/query.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tektoninstallerset
+
+import (
+	"context"
+
+	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	clientset "github.com/tektoncd/operator/pkg/client/clientset/versioned"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CurrentInstallerSetName(ctx context.Context, client clientset.Interface, labelSelector string) (string, error) {
+	iSets, err := client.OperatorV1alpha1().TektonInstallerSets().List(ctx, v1.ListOptions{
+		LabelSelector: labelSelector,
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(iSets.Items) == 0 {
+		return "", nil
+	}
+	if len(iSets.Items) == 1 {
+		iSetName := iSets.Items[0].GetName()
+		return iSetName, nil
+	}
+
+	// len(iSets.Items) > 1
+	// delete all installerSets as it cannot be decided which one is the desired one
+	err = client.OperatorV1alpha1().TektonInstallerSets().DeleteCollection(ctx,
+		v1.DeleteOptions{},
+		v1.ListOptions{
+			LabelSelector: labelSelector,
+		})
+	if err != nil {
+		return "", err
+	}
+	return "", v1alpha1.RECONCILE_AGAIN_ERR
+}
--- a/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
+++ b/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
@@ -61,11 +61,11 @@ type Reconciler struct {
 	extension common.Extension
 	// enqueueAfter enqueues a obj after a duration
 	enqueueAfter func(obj interface{}, after time.Duration)
-	// releaseVersion describes the current pipelines version
-	releaseVersion string
 	// metrics handles metrics for pipeline install
-	metrics       *Recorder
-	kubeClientSet kubernetes.Interface
+	metrics         *Recorder
+	kubeClientSet   kubernetes.Interface
+	operatorVersion string
+	pipelineVersion string
 }
 
 // Check that our Reconciler implements controller.Reconciler
@@ -141,7 +141,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
 			return err
 		}
 		// If there was no existing installer set, that means its a new install
-		r.metrics.logMetrics(metricsNew, r.releaseVersion, logger)
+		r.metrics.logMetrics(metricsNew, r.pipelineVersion, logger)
 
 		return r.updateTektonPipelineStatus(ctx, tp, createdIs)
 	}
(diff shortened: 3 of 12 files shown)


pkg/reconciler/openshift/tektonpipeline/installerset.go
```diff
@@ -35,58 +36,72 @@ const createdByValue = "TektonPipeline"
...
+		return false, v1alpha1.RECONCILE_AGAIN_ERR
+	}
 
-		// Check release version and target namespace
-		// If anyone of this is not as expected then delete existing
-		// installer set and create a new one
+	// if the InstallerSet already exists then check then validate it:
+	// Check release version and target namespace
+	// If anyone of this is not as expected then delete existing
+	// InstallerSet and return false
```
savitaashture, line 75: @nikhil-thomas 
can we do 
if len(installerSets.Items) == 1 {
// check version n namespace match
// If matched return true
}

// Delete installer sets like below
 oc.OperatorV1alpha1().TektonInstallerSets().
		DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{
			LabelSelector: labelSelector,
		})
// which covers when len is 0, >1 and version/ns mismatch scenario

WDYT?

pkg/reconciler/openshift/tektonpipeline/installerset.go
```diff
@@ -130,27 +147,23 @@ func makeInstallerSet(tp *v1alpha1.TektonPipeline, manifest mf.Manifest, prefix,
...
-		delete(ta.Status.ExtentionInstallerSets, component)
-		_, err = oc.OperatorV1alpha1().TektonPipelines().UpdateStatus(ctx, ta, metav1.UpdateOptions{})
-		if err != nil && !errors.IsNotFound(err) {
-			return err
-		}
+	// clear the name of installer set from TektonPipeline status
+	delete(ta.Status.ExtentionInstallerSets, component)
+	_, err = oc.OperatorV1alpha1().TektonPipelines().UpdateStatus(ctx, ta, metav1.UpdateOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
```
savitaashture, line 166: As we are not storing installerset name in status 
do we still required to do this :thinking:

pkg/reconciler/openshift/tektonpipeline/installerset.go
```diff
@@ -130,27 +147,23 @@ func makeInstallerSet(tp *v1alpha1.TektonPipeline, manifest mf.Manifest, prefix,
...
-		return nil
+	// delete the installer set
+	err := oc.OperatorV1alpha1().TektonInstallerSets().
+		DeleteCollection(ctx, metav1.DeleteOptions{},
+			metav1.ListOptions{
+				LabelSelector: labelSelector,
+			})
+	if err != nil {
+		return err
 	}
```
savitaashture, line 160:
```suggestion
	if err := oc.OperatorV1alpha1().TektonInstallerSets().
		DeleteCollection(ctx, metav1.DeleteOptions{},
			metav1.ListOptions{
				LabelSelector: labelSelector,
			}); err != nil {
		return err
	}
```

pkg/reconciler/openshift/tektonpipeline/extension.go
```diff
@@ -168,32 +179,27 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
...
 func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonComponent) error {
 	pipeline := comp.(*v1alpha1.TektonPipeline)
-
-	installerSets := pipeline.Status.ExtentionInstallerSets
-	if len(installerSets) == 0 {
-		return nil
+	err := deleteInstallerSet(ctx, oe.operatorClientSet, pipeline, postPipelineInstallerSet, postReconcileSelector)
+	if err != nil {
+		return err
 	}
```
savitaashture, line 198: I think we need `prePipelineInstallerSet` & `preReconcileSelector` here as post thing are handled below

```suggestion
	if err := deleteInstallerSet(ctx, oe.operatorClientSet, pipeline, prePipelineInstallerSet, preReconcileSelector); err != nil {
		return err
	}
```

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -316,16 +319,18 @@ func (r *Reconciler) createInstallerSet(ctx context.Context, tp *v1alpha1.Tekton
...
 			GenerateName: fmt.Sprintf("%s-", v1alpha1.PipelineResourceName),
 			Labels: map[string]string{
-				tektoninstallerset.CreatedByKey: createdByValue,
+				tektoninstallerset.CreatedByKey:      createdByValue,
+				tektoninstallerset.InstallerSetType:  v1alpha1.PipelineResourceName,
+				tektoninstallerset.ReleaseVersionKey: r.releaseVersion,
 			},
 			Annotations: map[string]string{
-				tektoninstallerset.ReleaseVersionKey:  releaseVersion,
+				tektoninstallerset.ReleaseVersionKey:  r.releaseVersion,
```
savitaashture, line 333: I think we can skip release version from annotation as its added to label

pkg/reconciler/openshift/tektonpipeline/installerset.go
nikhil-thomas: i will capture this in an issue and remove this part properly as this will involve change to Status struct.
ie, let us make that change in a minor release instead of a patch release.

That way we can properly deprecate that field.

issue: https://github.com/tektoncd/operator/issues/590

pkg/reconciler/openshift/tektonpipeline/installerset.go
nikhil-thomas: We could do that if the response of `>1` and `==0`were same.
but, here are double checking that the installer set is completely deleted even after no error from DeleteCollection API call.

if we don't do this then deleteCollection call will add DeleteTimeStamps and infact resources could linger around when the next new InstallerSet of same type comes up.

As mentioned in the `// only scenario where this function returns // false, nil` comment. 
this function returns `false, nil` (so that the caller could continue) only if there are 0 installerSets on cluster.

In `>1` scenario,  DeleteCollection() with no error, the function returns `Reconcile Again error` so that the the function can ensure in the next loop that the state is indeed no: of installerSets == 0.

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -131,14 +133,21 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
 	tp.Status.MarkPreReconcilerComplete()
 
 	// Check if an tekton installer set already exists, if not then create
-	existingInstallerSet := tp.Status.GetTektonInstallerSet()
+	labelSelector := fmt.Sprintf("%s=%s,%s=%s",
```
sm43, line 136: jfyi k8s pkg "k8s.io/apimachinery/pkg/labels"
provides a func to build labels for `ListOptions`
so something like below
```
// created by requirement
	createdByReq, err := labels.NewRequirement("CreatedByKey", selection.Exists, []string{"TektonPipeline"})
	if err != nil {
		panic(err)
	}
	labelSelector := labels.NewSelector().Add(*createdByReq)
// add another version requirement
	labelSelector = labelSelector.Add(versionReq)
	options.LabelSelector = labelSelector.String()
```

pkg/reconciler/kubernetes/tektoninstallerset/query.go
```diff
@@ -0,0 +1,37 @@
+package tektoninstallerset
```
sm43, line 1: nit: license

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -131,14 +133,21 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
...
 	if existingInstallerSet == "" {
 		createdIs, err := r.createInstallerSet(ctx, tp)
 		if err != nil {
 			return err
 		}
 		// If there was no existing installer set, that means its a new install
-		r.metrics.logMetrics(metricsNew, r.releaseVersion, logger)
+		r.metrics.logMetrics(metricsNew, r.pipelineVersion, logger)
 
 		return r.updateTektonPipelineStatus(ctx, tp, createdIs)
```
sm43, line 152: in `updateTektonPipelineStatus`
we should now skip the `UpdateStatus` api n call 
As it is not necessary to explicitly update now
and let knative/pkg do that at then end of reconciler

sm43: we could just move this 
```
	tp.Status.SetTektonInstallerSet(createdIs.Name)
	tp.Status.SetVersion(r.pipelineVersion)
```
to the end of main reconciler now

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
nikhil-thomas: yup, this is way better. Thanks 🧑‍💻 👍 
the `sprintf` labels are at multiple places. so i shall add an issue so that all of them can be handled together later.

https://github.com/tektoncd/operator/issues/593

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
nikhil-thomas: we could, but in the current flow, what follows is comparisson of version and targetnamespace, which will be unnecessary right after a new installSet creation.

one goal of this patch was to fix the bug with minimal changes in lifecycle (install/uninstall or metrics). 

So shall we handle all refactoring like this in a different patch.

pradeepitm12:
most of the review comments already addressed :+1: 
/lgtm

savitaashture (approved):
Thank you :+1:

pkg/reconciler/kubernetes/initcontroller/initcontroller.go
```diff
@@ -41,6 +42,16 @@ type PayloadOptions struct {
 	ReadOnly bool
 }
 
+func OperatorVersion(ctx context.Context) (string, error) {
```
PuneetPunamiya, line 45: Can't we read this from the config map ?

sm43:
Thank you ! 
/meow

pkg/reconciler/kubernetes/initcontroller/initcontroller.go
sm43: yes but the version is already in the env of pod so easier that way
why to do an API call.. :)
also configmap is for other external tools to know the version of operator

PR #592, opened 2022-01-27
Fix upgrade deadlock in Pipeline and Triggers

nikhil-thomas:
Add a mechanism to disable webhook validation if the configMaps
are not already there in the cluster.

**What this patch does:** when webhook endpoint slice is [], the "rules" section of `config.webhook.**` `validatingwebhookconfiguration` is remove. So that the updates on configMaps eg: `config-defaults` are not deadlocked.

The dead lock is caused due to a cluster level cyclic dependency in
tektoncd/pipelines and in tektoncd/triggers. More details will be captured in
issues in the both repository.

Note: infact this deadlock can occur in any tektoncd project which follows tektoncd/pipeline
file list structure in `config/`.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/common/deadlockbreaker.go
+++ b/pkg/reconciler/common/deadlockbreaker.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+
+	"github.com/manifestival/manifestival"
+	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+)
+
+var webhookNames = map[string]string{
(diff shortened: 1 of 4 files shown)


pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -250,8 +250,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tp *v1alpha1.TektonPipel
 		return nil
 	} else if ready.Status == corev1.ConditionFalse {
```
sm43, line 251: there is a short period when deployments are applied and pods are not yet running 
installerset sets ready status as false 
which TP reads that and sets its own status as false too
in that case would this logic consider that as `PreemptDeadlock` and delete resources?

sm43: https://github.com/tektoncd/operator/blob/eb53caf21f0eb83f7820ccd42accfa29de5caf1e/pkg/reconciler/kubernetes/tektoninstallerset/install.go#L334
this has to be improved ..
where we need to differentiated false and waiting state

nikhil-thomas: `PreemptDeadlock` will only remove "rules" from Webhook, ie, set it to its initial state. As the webhookDefinition itself is not removed when the Deployment finally comes up, the rules will get populated and future edit operations on configMaps will be validated.

In addition, the `PreemptDeadlock` acts only if 1 or more configMaps are missing. So in majority of the time, if configMaps are there `PreemptDeadlock` does nothing. 

the deleteDeploymentCall in the `PreemptDeadlock` is a simple and effective way to trigger a reconcile loop in Installlerset making in recreate the configMaps.

nikhil-thomas: changed the mechanism by moving the focus to "Endpoint" resource rather than configMaps.

pkg/reconciler/common/deadlockbreaker.go
```diff
@@ -0,0 +1,76 @@
+package common
```
sm43, line 1: nit: license

vdemeester (approved):
SGTM 👍🏼

pradeepitm12 (approved):
/lgtm

PR #602, opened 2022-01-28
[cherry-pick] Fix versioned clustertask getting deleted during upgrade

piyush-garg:
This is cherry-pick PR of #599 

This will fix the issue of versioned clustertask getting
deleted during upgrade
Now two installersets will be created for versioned and
non version clustertask. There is new label getting added
on versioned clustertask installerset having major and minor version
details and the versioned clustertask installetset will only be deleted
when there will be two of same minor and major version and installerset
for latest operator version will be created.
Eventually installerset will remain there of different operator minor version
only ending in versioned clustertask to stay on cluster

```release-note
Fix versioned clustertask getting deleted during upgrade
```

### DIFF
--- a/pkg/reconciler/kubernetes/tektoninstallerset/const.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/const.go
@@ -17,9 +17,10 @@ limitations under the License.
 package tektoninstallerset
 
 const (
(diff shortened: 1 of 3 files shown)


savitaashture (commented):
/lgtm

PR #620, opened 2022-02-09
Improve TekktonInstallerSet Reconciler

nikhil-thomas:
Add explicit check to ensure that resources are present on cluster
after an manifestival.Apply() call

Add DeletePropagation policy 'Foreground' to TektonInstallerSet
finalizer logic

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/kubernetes/tektoninstallerset/install.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/install.go
@@ -71,15 +71,44 @@ type installer struct {
 	Manifest mf.Manifest
 }
 
-func (i *installer) EnsureCRDs() error {
-	if err := i.Manifest.Filter(mf.Any(mf.CRDs)).Apply(); err != nil {
+func ensureResources(mani *mf.Manifest) error {
+	freshCreate := true
+	// check if all resources in a given set of resources exists
+	ok, err := allResourcesExists(mani)
+	if err != nil {
+		if !apierrs.IsNotFound(err) {
+			return err
+		}
+	}
+	// if error == NotFound error or !ok
+	// then Apply all resources
+	// if ok, that means the reosource already exists
+	// but this reconcile could be an modification eg: concig-defaults change
+	// set freshCreate flag to false, and then Apply all
+	// so that we can skip (break) RECONCILE_AGAIN loop path
+	if ok {
+		freshCreate = false
+	}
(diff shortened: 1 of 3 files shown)


pkg/reconciler/kubernetes/tektoninstallerset/install.go
```diff
@@ -71,15 +71,36 @@ type installer struct {
 	Manifest mf.Manifest
 }
 
-func (i *installer) EnsureCRDs() error {
-	if err := i.Manifest.Filter(mf.Any(mf.CRDs)).Apply(); err != nil {
+func ensureResources(mani *mf.Manifest) error {
```
sm43, line 74: will this allow to update resources if it is changed?
for eg. when user changes some fields in TektonConfig for config default/feature flag
we update the configmap

nikhil-thomas: no, that is tricky.
either we can make all updates immutable.
or move the allResourcesExists Check to after the Apply() call. this second option is not perfect but it could be good enough

nikhil-thomas: updated the logic 🧑‍💻

pkg/reconciler/kubernetes/tektoninstallerset/install.go
```diff
@@ -71,15 +71,44 @@ type installer struct {
...
+	// then Apply all resources
+	// if ok, that means the reosource already exists
+	// but this reconcile could be an modification eg: concig-defaults change
+	// set freshCreate flag to false, and then Apply all
+	// so that we can skip (break) RECONCILE_AGAIN loop path
+	if ok {
+		freshCreate = false
+	}
+
+	if err := mani.Apply(); err != nil {
```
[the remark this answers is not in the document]
sm43: > we should check any one resource
the first resource in the list (except ns, crd)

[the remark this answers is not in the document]
nikhil-thomas: it is alright, at max the calls will be issued only once.
ideally, the operaotr should check (issue get calls) for each resource in each reconcile loop.

vdemeester (approved):

PR #628, opened 2022-02-20
Refactors the mechanism for creating and updating TektonPipeline CR and TektonTrigger CR

PuneetPunamiya:
# Changes

## Refactoring the TektonPipeline CR
  
- This patch now checks first if there is TektonPipeline CR
    already present, if yes then it checks if there are any
    changes in the instance, if yes then it makes the changes
    and updates the CR, if not then it creates a new CR

 - This patch also removes the polling state mechanism for
    tekton Pipeline CR

---

## Refactoring the TektonTrigger CR

  - This patch now checks first if there is TektonTrigger CR
         already present, if yes then it checks if there are any
         changes in the instance, if yes then it makes the changes
         and updates the CR, if not then it creates a new CR
    
  - This patch also removes the polling state mechanism for
     tekton Trigger CR

---

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
+++ b/pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
@@ -20,83 +20,44 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
 	"reflect"
 
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
 
 	op "github.com/tektoncd/operator/pkg/client/clientset/versioned/typed/operator/v1alpha1"
-	operatorv1alpha1 "github.com/tektoncd/operator/pkg/client/clientset/versioned/typed/operator/v1alpha1"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"knative.dev/pkg/test/logging"
 )
 
-func CreatePipelineCR(ctx context.Context, instance v1alpha1.TektonComponent, client operatorv1alpha1.OperatorV1alpha1Interface) error {
-	configInstance := instance.(*v1alpha1.TektonConfig)
-	if _, err := ensureTektonPipelineExists(ctx, client.TektonPipelines(), configInstance); err != nil {
-		return errors.New(err.Error())
-	}
-	if _, err := waitForTektonPipelineState(ctx, client.TektonPipelines(), v1alpha1.PipelineResourceName,
-		isTektonPipelineReady); err != nil {
-		log.Println("TektonPipeline is not in ready state: ", err)
-		return err
-	}
-	return nil
-}
-
-func ensureTektonPipelineExists(ctx context.Context, clients op.TektonPipelineInterface, config *v1alpha1.TektonConfig) (*v1alpha1.TektonPipeline, error) {
+func EnsureTektonPipelineExists(ctx context.Context, clients op.TektonPipelineInterface, config *v1alpha1.TektonConfig) (*v1alpha1.TektonPipeline, error) {
 	tpCR, err := GetPipeline(ctx, clients, v1alpha1.PipelineResourceName)
-	if err == nil {
-		// if the pipeline spec is changed then update the instance
-		updated := false
 
-		if config.Spec.TargetNamespace != tpCR.Spec.TargetNamespace {
-			tpCR.Spec.TargetNamespace = config.Spec.TargetNamespace
-			updated = true
+	if err != nil {
+		if !apierrs.IsNotFound(err) {
+			return nil, err
 		}
-
-		if !reflect.DeepEqual(tpCR.Spec.Pipeline, config.Spec.Pipeline) {
-			tpCR.Spec.Pipeline = config.Spec.Pipeline
-			updated = true
+		_, err = CreatePipeline(ctx, clients, config)
+		if err != nil {
+			return nil, err
 		}
-
-		if !reflect.DeepEqual(tpCR.Spec.Config, config.Spec.Config) {
-			tpCR.Spec.Config = config.Spec.Config
-			updated = true
-		}
-
-		if tpCR.ObjectMeta.OwnerReferences == nil {
-			ownerRef := *metav1.NewControllerRef(config, config.GroupVersionKind())
-			tpCR.ObjectMeta.OwnerReferences = []metav1.OwnerReference{ownerRef}
-			updated = true
-		}
-
-		if updated {
-			return clients.Update(ctx, tpCR, metav1.UpdateOptions{})
-		}
-
-		return tpCR, err
+		return nil, v1alpha1.RECONCILE_AGAIN_ERR
 	}
 
-	if apierrs.IsNotFound(err) {
-		tpCR = &v1alpha1.TektonPipeline{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: v1alpha1.PipelineResourceName,
-			},
-			Spec: v1alpha1.TektonPipelineSpec{
-				CommonSpec: v1alpha1.CommonSpec{
-					TargetNamespace: config.Spec.TargetNamespace,
-				},
-				Pipeline: config.Spec.Pipeline,
-				Config:   config.Spec.Config,
-			},
-		}
-
-		return clients.Create(ctx, tpCR, metav1.CreateOptions{})
+	tpCR, err = UpdatePipeline(ctx, tpCR, config, clients)
+	if err != nil {
+		return nil, err
 	}
+
+	ok, err := isTektonPipelineReady(tpCR, err)
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		return nil, v1alpha1.RECONCILE_AGAIN_ERR
+	}
+
 	return tpCR, err
 }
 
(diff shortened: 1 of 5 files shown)


pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -123,10 +123,13 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
-	if err := pipeline.CreatePipelineCR(ctx, tc, r.operatorClientSet.OperatorV1alpha1()); err != nil {
+	// Ensure if the pipeline CR already exists, if not create Pipeline CR
+	if _, err := pipeline.EnsureTektonPipelineExists(ctx, r.operatorClientSet.OperatorV1alpha1().TektonPipelines(), tc); err != nil {
 		tc.Status.MarkComponentNotReady(fmt.Sprintf("TektonPipeline: %s", err.Error()))
-		r.enqueueAfter(tc, 10*time.Second)
+		if err == v1alpha1.RECONCILE_AGAIN_ERR {
+			r.enqueueAfter(tc, 10*time.Second)
+			return nil
+		}
 		return nil
```
nikhil-thomas, line 133:
```suggestion
		return err
```

if the error is not `RECONCILE_AGAIN_ERR` we have to return the error

pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
```diff
@@ -20,35 +20,51 @@ import (
...
+						TargetNamespace: config.Spec.TargetNamespace,
+					},
+					Pipeline: config.Spec.Pipeline,
+					Config:   config.Spec.Config,
+				},
+			}
+			_, err = clients.Create(ctx, tpCR, metav1.CreateOptions{})
+			if err != nil {
+				return nil, err
+			}
```
nikhil-thomas, line 54:
```suggestion
			tp, err = clients.Create(ctx, tpCR, metav1.CreateOptions{})
			if err != nil {
				return nil, err
			}
			return tp, v1alpha1.RECONCILE_AGAIN_ERR
```

if err is nil, we return `RECONCILE_AGAIN_ERR` 

1.  check a small piece of state (eg: does pipeline cr exists)
     1.a if the check is false, take smallest unit of action towards reaching the checked state
     1.b  return and try the check in step 1
2. check next small piece of state (eg: is pipeline cr valid or ready)
...

nikhil-thomas: in other words, the reconciler should try `ok, err := isTektonPipelineReady(tpCR, err)` 
only if the previous check `tpCR, err := GetPipeline(ctx, clients, v1alpha1.PipelineResourceName)` was successful in the same loop.

note: This is pattern  is not just something that we need to follow in this specific function or this operator. This is the core tenet of declarative coding paradigm in k8s. 🧑‍💻 


~~~

         |                                    |
         |                                    |
         |                                    |
       /   \                                  |
      /     \                                 | Requeue reconcile
     /       \   false ____________________   |
    / check 1 \________| Act on the cluster|__|
    \         /        |___________________|  |
     \       /                                |
       \   /                                  |
         |                                    |
         | true                               |
         |                                    |
       /   \                                  |
      /     \                                 | Requeue reconcile
     /       \   false ____________________   |
    / check 2 \________| Act on the cluster|__|
    \         /        |___________________|
     \       /
       \   /
         |
         | true
         |
         .
         .
         .
~~~

nikhil-thomas: yeah, the logic looks right, i was tripped by the `else` 🤓

pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
```diff
@@ -20,35 +20,51 @@ import (
...
+				},
+				Spec: v1alpha1.TektonPipelineSpec{
+					CommonSpec: v1alpha1.CommonSpec{
+						TargetNamespace: config.Spec.TargetNamespace,
+					},
+					Pipeline: config.Spec.Pipeline,
+					Config:   config.Spec.Config,
+				},
+			}
+			_, err = clients.Create(ctx, tpCR, metav1.CreateOptions{})
```
nikhil-thomas, line 51: just an observation, here we have a separate function for 1 line get operation but not separate function for the multiline create operation. 

If we could extract the create logic (initialize struct and call create Pipeline api) into a function, this code block will have better flow. 🧑‍💻

pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
```diff
@@ -20,35 +20,51 @@ import (
...
-		isTektonPipelineReady); err != nil {
-		log.Println("TektonPipeline is not in ready state: ", err)
-		return err
+
+	ok, err := isTektonPipelineReady(tpCR, err)
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		return nil, v1alpha1.RECONCILE_AGAIN_ERR
```
nikhil-thomas, line 65: in the reconcile logic, we need to check if an upadate is needed before checking isReady. If we check ifReady before checking if an update is needed, then the reconciler will never reach the update check.

so the reconciler has to check the cluster state in the following order.
1. check if exists
2. check if update needed
3. check if ready

PuneetPunamiya: yeah make sense
I had one question, in both the states, i.e. while updating and while creation we need to check if the pipeline is ready or not right ?

nikhil-thomas: while creation
1. create api call then reconcile again if no error
1. in the next loop if update needed make update api call, then reconcile again if no error, if update is not needed go to next check
1. in the next loop if update was needed or in the same loop if update was not needed,  check for ready condition, if not ready reconcile again, if err return err

so the check cascade in different execution context of the reconcile function not in the same running instance (context) of reconciler.

pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
```diff
@@ -20,35 +20,31 @@ import (
...
+		} else {
+			return nil, v1alpha1.RECONCILE_AGAIN_ERR
+		}
 	}
-	return nil
-}
 
-func ensureTektonPipelineExists(ctx context.Context, clients op.TektonPipelineInterface, config *v1alpha1.TektonConfig) (*v1alpha1.TektonPipeline, error) {
-	tpCR, err := GetPipeline(ctx, clients, v1alpha1.PipelineResourceName)
 	if err == nil {
```
nikhil-thomas, line 48:
```suggestion
	
```
we don't need this explicity `err == nil` check as that is the default code path here.

nikhil-thomas: and it will be great if we can move lines `50 updated := flase` to `74 tpCR, err = clients.Update(ctx, tpCR, metav1.UpdateOptions{})` into a function

PuneetPunamiya: yes I was thinking the same 
So our code would have functions like
```
Get pipeline
Create pipeline
Update pipeline
```

pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
```diff
@@ -20,108 +20,104 @@ import (
...
-			return clients.Update(ctx, tpCR, metav1.UpdateOptions{})
+	if err != nil {
+		if apierrs.IsNotFound(err) {
+			tpCR, err = CreatePipeline(ctx, clients, config)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, v1alpha1.RECONCILE_AGAIN_ERR
 		}
```
nikhil-thomas, line 45:
```suggestion
	if err != nil {
		if apierrs.IsNotFound(err) {
			tpCR, err = CreatePipeline(ctx, clients, config)
			if err != nil {
				return nil, err
			}
			return nil, v1alpha1.RECONCILE_AGAIN_ERR
		} 
		return nil, err
```

nikhil-thomas: if any of the calls return an actual error, we return it as it is. When there is no error from Create call, we return a `Reconcile_Again` to make the reconciler come back.

pkg/reconciler/shared/tektonconfig/pipeline/pipeline.go
```diff
@@ -20,108 +20,104 @@ import (
...
+		updated = true
+	}
+
+	if updated {
+		tpCR, err := clients.Update(ctx, tpCR, metav1.UpdateOptions{})
+		if err != nil {
+			return nil, err
+		}
+		return tpCR, nil
 	}
```
nikhil-thomas, line 119:
```suggestion
	if updated {
		tpCR, err := clients.Update(ctx, tpCR, metav1.UpdateOptions{})
		if err != nil {
			return nil, err
		}
		return nil, v1alpha1.RECONCILE_AGAIN_ERR
	}
```

As a general best practice, if the reconciler (or subfuctions) changes the cluster (API data or trigger a state change on cluster), then we reconcile get out and retry.

nikhil-thomas:
@PuneetPunamiya we also need similar changes for triggers in the code block following the the changes this patch makes. 

Could you handle that also in this patch.

pkg/reconciler/shared/tektonconfig/trigger/trigger.go
```diff
@@ -170,3 +167,17 @@ func verifyNoTektonTriggerCR(ctx context.Context, clients op.TektonTriggerInterf
...
+			Name: v1alpha1.ConfigResourceName,
+		},
+		Spec: v1alpha1.TektonConfigSpec{
+			Profile: "all",
+			CommonSpec: v1alpha1.CommonSpec{
+				TargetNamespace: "tekton-pipelines",
+			},
+		},
+	}
+}
```
nikhil-thomas, line 183: if this is only used in testing, could you move this to the test file.

PuneetPunamiya: sure will move it to test file

nikhil-thomas: will take it up later. could you create a test improvement issue. 
Im ok with merging this.

PuneetPunamiya: yeah will create an issue for this 👍🏻

vdemeester (approved):
/lgtm

PR #634, opened 2022-02-25
[OpenShift] Fix RBAC bugs

nikhil-thomas:

# Changes

- Rename rolebinding created by RBAC reconciler to
  'openshift-pipelines-edit' (was 'edit' earlier)

- Add mechanism to ensure that missing RBAC resources are recreated if
  the RBAC installerSet is recreated during an upgrade.
  - this ensures that RBAC in the version label user 
    namespaces are removed during an upgrade and the presence of RBAC
resources are re-verified

- Make TektonConfig reconciler listen to TektonInstallerSet events so
  that it can recreate RBAC InstallerSet if it is deleted manually from
a cluster

- Add a mechanism to remove ownerReference and 'pipeline' sa subject
  from 'edit' rolebinding in namespaces
  - This ensures that operator upgrades won't delete/reset 'edit'
    rolebinding in usernamespaces.
  - This mechanism will delete the edit role binding if the only subject is `pipeline` serviceaccount.
    

# Submitter Checklist

review them:

# Release Notes

```release-note
Name of rolebinding which binds ClusterRole edit to 'pipeline' ServiceAccount in usernamespaces is now 'openshift-pipelines-edit' instead of 'edit'.
```

### DIFF
--- a/pkg/reconciler/openshift/tektonconfig/common.go
+++ b/pkg/reconciler/openshift/tektonconfig/common.go
@@ -103,40 +103,37 @@ func deleteInstallerSet(ctx context.Context, oc versioned.Interface, tc *v1alpha
 // and if installer set which already exist is of older version then it deletes and return false to create a new
 // installer set
 func checkIfInstallerSetExist(ctx context.Context, oc versioned.Interface, relVersion string,
-	tc *v1alpha1.TektonConfig, component string) (bool, error) {
+	tc *v1alpha1.TektonConfig, component string) (*v1alpha1.TektonInstallerSet, error) {
 
 	// Check if installer set is already created
 	compInstallerSet, ok := tc.Status.TektonInstallerSet[component]
-	if !ok {
-		return false, nil
+	if !ok || compInstallerSet == "" {
+		return nil, nil
 	}
 
-	if compInstallerSet != "" {
-		// if already created then check which version it is
-		ctIs, err := oc.OperatorV1alpha1().TektonInstallerSets().
-			Get(ctx, compInstallerSet, metav1.GetOptions{})
-		if err != nil {
-			if errors.IsNotFound(err) {
-				return false, nil
-			}
-			return false, err
+	// if already created then check which version it is
+	ctIs, err := oc.OperatorV1alpha1().TektonInstallerSets().
+		Get(ctx, compInstallerSet, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
 		}
-
-		if version, ok := ctIs.Annotations[v1alpha1.ReleaseVersionKey]; ok && version == relVersion {
-			// if installer set already exist and release version is same
-			// then ignore and move on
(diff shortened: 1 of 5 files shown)


pkg/reconciler/openshift/tektonconfig/common.go
```diff
@@ -103,40 +103,37 @@ func deleteInstallerSet(ctx context.Context, oc versioned.Interface, tc *v1alpha
...
 // installer set
 func checkIfInstallerSetExist(ctx context.Context, oc versioned.Interface, relVersion string,
-	tc *v1alpha1.TektonConfig, component string) (bool, error) {
+	tc *v1alpha1.TektonConfig, component string) (*v1alpha1.TektonInstallerSet, error) {
 
 	// Check if installer set is already created
 	compInstallerSet, ok := tc.Status.TektonInstallerSet[component]
-	if !ok {
-		return false, nil
+	if !ok || compInstallerSet == "" {
```
sm43, line 110: do we require this check if the name exist in status?

i understand we create only one for rbac
but as we aim to make logic consistent across all platform and
later move this function in a common pkg
so that we can re use,
we can skip this check and query using label

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -34,16 +35,21 @@ import (
...
-	pipelinesSCCClusterRole  = "pipelines-scc-clusterrole"
-	pipelinesSCCRoleBinding  = "pipelines-scc-rolebinding"
-	pipelinesSCC             = "pipelines-scc"
-	pipelineSA               = "pipeline"
+	pipelinesSCCClusterRole = "pipelines-scc-clusterrole"
+	pipelinesSCCRoleBinding = "pipelines-scc-rolebinding"
+	pipelinesSCC            = "pipelines-scc"
+	pipelineSA              = "pipeline"
+	PipelineRoleBinding     = "openshift-pipelines-edit"
+	// TODO: Remove this after 1.7.x releases
```
vdemeester, line 43: Note: "1.7.x releases" is a downstream "thing", it doesn't have any meaning here (upstream). We should most likely say "TODO: Remove this after a deprecation notice time, which would be 3 weeks" or something..

nikhil-thomas: i agree :)

pkg/reconciler/openshift/tektonconfig/common.go
nikhil-thomas: yes, absolutely we can.
in this patch i was focusing on fixing the bug without changing anything around it.

At present, the name of the rbac-istallerset does not change (name: "rbac-resources") so the risk of duplication is less.

nikhil-thomas: In a future refactoring we need make the installerset(s) create and manage all rbac related things. This would have been a huge change for next release in the given time frame.

vdemeester (approved):

PR #636, opened 2022-02-28
Updates Hub Db deployment in case of Openshift

PuneetPunamiya:
# Changes

-   This patch adds a transformer which updates the db deployment,
    in case of openshift to use productized postgres images and also
    replace a few env and commands for readiness and liveness probes
    
# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/openshift/tektonhub/extension.go
+++ b/pkg/reconciler/openshift/tektonhub/extension.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/go-logr/zapr"
 	mfc "github.com/manifestival/client-go-client"
@@ -31,7 +32,11 @@ import (
 	operatorclient "github.com/tektoncd/operator/pkg/client/injection/client"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"go.uber.org/zap"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	"knative.dev/pkg/injection"
@@ -43,6 +48,12 @@ const (
 	tektonHubAuthResourceKey string = "auth"
 	tektonHubUiResourceKey   string = "ui"
 )
+
+var replaceVal = map[string]string{
+	"POSTGRES_DB":       "POSTGRESQL_DATABASE",
+	"POSTGRES_USER":     "POSTGRESQL_USER",
+	"POSTGRES_PASSWORD": "POSTGRESQL_PASSWORD",
+}
 
 func OpenShiftExtension(ctx context.Context) common.Extension {
 	logger := logging.FromContext(ctx)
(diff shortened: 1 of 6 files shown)


pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -196,3 +201,64 @@ func getRouteHost(manifest *mf.Manifest, routeName string) (string, error) {
...
+func replaceEnv(envs []corev1.EnvVar) {
+	for i, e := range envs {
+		if e.Name == "POSTGRES_DB" {
+			envs[i].Name = strings.ReplaceAll(e.Name, "POSTGRES_DB", "POSTGRESQL_DATABASE")
+		}
+		if e.Name == "POSTGRES_USER" {
+			envs[i].Name = strings.ReplaceAll(e.Name, "POSTGRES_USER", "POSTGRESQL_USER")
+		}
+		if e.Name == "POSTGRES_PASSWORD" {
+			envs[i].Name = strings.ReplaceAll(e.Name, "POSTGRES_PASSWORD", "POSTGRESQL_PASSWORD")
```
sm43, line 261: this could be simplied as as
```
_, ok := map[e.name]
if ok {
envs[i].Name = map[e.name]
}
```
and the map would be
```
replaceVal := map[string]string {
"POSTGRES_DB": "POSTGRESQL_DATABASE"
}
```

pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -196,3 +201,64 @@ func getRouteHost(manifest *mf.Manifest, routeName string) (string, error) {
 	}
 	return hostUrl, nil
 }
+
+func UpdateDbDeployment() mf.Transformer {
```
sm43, line 205: add unit test pls

vdemeester (approved):
/hold
We need the two dependent PR to get in before this one, but LGTM

pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -116,6 +128,34 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
...
+		mf.InjectNamespace(targetNs),
+	)
+	if err != nil {
+		logger.Error("failed to transform manifest")
+		return err
+	}
+
+	if err := uiManifest.Filter(mf.ByKind("Route")).Apply(); err != nil {
+		return err
+	}
```
nikhil-thomas, line 150:
```suggestion
        routeManifest := uiManifest.Filter(mf.ByKind("Route"))
	routeManifest, err = routeManifest.Transform(
		mf.InjectOwner(th),
		mf.InjectNamespace(targetNs),
	)
	if err != nil {
		logger.Error("failed to transform manifest")
		return err
	}

	if err := routeManifest.Apply(); err != nil {
		return err
	}
```

how about this, as we are only creating the routes in here

nikhil-thomas: same applies to lines 95 for API route

nikhil-thomas:
@PuneetPunamiya @piyush-garg should the route creation for ui be handled in the post-reconcile instead of pre-reconcile.

PuneetPunamiya:
> @PuneetPunamiya @piyush-garg should the route creation for ui be handled in the post-reconcile instead of pre-reconcile.

@nikhil-thomas  The reason we are creating the the ui route in the pre-reconciler is because the config map of ui will require the ui route, 

For this one option could be
- We create the route in `manageUiComponent` function before creating the configMap for ui, 
- The other alternative would be we can add a field in hub cr as `ui: true` which means the user wants ui as well and in the pre-reconciler we can check the value of `ui` and based on that value we create the ui route, but then we'll have to add a check in the reconciler as well to not go forward to create the ui component

PR #645, opened 2022-03-02
Improves error handling in TektonHub and reorder the job in TektonInstallerSet

PuneetPunamiya:
# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/kubernetes/tektonhub/tektonhub.go
+++ b/pkg/reconciler/kubernetes/tektonhub/tektonhub.go
@@ -139,19 +139,19 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, th *v1alpha1.TektonHub) 
 
 	// Manage DB
 	if err := r.manageDbComponent(ctx, th, hubDir, version); err != nil {
-		return err
+		return r.handleError(err, th)
 	}
 	th.Status.MarkDbInstallerSetAvailable()
 
 	// Manage DB migration
 	if err := r.manageDbMigrationComponent(ctx, th, hubDir, version); err != nil {
-		return err
+		return r.handleError(err, th)
 	}
 	th.Status.MarkDatabasebMigrationDone()
 
 	// Manage API
 	if err := r.manageApiComponent(ctx, th, hubDir, version); err != nil {
-		return err
+		return r.handleError(err, th)
 	}
 	th.Status.MarkApiInstallerSetAvailable()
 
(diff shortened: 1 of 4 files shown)


pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
```diff
@@ -166,6 +166,12 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, installerSet *v1alpha1.T
 	// Update Ready status of Controller
 	installerSet.Status.MarkControllerReady()
 
+	// job
+	err = installer.IsJobCompleted()
```
nikhil-thomas, line 170: Could you add a log here (info nor error) as the result of this check is not going to show up in any of the existing InstallerSet status conditions.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -139,18 +139,30 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, th *v1alpha1.TektonHub)
 
 	// Manage DB
 	if err := r.manageDbComponent(ctx, th, hubDir, version); err != nil {
+		if err == v1alpha1.RECONCILE_AGAIN_ERR {
+			r.enqueueAfter(th, 10*time.Second)
+			return nil
+		}
 		return err
 	}
```
nikhil-thomas, line 147:
```suggestion
	if err := r.manageDbComponent(ctx, th, hubDir, version); err != nil {
		return r.handleError(err)
	}
```

define a function
```
func (r *reconciler) handleError(err error) error {
    if err == v1alpha1.RECONCILE_AGAIN_ERR {
			r.enqueueAfter(th, 10*time.Second)
			return nil
		}
		return err
}
```

PuneetPunamiya: Yeah i did that, just wanted to confirm if the logic behind the errors was correct 👍🏻 
will update the patch

pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
```diff
@@ -166,6 +166,15 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, installerSet *v1alpha1.T
...
+	if err != nil {
+		job := installer.Manifest.Filter(mf.ByKind("Job")).Resources()[0]
+		labels := installerSet.GetLabels()
+		logger.Info("job not ready in installerset, name: %s, created-by: %s, in namespace: %s", installerSet.GetName(), labels[v1alpha1.CreatedByKey], job.GetNamespace())
+		return err
```
nikhil-thomas, line 175: this will work but let us move this to https://github.com/tektoncd/operator/blob/c6c96e1503dbc31f52a64e14e0f285af7624f476/pkg/reconciler/kubernetes/tektoninstallerset/install.go#L368

Doing so will make it easy to report on each job in the installerset instead of focusing on Job[0].

nikhil-thomas: on this line (368) we still need to return the error, but add the log right before it.

PuneetPunamiya: I have updated the patch could you please take a look at it once 🙂

vdemeester (approved):

PR #648, opened 2022-03-03
Mount CA bundles in Tekton components as well

concaf:
# Changes

Prior to this commit, the CA bundles were:
- mounted inside containers
- exposed via VolumeMounts and SSL_CERT_DIR
only in TaskRun pods via `pkg/reconciler/proxy/proxy.go` but were not
exposed in Tekton components controllers like pipelines, triggers and
chains via `pkg/reconciler/openshift/common/cabundle.go`.

This meant that these Tekton components could not talk to internal
OpenShift services or to hosts for which users added certificates to the
cluster via CA bundles.

One case where this is a problem is when the Chains controller wants to
pull images from the internal OpenShift registry to sign and then push
the signatures, attestations back to it since it does not have access to
the Service CA Bundle.

This commit abstracts away the exposure of certificates via Volumes,
VolumeMounts and SSL_CERT_DIR in `pkg/reconciler/common/certificates.go`
and makes this behavior the same for both TaskRuns and controllers which
want to mount these certificates.
# Submitter Checklist

review them:

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/common/certificates.go
+++ b/pkg/reconciler/common/certificates.go
@@ -1,0 +1,153 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


sm43:
LGTM :)

vdemeester (approved):

PR #664, opened 2022-03-09
Replace PAC images in TriggerTemplates

concaf:
# Changes

    Before this commit, the images in TriggerTemplates deployed by Pipelines
    as Code are not substituted with images required for OperatorHub
    deployment.
    
    This commit adds a transformer that replaces images per the environment
    variable `IMAGE_PAC_PIPELINES_AS_CODE`.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/common/transformers.go
+++ b/pkg/reconciler/common/transformers.go
@@ -40,6 +40,7 @@ const (
 	PipelinesImagePrefix          = "IMAGE_PIPELINES_"
 	TriggersImagePrefix           = "IMAGE_TRIGGERS_"
 	AddonsImagePrefix             = "IMAGE_ADDONS_"
+	PacImagePrefix                = "IMAGE_PAC_"
 	ChainsImagePrefix             = "IMAGE_CHAINS_"
 	HubImagePrefix                = "IMAGE_HUB_"
 
--- a/pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
+++ b/pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
 
 	mf "github.com/manifestival/manifestival"
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
@@ -75,7 +76,13 @@ func (r *Reconciler) ensurePAC(ctx context.Context, ta *v1alpha1.TektonAddon) er
 	pacManifest = pacManifest.Filter(mf.Not(mf.ByKind("Namespace")))
 
 	// Run transformers
-	if err := r.addonTransform(ctx, &pacManifest, ta); err != nil {
+	var tfs []mf.Transformer
+	// don't run the transformer if no replace images are found
+	if triggerTemplateSteps := pacTriggerTemplateStepImages(); len(triggerTemplateSteps) > 0 {
+		tfs = append(tfs, replacePACTriggerTemplateImages(triggerTemplateSteps))
+	}
+
+	if err := r.addonTransform(ctx, &pacManifest, ta, tfs...); err != nil {
 		return err
 	}
 
@@ -85,4 +92,27 @@ func (r *Reconciler) ensurePAC(ctx context.Context, ta *v1alpha1.TektonAddon) er
 	}
 
 	return nil
+}
+
+// pacTriggerTemplateStepImages returns a map[string]string with key as step name and
+// value as image name to be replaced with, from the env vars that start with
+// IMAGE_PAC_
+func pacTriggerTemplateStepImages() map[string]string {
+	triggerTemplateSteps := make(map[string]string)
+
+	// pacImage is a map[string]string which will have key-values like
+	// "triggertemplate_apply_and_launch": "registry.example.io/pac-image"
+	pacImages := common.ToLowerCaseKeys(common.ImagesFromEnv(common.PacImagePrefix))
+	for env, image := range pacImages {
+		prefix := "triggertemplate_"
+		if strings.HasPrefix(env, prefix) {
+			// step 3: "apply-and-launch": "registry.example.io/pac-image"
+			triggerTemplateSteps[
+			// step 2: apply_and_launch --> apply-and-launch
+			strings.ReplaceAll(
+				// step 1: triggertemplate_apply_and_launch --> apply_and_launch
+				strings.TrimPrefix(env, prefix), "_", "-")] = image
+		}
+	}
+	return triggerTemplateSteps
 }
--- a/pkg/reconciler/openshift/tektonaddon/transformer.go
+++ b/pkg/reconciler/openshift/tektonaddon/transformer.go
@@ -17,7 +17,11 @@ limitations under the License.
 package tektonaddon
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	pipelinev1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	triggersv1beta1 "github.com/tektoncd/triggers/pkg/apis/triggers/v1beta1"
 
 	mf "github.com/manifestival/manifestival"
 	console "github.com/openshift/api/console/v1"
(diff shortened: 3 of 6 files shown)


vdemeester (approved):
/cc @nikhil-thomas 
Needs to be backported to `release-v0.55.x`

vdemeester:
(ah it's already done in https://github.com/tektoncd/operator/pull/666)

concaf:
@vdemeester #666 is a cherry pick

piyush-garg:
You need to add the prefix here https://github.com/tektoncd/operator/blob/main/pkg/reconciler/common/transformers.go#L38

also add the transformer in reconciler, just adding env will not work

/hold

piyush-garg:
Also you need to write a new transformer, because this is a special case where image is getting replaced in taskspec of a resource of triggertemplate. Current logic is for deployment

pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
```diff
@@ -86,3 +94,12 @@ func (r *Reconciler) ensurePAC(ctx context.Context, ta *v1alpha1.TektonAddon) er
 
 	return nil
 }
+
+// getPACImage gets the PAC image name from the IMAGE_PAC_PIPELINES_AS_CODE env var
+func getPACImage() (string, error) {
+	pacImage := os.Getenv(common.PACImageEnv)
+	if pacImage == "" {
+		return "", fmt.Errorf("environment variable %v not found", common.PACImageEnv)
```
sm43, line 102: is it required to return an error here?
if image not found then we can continue with existing img right?
what happens in development case? when we do `make apply` locally
do we have a image in env?

pkg/reconciler/common/transformers.go
```diff
@@ -48,6 +48,8 @@ const (
 
 	resultAPIDeployment     = "tekton-results-api"
 	resultWatcherDeployment = "tekton-results-watcher"
+
+	PACImageEnv = "IMAGE_PAC_PIPELINES_AS_CODE"
```
piyush-garg, line 52: Should we consider this whole, or have a specific format? Like Having Triggertemplate in name so that we know this is for triggertemplates, otherwise if we go with the pattern folowed, this will look like it finds a container with name - `pipelines-as-code` in deployment

pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
piyush-garg, line 102: there can be a case like this env is not present in the development cycle, or in k8s target for now

pkg/reconciler/openshift/tektonaddon/transformer.go
```diff
@@ -133,3 +137,79 @@ func setVersionedNames(operatorVersion string) mf.Transformer {
...
+			// let' start by checking that we are only dealing with a TaskRun
+
+			// first we need to decode the raw ResourceTemplate
+			rawReader := mf.Reader(bytes.NewReader(resourceTemplate.RawExtension.Raw))
+			rawObjects, err := rawReader.Parse()
+			if err != nil {
+				return err
+			}
+
+			if len(rawObjects) == 0 {
```
piyush-garg, line 171: I am not able to think of any case making this

pkg/reconciler/openshift/tektonaddon/transformer.go
```diff
@@ -133,3 +137,79 @@ func setVersionedNames(operatorVersion string) mf.Transformer {
...
+			}
+
+			// now that we know it's a TaskRun, let's decode and move forward
+			taskRun := pipelinev1beta1.TaskRun{}
+			decoder := json.NewDecoder(bytes.NewBuffer(resourceTemplate.RawExtension.Raw))
+			if err := decoder.Decode(&taskRun); err != nil {
+				return err
+			}
+
+			for j, step := range taskRun.Spec.TaskSpec.Steps {
```
piyush-garg, line 191: this method with replace any image in taskruns of triggertemplate, we should fins a pattern with name or something that pac image is replaced, this may result in replacing images of any other triggertemplates also

pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
concaf: good catch, fixed in https://github.com/tektoncd/operator/pull/664/commits/1d5be1021ef4faedc26493b48628efb90e6012b8

pkg/reconciler/common/transformers.go
concaf: I see how this can be confusing, I'll rename it to `IMAGE_PAC_TRIGGER_TEMPLATES`

pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
concaf: fixed in https://github.com/tektoncd/operator/pull/664/commits/1d5be1021ef4faedc26493b48628efb90e6012b8

pkg/reconciler/common/transformers.go
concaf: Fixed in e1baf385c4f2012d01c33ca0783f29ac912f6d24

pkg/reconciler/openshift/tektonaddon/transformer.go
concaf: me neither, but I'd rather err on the side of caution here 🙃

pkg/reconciler/openshift/tektonaddon/transformer.go
concaf: I thought about this as well, but there is no pattern here. I'll ping @chmouel and @sm43 if PAC can guarantee same step names in the future.

chmouel: what kind of guarantee you need? i am not quite understanding the question,

concaf: @chmouel any pattern that can we identify to replace the image, like step name, etc.
For now, I'll move ahead with what @piyush-garg has suggested.

nikhil-thomas: or it could be any label in the triggerTemplate which this function could use to ensure that the triggertemplate it is working on is from PAC. 

cc @piyush-garg @concaf may be we can keep PAC project as it is and add label through [this](https://github.com/tektoncd/operator/blob/42d68c8428211e993b8455d7de292e0773fc54c9/hack/fetch-releases.sh#L97) if needed.

concaf: @nikhil-thomas for now, since we don't have any other TriggerTemplate transformer, the current solution should work but going forward we could do something like that, WDYT?

@piyush-garg I've made the changes at 668b77014b8107e0f24ccdda5c25ddbbefb8be71

concaf: @nikhil-thomas thinking about it, what you said makes sense 😅 added a check here - f43d62ad909f6dc783c22631b98dbaeb9022e465

pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
```diff
@@ -86,3 +93,26 @@ func (r *Reconciler) ensurePAC(ctx context.Context, ta *v1alpha1.TektonAddon) er
...
 }
+
+// pacTriggerTemplateStepImages returns a map[string]string with key as step name and
+// value as image name to be replaced with, from the env vars that start with
+// IMAGE_PAC_
+func pacTriggerTemplateStepImages() map[string]string {
+	triggerTemplateSteps := make(map[string]string)
+
+	// pacImage is a map[string]string which will have key-values like
+	// "TRIGGERTEMPALTE_APPLY_AND_LAUNCH": "registry.example.io/pac-image"
```
piyush-garg, line 104: nit: // "triggertemplate_apply_and_launch": "registry.example.io/pac-image"

pkg/reconciler/openshift/tektonaddon/pipelinesascode.go
```diff
@@ -86,3 +93,26 @@ func (r *Reconciler) ensurePAC(ctx context.Context, ta *v1alpha1.TektonAddon) er
...
+	pacImages := common.ToLowerCaseKeys(common.ImagesFromEnv(common.PacImagePrefix))
+	for env, image := range pacImages {
+		prefix := "triggertemplate_"
+		if strings.HasPrefix(env, prefix) {
+			// step 3: "apply-and-launch": "registry.example.io/pac-image"
+			triggerTemplateSteps[
+			// step 2: apply_and_launch --> apply-and-launch
+			strings.ReplaceAll(
```
piyush-garg, line 112: is not better to have this replace on step name inside transformer, so any one using this function has not need to do this every time and also it will be consistent with other transformers

concaf: which replace are you referring to that happens in other transformers?
https://github.com/concaf/operator/blob/307eafa4a3cce8f01e2354e651c50afb84f3ae93/pkg/reconciler/common/transformers.go#L262
^ `formKey` converts container names to env var names so we can match it, but that's not what I'm doing here. I'm normalizing the step names before so the transformer gets the right step names instead of having to modify it later.

anyway, I'd rather have the transformer get the correct normalized step names as input instead of having the modify and match inside it.

piyush-garg: Yes the idea behind this is, anyone calling this function does not need to do a modification for all calls, rather the transformer is doing the normalization and this is the same thing for all transformers

concaf: normalization is happening out of the transformer in other calls except the `formKey` 🤔

piyush-garg: I don't have any other strong points behind keeping it inside and outside apart from these two

1. Consistency with the codebase
2. Benefit of having every caller does not need to do the modification

I will let the owners of codebase take the call cc @tektoncd/operator-maintainers

piyush-garg: we can discuss this in WG and can do a follow-up PR for the pattern we agree

concaf: I think the other transformers are a bit inconsistent in the way of normalizing input names. Let's take an example of `common.DeploymentImages(images)` https://github.com/tektoncd/operator/blob/307eafa4a3cce8f01e2354e651c50afb84f3ae93/pkg/reconciler/common/transformers.go#L177

It takes `images` as input which I'd expect to be a mapping of `container name: image name` and this is true to an extent but not fully.

Some normalization of the container names happens outside the transformer here:
https://github.com/tektoncd/operator/blob/307eafa4a3cce8f01e2354e651c50afb84f3ae93/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L338 where the prefix is removed, lowercased, etc but `_` is not yet replaced by `-` which happens inside the transformer instead - https://github.com/tektoncd/operator/blob/307eafa4a3cce8f01e2354e651c50afb84f3ae93/pkg/reconciler/common/transformers.go#L230

I didn't want this inconsistent normalization to be happening for this transformer and I think it's the caller's responsibility to provide normalized container names and not the transformer's; unless you want to tie the transformer to env variables which I don't think is ideal because it reduces the transformers' scope by quite a bit and it cannot be called for non env var cases.

At some point, we should consider doing normalization outside of other transformers as well 🤔

Issue #670, opened 2022-03-10
Create operator-info config map in operator's namespace


nikhil-thomas:
At present, the operator-info configMap is created in teh `targetNamespace` specified in tektonConfig instance. This was done assuming that it will be easier for users and tools to find version information if all the *-info configMaps are in the same namespace (targetNamespace). 

This is confusing as most of our users expect the configMap to be in the namespace where the operator is installed (eg: tekton-operator (k8s), openshift-operators(openshift)). 

# Expected

operator-info configMap is created in the namespace where operator is installed. (eg: tekton-operator (k8s), openshift-operators(openshift))

PR #673, opened 2022-03-11
Creates the operator version configMap in the operator namespace

PuneetPunamiya:
Initially the operator version configMap was created in the targetNs,
assuming that it will be easier for users and tools to find version
information

But this is confusing as most of our users expect the configMap to be
in the namespace where the operator is installed, hence this patch
creates the configMap in the namespace where the operator is installed

Fixes: #670 

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/shared/tektonconfig/tektonconfig.go
@@ -104,10 +104,6 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 		return err
 	}
 
-	if err := common.CreateOperatorVersionConfigMap(r.manifest, tc); err != nil {
-		return err
-	}
-
 	if err := r.extension.PreReconcile(ctx, tc); err != nil {
 		if err == v1alpha1.RECONCILE_AGAIN_ERR {
 			r.enqueueAfter(tc, 10*time.Second)
--- a/config/base/kustomization.yaml
+++ b/config/base/kustomization.yaml
@@ -29,3 +29,4 @@ resources:
 - operator.yaml
 - tekton_config_role.yaml
 - tekton_config_role_binding.yaml
+- operator-config.yaml
--- a/config/base/operator-config.yaml
+++ b/config/base/operator-config.yaml
@@ -1,0 +1,24 @@
+# Copyright 2022 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tekton-operator-info
+  labels:
+    app.kubernetes.io/instance: default
+data:
+  # Contains operator version which can be queried by external
+  # tools such as CLI.
+  version: 'devel'
--- a/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
+++ b/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
@@ -372,9 +372,6 @@ func (r *Reconciler) targetNamespaceCheck(ctx context.Context, tp *v1alpha1.Tekt
 			if err := common.CreateTargetNamespace(ctx, labels, tp, r.kubeClientSet); err != nil {
 				return err
 			}
-			if err := common.CreateOperatorVersionConfigMap(r.manifest, tp); err != nil {
-				return err
-			}
 			return nil
 		}
 		return err


config/kubernetes/base/operator-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operators-info
```
concaf, line 18: Thinking about OpenShift, since this is going to be created in `openshift-operators` where other operators exist as well, will the name `operators-info` not be confusing? How about renaming it to `tekton-operator-info` or something that's more descriptive?

PuneetPunamiya: yes I agree, but in that case we'll have to think about the name for pipelines version configMap as well because
the configMap name is `pipelines-info` in `openshift-pipelines` namespace
wdyt ?

concaf: ... maybe? I think the broader problem is that there can be other component's operators installed in `openshift-operators` namespace as well and `operators-info` seems like a generic name for all the operators. About `pipelines-info`, it at least belongs to OpenShift Pipelines so it's not as confusing maybe?

PuneetPunamiya: yep that make sense as well, just had a question, would it break the consistency, i.e. all components have name as `<component's-name>-info` ?

cc @nikhil-thomas

concaf: good point, we can consider naming all the configmaps to `tekton-<component>-info` as well 🤔

nikhil-thomas: for all the other components, as the config-maps live in the targetNamespace, we can assume that it will create less confusion. For operator this could be confusing especially on OpenShift as we developers or users cannot change the installation namespace (because OLM expects our operator to be in openshift-operators.. long story 🤓 ). 

I think in the scope of this pr, let us rename the operator configmap to `tekton-operator-info`. 
cc @piyush-garg

nikhil-thomas: > good point, we can consider naming all the configmaps to tekton-<component>-info as well 🤔

we can. 
However, there was a community wide TEP to get consensus on the current naming 🤓 . So we might have to take the TEP path if we want to change it. The reason is the `tkn version` command depend upon these names.

[TEP-0041: Tekton Component Versioning](https://github.com/tektoncd/community/blob/main/teps/0041-tekton-component-versioning.md)

nikhil-thomas: i think this is another `namespaces are one honking great idea` moment 😇

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -100,14 +99,6 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
 	}
 
-	if err := r.ensureTargetNamespaceExists(ctx, tc); err != nil {
-		return err
-	}
-
-	if err := common.CreateOperatorVersionConfigMap(r.manifest, tc); err != nil {
-		return err
-	}
-
```
nikhil-thomas, line 110: @PuneetPunamiya i think we should retain all the other code except this one. 

i think, creation of targetnamespace from tetkonConfig reconcier is a great idea as it gives us better control over the ownership of target namespace.

nikhil-thomas: in my opinion, this pr should remove only
```golang
if err := common.CreateOperatorVersionConfigMap(r.manifest, tc); err != nil {
		return err
	}
```

cc @sm43

sm43: @nikhil-thomas 
Yep, creating namespace in TektonConfig might be a good idea if we can add tektonconfig as owner to namespace.

PuneetPunamiya: yeah make sense 👍🏻 
I've updated the patch

config/kubernetes/base/operator-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tekton-operators-info
+  labels:
+    app.kubernetes.io/instance: default
+data:
+  # Contains operator version which can be queried by external
+  # tools such as CLI.
+  version: 'devel'
```
nikhil-thomas, line 24: as the content is the same for kubernetes and openshfit, shall we move this configMap intohttps://github.com/tektoncd/operator/tree/main/config/base

config/kubernetes/base/operator-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tekton-operators-info
```
concaf, line 18:
```suggestion
  name: tekton-operator-info
```
per this comment https://github.com/tektoncd/operator/pull/673#discussion_r825686491

config/openshift/base/operator-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tekton-operators-info
```
concaf, line 18:
```suggestion
  name: tekton-operator-info
```
per this comment https://github.com/tektoncd/operator/pull/673#discussion_r825686491

concaf (changes requested):

sm43:
@PuneetPunamiya @nikhil-thomas 
if we are letting TektonConfig create namespace which is okay but we need to see whether
later namespace is getting overwritten by any other component.. (with owner ref)
ClusterScope resources like admission controller have their owner ref as Namespace
so we need to make sure namespace is deleted when TC is deleted.

PR #708, opened 2022-03-21
This will fix scheduling the cron jobs pod on infra node if configured

rupalibehera:
This will fix scheduling the cron jobs pod on infra node if configured in TektonConfig CR
1. Modified the checkAndCreate function to also check if there is a change
   in TektonConfig.Spec.Config
2. Adding NodeSelector and Tolerations while creating CronJob
3. Added unit test TestNodeSelectorChange to check if nodeSelector and Toleration change creates a new Cron Job

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/common/prune.go
+++ b/pkg/reconciler/common/prune.go
@@ -129,7 +129,7 @@ func Prune(ctx context.Context, k kubernetes.Interface, tC *v1alpha1.TektonConfi
 	}
 	if pruningNamespaces.commonScheduleNs != nil && len(pruningNamespaces.commonScheduleNs) > 0 {
 		jobs := getJobContainer(generateAllPruneConfig(pruningNamespaces.commonScheduleNs), pruner.targetNamespace, pruner.tknImage)
-		if err := pruner.checkAndCreate(ctx, "", tC.Spec.Pruner.Schedule, jobs, pruneCronLabel, pruningNamespaces.configChanged); err != nil {
+		if err := pruner.checkAndCreate(ctx, "", tC.Spec.Pruner.Schedule, jobs, pruneCronLabel, pruningNamespaces.configChanged, tC); err != nil {
 			logger.Error("failed to create cronjob ", err)
 		}
 	}
@@ -138,7 +138,7 @@ func Prune(ctx context.Context, k kubernetes.Interface, tC *v1alpha1.TektonConfi
 		for ns, con := range pruningNamespaces.uniqueScheduleNS {
 			jobs := getJobContainer(generatePruneConfigPerNamespace(con, ns), ns, pruner.tknImage)
 			listOpt := pruneCronNsLabel + "=" + ns
-			if err := pruner.checkAndCreate(ctx, ns, con.config.Schedule, jobs, listOpt, pruningNamespaces.configChanged); err != nil {
+			if err := pruner.checkAndCreate(ctx, ns, con.config.Schedule, jobs, listOpt, pruningNamespaces.configChanged, tC); err != nil {
 				logger.Error("failed to create cronjob ", err)
 			}
 		}
@@ -262,7 +262,12 @@ func getJobContainer(cmdArgs, ns string, tknImage string) []corev1.Container {
 	return []corev1.Container{container}
 }
 
-func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedule string, pruneContainers []corev1.Container, listOpt string, configChanged bool) error {
+func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedule string, pruneContainers []corev1.Container, listOpt string, configChanged bool, tC *v1alpha1.TektonConfig) error {
+	// calculate tektonConfig.Spec.Config hash to check if nodeSelector and tolerations have changed
+	computeTektonConfigSpecConfigHash, err := hash.Compute(tC.Spec.Config)
+	if err != nil {
+		return err
+	}
 	suffixedCronName := SimpleNameGenerator.RestrictLengthWithRandomSuffix(CronName)
 	cronList, err := pruner.listCronJobs(ctx, pruner.targetNamespace, listOpt)
 	if err != nil {
@@ -275,22 +280,49 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
 				return err
 			}
 		}
-		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef)
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTektonConfigSpecConfigHash)
 	}
 
 	// no change in config but the cronjob does not exist
 	if len(cronList.Items) == 0 && !configChanged {
-		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef)
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTektonConfigSpecConfigHash)
 	}
 
 	// any case with config change
 	if configChanged {
-		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef)
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTektonConfigSpecConfigHash)
+	}
+	// In case nodeSelector or Tolerations changed
+	if len(cronList.Items) > 0 {
+		// get the cronjob to check previous annotations
+		annotations := cronList.Items[0].Annotations
+		nodeSelectorOrTolerationsChanged := checkTektonConfigSpecConfigHashChange(annotations, computeTektonConfigSpecConfigHash)
+
+		if nodeSelectorOrTolerationsChanged {
+			for _, cronjob := range cronList.Items {
+				if err := pruner.deleteCronJob(ctx, cronjob.Name); err != nil {
+					return err
+				}
+			}
+
+			return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTektonConfigSpecConfigHash)
+
+		}
 	}
 	return nil
 }
 
-func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targetNs, pruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference) error {
+func checkTektonConfigSpecConfigHashChange(annotations map[string]string, currentComputedHash string) bool {
+	// check if the annotation in the cronJob is same or changed.
+	lastHash, ok := annotations[v1alpha1.LastAppliedHashKey]
+	if ok {
+		return lastHash != currentComputedHash
+	}
+	return true
+}
+
+func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targetNs, pruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference, tC *v1alpha1.TektonConfig, computedTektonConfigSpecConfigHash string) error {
+
 	backOffLimit := int32(3)
 	ttlSecondsAfterFinished := int32(3600)
 	cj := &batchv1.CronJob{
@@ -302,6 +334,7 @@ func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targe
 			Name:            cronName,
 			OwnerReferences: []v1.OwnerReference{oR},
 			Labels:          map[string]string{pruneCronLabel: "true"},
+			Annotations:     map[string]string{v1alpha1.LastAppliedHashKey: computedTektonConfigSpecConfigHash},
 		},
 		Spec: batchv1.CronJobSpec{
 			Schedule:          schedule,
@@ -317,6 +350,8 @@ func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targe
 							Containers:         pruneContainers,
 							RestartPolicy:      "OnFailure",
 							ServiceAccountName: tektonSA,
+							NodeSelector:       tC.Spec.Config.NodeSelector,
+							Tolerations:        tC.Spec.Config.Tolerations,
 						},
 					},
 				},
--- a/pkg/reconciler/common/prune_test.go
+++ b/pkg/reconciler/common/prune_test.go
@@ -31,8 +31,9 @@ import (
 )
 
 const (
-	scheduleCommon = "*/2 * * * *"
-	scheduleUnique = "*/4 * * * *"
+	scheduleCommon  = "*/2 * * * *"
+	scheduleUnique  = "*/4 * * * *"
+	scheduleUnique2 = "*/6 * * * *"
 )
 
 func TestGetPrunableNamespaces(t *testing.T) {
@@ -109,6 +110,7 @@ func TestCompleteFlowPrune(t *testing.T) {
 		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-four", Annotations: anno4}},
 	)
 	os.Setenv(JobsTKNImageName, "some")
+	defer os.Unsetenv(JobsTKNImageName)
 
 	err := Prune(context.TODO(), client, config)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/reconciler/common/prune.go
```diff
@@ -279,22 +280,42 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
-		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef)
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC)
 	}
+
 	return nil
 }
+func getPodSpec(pruneContainers []corev1.Container, tC *v1alpha1.TektonConfig) corev1.PodSpec {
+	targetNamespace := tC.Spec.TargetNamespace
+
+	if targetNamespace == "openshift-pipelines" {
```
nikhil-thomas, line 301: i see that we are using this check to determine if the platform is openshift. 
This is not a highly reliable way to check if the platform is openshift as users or olm dependency resolution can install the operator in a different namespace.

Instead, we should check for a presence of an OpenShift specific Resource for example any CRD that is only seen in OpneShift like ConsoleCLI, SampleYamls etc.

nikhil-thomas: cc @pradeepitm12 @piyush-garg

rupalibehera: Yes not very reliable way, let me change this.

pkg/reconciler/common/prune.go
```diff
@@ -422,7 +439,9 @@ func (pruner *Pruner) checkAnnotationsRemovedNamespaces(ctx context.Context) ([]
 	return uniqueCronName, err
 }
 
-func diffLastAppliedPruneConfig(ctx context.Context, k kubernetes.Interface, defaultconf v1alpha1.Prune, ns *corev1.Namespace) (bool, error) {
+func diffLastAppliedPruneConfig(ctx context.Context, k kubernetes.Interface, tC *v1alpha1.TektonConfig, ns *corev1.Namespace) (bool, error) {
+	// Pruner Config
+	defaultconf := tC.Spec.Pruner
```
pradeepitm12, line 444: I think here we need `tC.Spec.Config` instead of `tC.Spec.Pruner`, as we want to check if there is a change in NodeSelector  or Tolerations

rupalibehera: we have computed the `tC.Spec.Config` below here I just added a comment so that we do not get confused with Pruner config and there is a difference between them

pradeepitm12: How about creating a different check itself for the tolerations and nodeselectors. 
And to check the last applied config we can set an annotation on the cronjob itself.

rupalibehera:
@pradeepitm12 @nikhil-thomas could you guys take a look again

pkg/reconciler/common/prune.go
```diff
@@ -266,7 +265,12 @@ func (pruner *Pruner) createJobContainers(nsConfig *pruneConfigPerNS, ns string)
 	return containers
 }
 
-func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedule string, pruneContainers []corev1.Container, listOpt string, configChanged bool) error {
+func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedule string, pruneContainers []corev1.Container, listOpt string, configChanged bool, tC *v1alpha1.TektonConfig) error {
+	// calculate tektonConfig hash
+	computeTcConfigHash, err := hash.Compute(tC.Spec.Config)
```
pradeepitm12, line 270: Should we change `computeTcConfigHash` to `nodeselectorConfigHash`, as we are computing the hash for the NodeSelector and Tolerations .

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +283,74 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
 	}
 
 	// any case with config change
 	if configChanged {
-		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef)
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTcConfigHash)
+	}
+	// get the cronjob to check previous annotations
+	cronjob := cronList.Items[0]
+	cronjobName := cronjob.Name
```
pradeepitm12, line 300: How about adding a check here, if there are no cronjobs then we may get a nil Pointer exception here.

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +283,74 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
+	annotations := cronjob.Annotations
+	tektonConfigChanged := checkTektonConfigHashChange(annotations, computeTcConfigHash)
+
+	// if tektonConfig changes tC.Spec.Config
+	if tektonConfigChanged {
```
pradeepitm12, line 305: How about changing `tektonConfigChanged`  to `nodeSelectorChanged` or something else.

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +283,74 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
 }
 
-func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targetNs, pruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference) error {
+func checkTektonConfigHashChange(annotations map[string]string, currentComputedHash string) bool {
+	// check if the annotaion in the cronJob is same or changed.
+	lastHash := annotations[tCconfigChangeHash]
+	return lastHash != currentComputedHash
+}
+
+func getPlatform(ctx context.Context, kc kubernetes.Interface, tC *v1alpha1.TektonConfig) bool {
```
pradeepitm12, line 321: @nikhil-thomas should this be good enough.

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +283,74 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
+		}
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTcConfigHash)
 	}
+
 	return nil
 }
 
-func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targetNs, pruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference) error {
+func checkTektonConfigHashChange(annotations map[string]string, currentComputedHash string) bool {
+	// check if the annotaion in the cronJob is same or changed.
```
pradeepitm12, line 316: nit : `annotaion` -> `annotation` :smile_cat:

pkg/reconciler/common/prune.go
rupalibehera: `computeTcConfigSpecConfigHash` how does this sound ? I know its little confusing but we are computing tekton's config spec.config's hash 
```
config:
    nodeSelector:
     node-role.kubernetes.io/master: ""
    tolerations:
    - key: node-role.kubernetes.io/master
      effect: NoSchedule
      operator: Exists
```

pkg/reconciler/common/prune.go
rupalibehera: Yes sure

pkg/reconciler/common/prune.go
rupalibehera: fixed

pkg/reconciler/common/prune.go
rupalibehera: changed

pkg/reconciler/common/prune.go
rupalibehera: done

pkg/reconciler/common/prune.go
rupalibehera: changed to nodeselectorConfigHash

pkg/reconciler/common/prune.go
piyush-garg: Why we want to do it only on openshift? we have the feature of providing the node selector and tolerations on config in both openshift and k8s afaik,

pkg/reconciler/common/prune.go
piyush-garg: Why do we want to do it only on openshift? we have the feature of providing the node selector and tolerations on config in both openshift and k8s afaik,

nikhil-thomas: @rupalibehera this pr https://github.com/tektoncd/operator/pull/710 [here](https://github.com/tektoncd/operator/pull/710/files#diff-93757650a499e93f88d69c44f58be32b522e5ec5f200aa5c77aee19f75b353d0R105)
adds a way to identify a platform in a simple way. 
so once that merges, any controller will be able to make a simple fucntion call and find the current platform.

nikhil-thomas: could you hold this pr until #710 merges

pkg/reconciler/common/prune.go
```diff
@@ -32,6 +27,10 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/logging"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
```
vdemeester, line 33: nit: those imports are misplaced, they should be on the "stdlib" group

pkg/reconciler/common/prune.go
rupalibehera: @piyush-garg , as per my understanding `tektonconfig.spec.config` can only contain nodeselector and tolerations @pradeepitm12 is my assumption correct ?

pkg/reconciler/common/prune.go
rupalibehera: @piyush-garg, earlier I thought we need it only for openshift, it does make sense to remove this check completely let me update the PR.

rupalibehera: > could you hold this pr until #710 merges

yeah, but I think so as @piyush-garg mentioned we can remove this check

pkg/reconciler/common/prune.go
rupalibehera: on save I guess the IDE puts it alphabetically, correcting it.

rupalibehera:
@piyush-garg @nikhil-thomas @pradeepitm12 could you guys please review this ?

rupalibehera:
@piyush-garg @nikhil-thomas @pradeepitm12 could you guys please review this again ?

pkg/reconciler/common/prune.go
piyush-garg: Compute function computes the hash of config struct irrespective of the fields inside that

rupalibehera: That is true, anyways I have changed it to calculate the hash of nodeSelector

pkg/reconciler/common/prune_test.go
```diff
@@ -324,3 +324,87 @@ func TestConfigChange(t *testing.T) {
...
+	}
+	client := fake.NewSimpleClientset(
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "openshift-pipelines"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "openshift-api"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "openshift-api-url"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "kube-system"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "kube-api"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "ns-one"}},
+	)
+	os.Setenv(JobsTKNImageName, "some")
```
chmouel, line 383: usually you would want to do defer function to unset the env (with https://pkg.go.dev/gotest.tools/v3/env for example)

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +284,46 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
+		}
 	}
+
 	return nil
 }
 
-func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targetNs, pruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference) error {
+func checkNodeSelectorHashChange(annotations map[string]string, currentComputedHash string) bool {
+	// check if the annotation in the cronJob is same or changed.
+	lastHash := annotations[nodeSelectorChangeHash]
```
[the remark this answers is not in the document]
rupalibehera: The annotation is set while creating the cronJob, this code path will be hit only after first cronjob is created, so annotation will always be there.

pkg/reconciler/common/prune.go
piyush-garg: Why we are depending on only nodeSelector, prune Jobs should change based on tolerations also

pkg/reconciler/common/prune.go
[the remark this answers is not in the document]
piyush-garg: there is nothing wrong in handling the case of no annotation, if you see the upgrade scenario, i think the annotation will not be there on old versions

pkg/reconciler/common/prune.go
```diff
@@ -35,22 +35,23 @@ import (
...
+	pruneSkip              = "operator.tekton.dev/prune.skip"
+	pruneResources         = "operator.tekton.dev/prune.resources"
+	pruneStrategy          = "operator.tekton.dev/prune.strategy"
+	pruneKeep              = "operator.tekton.dev/prune.keep"
+	pruneKeepSince         = "operator.tekton.dev/prune.keep-since"
+	pruneSchedule          = "operator.tekton.dev/prune.schedule"
+	pruneCronLabel         = "tektonconfig.operator.tekton.dev/pruner"
+	pruneCronNsLabel       = "tektonconfig.operator.tekton.dev/pruner.ns"
+	pruneLastAppliedHash   = "operator.tekton.dev/prune.hash"
+	nodeSelectorChangeHash = "operator.tekton.dev/nodeSelector.hash"
```
piyush-garg, line 50: we already have an annotation for this https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/const.go#L45, why to add a new one

rupalibehera: you mean to set the `LastAppliedHashKey` annotation on the cronJob, instead of creating a new one. 
I added a new annotaion to check if the nodeSelector changes and then to recreate the cronJob.

pkg/reconciler/common/prune.go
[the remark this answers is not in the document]
rupalibehera: ok, got it. did not think of the upgrade scenario.

nikhil-thomas:
/retest

as https://github.com/tektoncd/operator/pull/769 has merged, ci should pass

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +283,49 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
 
 	// any case with config change
 	if configChanged {
-		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef)
+		return createCronJob(ctx, pruner.kc, suffixedCronName, pruner.targetNamespace, uniquePruneNs, schedule, pruneContainers, pruner.ownerRef, tC, computeTektonConfigSpecConfigHash)
+	}
+
+	if len(cronList.Items) > 0 {
+		// get the cronjob to check previous annotations
+		cronjob := cronList.Items[0]
```
piyush-garg, line 301: why we are not doing the iteration over all cron jobs ?

pkg/reconciler/common/prune.go
```diff
@@ -279,22 +283,54 @@ func (pruner *Pruner) checkAndCreate(ctx context.Context, uniquePruneNs, schedul
...
+		}
+		return nil
+	}
+
 	return nil
 }
 
-func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targetNs, pruneNs, schedule string, pruneContainers []corev1.Container, oR v1.OwnerReference) error {
+func checkTektonConfigSpecConfigHashChange(annotations map[string]string, currentComputedHash string) bool {
+	if len(annotations) != 0 {
```
piyush-garg, line 324: Here we are checking the annotation length, but this does not guarantee whether this key value will exist or not

Is not better to do like

```
lastHash, ok := annotations[v1alpha1.LastAppliedHashKey]
if ok {
      return lastHash != currentComputedHash
}
return true
```

rupalibehera: ok will change it.

rupalibehera:
@pradeepitm12 @piyush-garg could you please take a look at this PR again

rupalibehera:
@pradeepitm12 @piyush-garg PTAL

PR #725, opened 2022-03-29
[cherry-pick][openshift] Use label selectors to find RBAC instalelr set name

nikhil-thomas:
Use labelselector based list query to find the existing RBAC installerset
name, instead of looking it up from a map in tektonconfig.status.

Kubernetes being "optimistically consistent", we cannot guarantee that
the value in tektonconfig.status will be updated within a predictable
time or that whether/when the update might pass or fail.

List queries based on label selectors is more reliable

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/openshift/tektonconfig/common.go
+++ b/pkg/reconciler/openshift/tektonconfig/common.go
@@ -18,17 +18,17 @@ package tektonconfig
 
 import (
 	"context"
-
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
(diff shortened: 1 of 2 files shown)


nikhil-thomas:
> Ideally, we would have a test for this :) /lgtm

the test for the mechanism used in this pr exists here: https://github.com/tektoncd/operator/blob/main/pkg/reconciler/kubernetes/tektoninstallerset/query_test.go

for the rbac package, yes need more tests. We will improve the coverage in the future. 🧑‍💻

PR #766, opened 2022-04-29
fixes fetch release script for pac

sm43:
it now uses branch to fetch the script instead of tags.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/hack/fetch-releases.sh
+++ b/hack/fetch-releases.sh
@@ -105,7 +105,11 @@ release_yaml_pac() {
     rm -rf ${dirPath} || true
     mkdir -p ${dirPath} || true
 
-    url="https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/${version}/release.yaml"
+    if [[ ${version} == "stable" ]]; then
+      url="https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/stable/release.yaml"
+    else
+      url="https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/release-${version}/release.yaml"
+    fi
 
     dest=${dirPath}/${fileName}.yaml
     http_response=$(curl -s -o ${dest} -w "%{http_code}" ${url})


hack/fetch-releases.sh
```diff
@@ -105,7 +105,11 @@ release_yaml_pac() {
     rm -rf ${dirPath} || true
     mkdir -p ${dirPath} || true
 
-    url="https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/${version}/release.yaml"
+    if [[ ${version} == "stable" ]]; then
+      url="https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/stable/release.yaml"
+    else
+      url="https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/release-${version}/release.yaml"
```
vdemeester, line 111: For a specific version, we could use the payload attached to the release pages, like
- https://github.com/openshift-pipelines/pipelines-as-code/releases/download/0.6.1/release.yaml

The would make more sense to me than using the branch.

vdemeester (commented):
@sm43 @chmouel I also see we have `release.yaml` and `release.k8s.yaml` ? What is the difference between the two ? Do they both work on vanilla k8s or not ?
(because depending on the answer, we might have to fetch a different one if we build for openshift or for kubernetes)

chmouel:
one has a Route for the controller one (k8) hasn't 🙃

hack/fetch-releases.sh
chmouel: either way i am fine :)

vdemeester:
> one has a Route for the controller one (k8) hasn't upside_down_face

@sm43 I guess, so far the "route" is added by the operator's code isn't it ? if yes, fetching `.k8s.yaml` would make the most sence.
This means however that this PR fails on upstream tests, because it fetches the one with `Route` that is not a k8s vanilla object 👼🏼

sm43:
> > one has a Route for the controller one (k8) hasn't upside_down_face
> 
> @sm43 I guess, so far the "route" is added by the operator's code isn't it ? if yes, fetching `.k8s.yaml` would make the most sence. This means however that this PR fails on upstream tests, because it fetches the one with `Route` that is not a k8s vanilla object 👼🏼

operator doesn't have any code which explicitly creates the route .. it will just create an installerset from the release yaml provided.


PAC is part of TektonAddon which is only enabled for openshift
for kubernetes/upstream test we don't fetch/or create PAC.
we don't run any test for addons in CI

vdemeester:
> > > one has a Route for the controller one (k8) hasn't upside_down_face
> > 
> > 
> > @sm43 I guess, so far the "route" is added by the operator's code isn't it ? if yes, fetching `.k8s.yaml` would make the most sence. This means however that this PR fails on upstream tests, because it fetches the one with `Route` that is not a k8s vanilla object 👼🏼
> 
> operator doesn't have any code which explicitly creates the route .. it will just create an installerset from the release yaml provided.
> 
> PAC is part of TektonAddon which is only enabled for openshift for kubernetes/upstream test we don't fetch/or create PAC. we don't run any test for addons in CI

Right, I see. Then it is probably fine for now, but 🙃

nikhil-thomas:
as https://github.com/tektoncd/operator/pull/769 has merged, ci should pass

PR #787, opened 2022-05-14
[openshift] [cherry-pick] Improve/rbac reconciling

nikhil-thomas:
# Changes

[OpenShift] Fix RBAC bugs

- Rename rolebinding created by RBAC reconciler to
  'openshift-pipelines-edit' (was 'edit' earlier)

- Add mechanism to ensure that missing RBAC resources are recreated if
  the RBAC installerSet is recreated during an upgrade.
- this ensures that RBAC in the version label in RBAC reconciled
  namespaces are removed during an upgrade and the presence of RBAC
  resources are verified

- Make TektonConfig reconciler listen to TektonInstallerSet events so
  that it can recreate RBAC InstallerSet if it is deleted manually from
  a cluster

- Add a mechanism to remove ownerReference and 'pipeline' sa subject
  from 'edit' rolebinding in namespaces
  - This ensures that operator upgrades won't delete/reset 'edit'
    rolebinding in usernamespaces.

Remove redundant 'UpdateStatus()' calls
    
Remove redundant UpdateStatus calls as UpdateStaus api call is carried
out by knative/pkg generated code at the end of each reconcile loop

[openshift] Use label selectors to find RBAC instalelr set name

Use labelselectr based list query to find the existing RBAC installerset
name, instead of looking it up from a map in tektonconfig.status.

Kubernetes being "optimistically consistent" we cannot guarantee that
the value in tektonconfig status will be updated withing a predictable
time or that whether/when the update will pass or fail.
    
List queries based on label selectors is more reliable

[openshift] Fix upgrade: Tektonconfig staying in Ready: False

Repalce metada.name with metadata.generateName in rbac installerSet creation

Add a mechanism to remove constant name based rbac installer set before an upgrade

[openshift] Fix rbac disable not working bug

Make the delete rbac installerSet logic use labelSelctor based `DeleteCollection` instead of
constant name based `Delete`.

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/kubernetes/tektondashboard/tektondashboard.go
+++ b/pkg/reconciler/kubernetes/tektondashboard/tektondashboard.go
@@ -297,16 +297,7 @@ func (r *Reconciler) updateTektonDashboardStatus(ctx context.Context, td *v1alph
 	// update the td with TektonInstallerSet and releaseVersion
 	td.Status.SetTektonInstallerSet(createdIs.Name)
 	td.Status.SetVersion(r.dashboardVersion)
-
-	// Update the status with TektonInstallerSet so that any new thread
-	// reconciling with know that TektonInstallerSet is created otherwise
-	// there will be 2 instance created if we don't update status here
-	if _, err := r.operatorClientSet.OperatorV1alpha1().TektonDashboards().
-		UpdateStatus(ctx, td, metav1.UpdateOptions{}); err != nil {
-		return err
-	}
-
-	return v1alpha1.RECONCILE_AGAIN_ERR
+	return nil
 }
 
 // transform mutates the passed manifest to one with common, component
--- a/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
+++ b/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
@@ -18,7 +18,6 @@ package tektonpipeline
 
 import (
 	"context"
-	stdError "errors"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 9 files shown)


pkg/reconciler/kubernetes/tektondashboard/tektondashboard.go
```diff
@@ -297,16 +297,7 @@ func (r *Reconciler) updateTektonDashboardStatus(ctx context.Context, td *v1alph
...
-
-	// Update the status with TektonInstallerSet so that any new thread
-	// reconciling with know that TektonInstallerSet is created otherwise
-	// there will be 2 instance created if we don't update status here
-	if _, err := r.operatorClientSet.OperatorV1alpha1().TektonDashboards().
-		UpdateStatus(ctx, td, metav1.UpdateOptions{}); err != nil {
-		return err
-	}
-
-	return v1alpha1.RECONCILE_AGAIN_ERR
```
savitaashture, line 309: @nikhil-thomas If i recall, this we don't require now because we are not storing any name in status as we are handling using labelselector right

nikhil-thomas: knative/pkg calls an updateStatus when ever we return out of reconcilers.

so explicit updateStatus call is not required in any of our reconcilers (or any reconciler written using knative/pkg).

in addition, explicit updateStatus call will make knative/pkg updateStatus call result in err as the revisionNumber will be changed.

👉 we don't use the name, but i haven't changed the writing the name to status part. The knative/pkg update will save the name change in status.

savitaashture: Yup got the reason its there in description :+1:

savitaashture (approved):
:tada:

PR #805, opened 2022-05-24
move prune commands to single container

pradeepitm12:
Issue : with older apporach pruner was crating one container for
        one namespace which says 100 container for 100 namespaces
Fix : We need to think of a proper appraoch to tackel this but for
      the time being moving all the prune commands to one container

# Changes

# Submitter Checklist

review them:

# Release Notes


pradeepitm12:
cc: @nikhil-thomas

nikhil-thomas:
/hold
as we need to resttucture the prune script.

at present all commands for the each namespaces gets appended into a single `;` separated string.
eg: 
```
echo ns1; echo ns2; echo ns3; ...
```

let us make it 
```
namespaces="ns1 ns2 ns3"; for ns in "${namespaces}"; do; echo ${ns}; done;

```

pradeepitm12:
need review please @nikhil-thomas :panda_face:

vdemeester (approved):

pkg/reconciler/common/prune.go
```diff
@@ -240,30 +240,32 @@ func prunableNamespaces(ctx context.Context, k kubernetes.Interface, defaultPrun
...
+
+	return cmds
 }
 
-func (pruner *Pruner) createJobContainers(nsConfig *pruneConfigPerNS, ns string) []corev1.Container {
-	var containers []corev1.Container
+func generatePruneCommand(nsConfig *pruneConfigPerNS, ns string) string {
+	cmd := pruneCommand(nsConfig, ns)
+	return cmd
+}
```
nikhil-thomas, line 256: do we need this function.
- either we can directly call `pruneCommand` directly where `generatePruneCommand` is called

or 

- we can rename `pruneCommand` as `generatePruneCommand`

pkg/reconciler/common/prune.go
```diff
@@ -240,30 +240,32 @@ func prunableNamespaces(ctx context.Context, k kubernetes.Interface, defaultPrun
...
+func getJobContainer(cmdArgs, ns string, tknImage string) []corev1.Container {
+	cmd := "function prune() {   n=$1; a=$2; resources=$3; old_ifs=\" \"; IFS=\",\"; for r in $resources; do tkn $r delete -n=$n $a -f; done; IFS=$old_ifs; }; echo $conf; for c in $*; do   ns=$(echo $c | cut -d \";\" -f 1); args=$(echo $c | cut -d \";\" -f 2); resources=$(echo $c | cut -d \";\" -f 3); prune $ns $args $resources; done;"
 	containerName := SimpleNameGenerator.RestrictLengthWithRandomSuffix("pruner-tkn-" + ns)
 	container := corev1.Container{
 		Name:                     containerName,
-		Image:                    pruner.tknImage,
+		Image:                    tknImage,
 		Command:                  []string{"/bin/sh", "-c"},
-		Args:                     []string{cmdArgs},
+		Args:                     []string{cmd, "-s", cmdArgs},
```
nikhil-thomas, line 265: let us make it
```golang
Command: []string{"/bin/sh", "-c", cmd}
Args:     []string{"-s", cmdArgs}
```

pkg/reconciler/common/prune.go
```diff
@@ -345,19 +347,24 @@ func createCronJob(ctx context.Context, kc kubernetes.Interface, cronName, targe
 }
 
 func pruneCommand(pru *pruneConfigPerNS, ns string) string {
```
nikhil-thomas, line 349: 1. rename this function to something like generatePruneConfig as it is not creating prune command instead it is just accumulating the configuration for each namespace.

2. this function needs a detailed comment(s) either at the start of the function or within. In comments mention a sample configuration string representing atleast 3 namespaces (output of this function).

I believe these changes are important to make it evident what this patch is improving.

pkg/reconciler/common/prune_test.go
```diff
@@ -210,40 +209,45 @@ func TestAnnotationCmd(t *testing.T) {
...
-		"one-" + scheduleCommon:        "tkn pipelinerun delete --keep=3 -n=ns-one -f ; ",
-		"two-" + scheduleUnique:        "tkn pipelinerun delete --keep=3 -n=ns-two -f ; ",
-		"four-" + scheduleCommon:       "tkn pipelinerun delete --keep=3 -n=ns-four -f ; ",
-		"seven-" + scheduleCommon:      "tkn pipelinerun delete --keep-since=3200 -n=ns-seven -f ; ",
-		"eight-" + scheduleCommon:      "tkn pipelinerun delete --keep=5 -n=ns-eight -f ; ",
-		"nine-" + scheduleCommon:       "tkn taskrun delete --keep=3 -n=ns-nine -f ; ",
-		"ten-" + scheduleCommon:        "tkn taskrun delete --keep=3 -n=ns-ten -f ; tkn pipelinerun delete --keep=3 -n=ns-ten -f ; ",
-		"ns-thirteen" + scheduleCommon: "tkn pipelinerun delete --keep=50 -n=ns-thirteen -f ; ",
-	}
+	commands := "function prune() {   n=$1; a=$2; resources=$3; old_ifs=\" \"; IFS=\",\"; for r in $resources; do tkn $r delete -n=$n $a -f; done; IFS=$old_ifs; }; echo $conf; for c in $*; do   ns=$(echo $c | cut -d \";\" -f 1); args=$(echo $c | cut -d \";\" -f 2); resources=$(echo $c | cut -d \";\" -f 3); prune $ns $args $resources; done;"
```
nikhil-thomas, line 212: why do we need this again in tests. 
if there is a valid reason then move it into a function instead of duplicating it here and [here](https://github.com/tektoncd/operator/pull/805/files#diff-44a5b2b09889314c6b8399cdf080f1ad3bc8e4b805d676ddd0bf671ca45b6efbR259)

nikhil-thomas (changes requested):

nikhil-thomas:
> /hold as we need to resttucture the prune script.
> 
> at present all commands for the each namespaces gets appended into a single `;` separated string. eg:
> 
> ```
> echo ns1; echo ns2; echo ns3; ...
> ```
> 
> let us make it
> 
> ```
> namespaces="ns1 ns2 ns3"; for ns in "${namespaces}"; do; echo ${ns}; done;
> ```
> 
> imagine the difference it will make for 1000+ namesapces with more than 1 command to be run per namespace. 🧑‍💻

the concern mentioned in this comment is resolved.

I have added some newer review comments. We can merge this once they are addressed.

pkg/reconciler/common/prune.go
pradeepitm12: I started with that, but earlier was getting file name too long, and `createContainerError`.
But now not getting that error. Updating it.

pkg/reconciler/common/prune.go
pradeepitm12: :+1:

pkg/reconciler/common/prune.go
pradeepitm12: :+1: Yeh We can get rid of it after these changes.

pkg/reconciler/common/prune_test.go
pradeepitm12: I wanted to check as I was getting createContainerError with the cmd and args, But now we don't need this part. Removing it.

PR #812, opened 2022-05-26
Install kn tasks only on architectures with knative client

yselkowitz:
# Changes

Which, currently, does not include ARM.  If/when that changes, it is suggested to simply add "arm64" instead of making these unconditional again (as was done in commit 4c80f347fadd2c4667aabe3412fbc120a1f692ca).

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/openshift/tektonaddon/pipelinetemplates/pipelinetemplates.go
+++ b/pkg/reconciler/openshift/tektonaddon/pipelinetemplates/pipelinetemplates.go
@@ -18,6 +18,7 @@ package tektonaddon
 
 import (
 	"path"
+	"runtime"
 
 	mf "github.com/manifestival/manifestival"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


yselkowitz:
tests have passed
/assign @concaf

nikhil-thomas:
@yselkowitz i see that this patch does two things:
1. disable installation of all kn, kn-apply ClusterTasks
2. skip the creation of +Add-flow Pipelines (default Pipelines) which use knative serving deployment as the deployment type 

@pradeepitm12 what do you think about the change in the pipeline generation part.


@yselkowitz  
1. could you confirm '2' is also an expected change.
1. could you add a release not mentioning both these changes.

pkg/reconciler/openshift/tektonaddon/pipelinetemplates/pipelinetemplates.go
```diff
@@ -73,7 +74,14 @@ func GeneratePipelineTemplates(templatePath string, manifest *mf.Manifest) error
 	workspacedTaskGenerators := []taskGenerator{
 		&pipeline{environment: "openshift", nameSuffix: "", generateDeployTask: openshiftDeployTask},
 		&pipeline{environment: "kubernetes", nameSuffix: "-deployment", generateDeployTask: kubernetesDeployTask},
-		&pipeline{environment: "knative", nameSuffix: "-knative", generateDeployTask: knativeDeployTask},
+	}
+
+	// install knative tasks only where knative is available
```
pradeepitm12, line 79: :+1:

yselkowitz:
>     1. could you confirm '2' is also an expected change.

Specifically, only on architectures which do not have OpenShift Serverless, namely `arm64`.  This matches what was done for `ppc64le` and `s390x` when they were in a similar state, and was then dropped by https://github.com/tektoncd/operator/commit/4c80f347fadd2c4667aabe3412fbc120a1f692ca once Serverless was available thereon.  Whenever Serverless does become available on ARM, instead of removing these conditionals again, I suggest that `"arm64"` simply be added to the list of knative architectures.

>     2. could you add a release not mentioning both these changes.

There should be no effect for existing users (which are on X/P/Z), this is only a prerequisite of the general ARM enablement.

nikhil-thomas:
> `390x` when they were in a similar state, and was then dropped by [4c80f34](https://github.com/tektoncd/operator/commit/4c80f347fadd2c4667aabe3412fbc120a1f692ca) once Serverless was available thereon. Whenever Serverless does become available on ARM, instead of removing these conditionals again, I suggest that `"arm64"` simply be added to the list of knative architectures

Thanks 👍 
/lgtm
/approve

PR #806, opened 2022-05-24
allow running buildah on newer nodes

ibotty:
# Changes
This will allow running buildah on nodes with kernel 5.12 or later.

Buildah requires the `SETFCAP` capability to work.  This pull requests updates  OpenShift SCC to allow adding this capability and the included `buildah` clustertask to request it.

I do not have access to clusters with older kernels so I don't know whether it will work on these nodes.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/operator/blob/master/CONTRIBUTING.md) for more details._

## Release Notes

```release-note
buildah ClusterTask is now allowed SETFCAP capability
```

PR #813, opened 2022-05-30
Add SETFCAP capability in pipelines-scc

piyush-garg:
This will add SETFCAP capability in pipelines-scc securitycontextconstraint
so that the pipeline sericeaccount which is getting used to
run buildah and s2i clustertask will not fail on ocp 4.11 as now
it will require this added capability to work properly

# Submitter Checklist

review them:

# Release Notes

```release-note
Add SETFCAP capability in pipelines-scc to make clustertask like buildah and pipeline work on OCP 4.11
```

### DIFF
--- a/cmd/openshift/operator/kodata/openshift/00-prereconcile/openshift-pipelines-scc.yaml
+++ b/cmd/openshift/operator/kodata/openshift/00-prereconcile/openshift-pipelines-scc.yaml
@@ -16,7 +16,8 @@ allowHostPorts: false
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: false
 allowedCapabilities: null
-defaultAddCapabilities: null
+defaultAddCapabilities:
+- SETFCAP
 fsGroup:
   type: MustRunAs
 groups:


ibotty:
This will by default add the capability even if the task do not need it.  Is that better than specifying it explicitly?

piyush-garg:
> This will by default add the capability even if the task do not need it. Is that better than specifying it explicitly?

I agree this is adding the capability by default, but the other solution which is requesting is not working for me in case of authenticated users, that is working only or kubeadmin

vdemeester:
@ibotty the "trick" we have today, at least for the openshift target, is that.. authenticated user are supposed to be able to run `buildah` with the `pipeline` SA, without anything to do *except* installing the operator 😅

ibotty:
I do concur that running buildah with the `pipeline` SA ought to be possible.  But adding the capability in the task works on my clusters also for non-kubeadmins as well.  Honestly, I don't care enough, but it leaves a sour taste in my mouth to add a capability that is not necessary.

piyush-garg:
> I do concur that running buildah with the `pipeline` SA ought to be possible. But adding the capability in the task works on my clusters also for non-kubeadmins as well. Honestly, I don't care enough, but it leaves a sour taste in my mouth to add a capability that is not necessary.

I agree with you here, but it was not working for me, even the taskrun pod was not getting created. I can try to debug more if anything extra needs to be done to make that work.

piyush-garg:
Closing this as #806 is working now

PR #819, opened 2022-06-01
Adds infrastructure for generalizing the creation of installerset and updates creation of TektonPipeline installerset

PuneetPunamiya:
# Changes

-  This patch adds the following for creation of installerset
    
     - Types for installerset
     - A default installerset with which we can add and retrieve metadata
     - A generic TektonInstallerSet Client
     - Installation logic to create the installerset for the component
     - Adds unit tests for installation and default installerset

---

- This patch also refactors the creation of TektonPipeline Installerset by
    
    - Creating a new default installerset
    - Adding metadata such as labels, annotations, manifest and ownerRef

# Submitter Checklist

review them:

# Release Notes


nikhil-thomas:
e2e timeout should get fixed after https://github.com/tektoncd/operator/pull/821 merges

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func newTisMetaWithName(name string) *tisMeta {
+	tis := newTisMeta()
+	tis.Name = name
+
+	return tis
+}
+
+func newTisMetaWithGenerateName(gname string) *tisMeta {
```
nikhil-thomas, line 53:
```suggestion
func newTisMetaWithGenerateName(namePrefix string) *tisMeta {
```

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,64 @@
...
+	tis.GenerateName = fmt.Sprintf("%s-", gname)
+
+	return tis
+}
+
+func (tis *tisMeta) config(ci ComponentInstaller) {
+	tis.Labels = ci.GetLabels(context.Background())
+	tis.Annotations = ci.GetAnnotations(context.Background())
+	tis.OwnerReferences = *ci.GetOwnerReferences(context.Background())
+}
```
nikhil-thomas, line 64:
```suggestion
func (tis *tisMeta) config(ctx context.Context, ci ComponentInstaller) {
	tis.Labels = ci.GetLabels(ctx)
	tis.Annotations = ci.GetAnnotations(ctx)
	tis.OwnerReferences = *ci.GetOwnerReferences(ctx)
}
```

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,64 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ComponentInstaller interface {
+	GetManifest(ctx context.Context) (*mf.Manifest, error)
+	GetLabels(ctx context.Context) map[string]string
+	GetAnnotations(ctx context.Context) map[string]string
+	GetOwnerReferences(ctx context.Context) *metav1.OwnerReference
+}
+
```
nikhil-thomas, line 33: add a comment with full name `tektoninstallerset meta`

pkg/reconciler/common/tektoninstallerset/install.go
```diff
@@ -0,0 +1,83 @@
...
+	return createInstallerSet(ctx, ci, newISM)
+}
+
+// Creates the installerset using generate name
+func CreateInstallerSetWithGenerateName(ctx context.Context, ci ComponentInstaller, namePrefix string) (*v1alpha1.TektonInstallerSet, error) {
+	newISM := newTisMetaWithGenerateName(namePrefix)
+
+	return createInstallerSet(ctx, ci, newISM)
+
+}
```
nikhil-thomas, line 41:
```suggestion
func CreateInstallerSetWithGenerateName(ctx context.Context, ci ComponentInstaller, namePrefix string) (*v1alpha1.TektonInstallerSet, error) {
	newISM := newTisMetaWithGenerateName(namePrefix)
	return createInstallerSet(ctx, ci, newISM)
}
```

nit

pkg/reconciler/common/tektoninstallerset/install.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Create the installerset
+func createInstallerSet(ctx context.Context, ci ComponentInstaller, tis *tisMeta) (*v1alpha1.TektonInstallerSet, error) {
+
+	client := getTektonInstallerSetClient()
+
+	return createInstallerSetWithClient(ctx, client, ci, tis)
+
+}
```
nikhil-thomas, line 50:
```suggestion
func createInstallerSet(ctx context.Context, ci ComponentInstaller, tis *tisMeta) (*v1alpha1.TektonInstallerSet, error) {

	client := getTektonInstallerSetClient()
	return createInstallerSetWithClient(ctx, client, ci, tis)
}
```
nit

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,64 @@
...
+	mf "github.com/manifestival/manifestival"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ComponentInstaller interface {
+	GetManifest(ctx context.Context) (*mf.Manifest, error)
+	GetLabels(ctx context.Context) map[string]string
+	GetAnnotations(ctx context.Context) map[string]string
+	GetOwnerReferences(ctx context.Context) *metav1.OwnerReference
```
nikhil-thomas, line 31:
```suggestion
	GetOwnerReferences(ctx context.Context) []metav1.OwnerReference
```
let us make this change right away. 
as the ownerReference field on k8s resources is an array anyway.

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type tisMeta struct {
+	Labels          map[string]string
+	Annotations     map[string]string
+	OwnerReferences metav1.OwnerReference
```
nikhil-thomas, line 37:
```suggestion
	OwnerReferences []metav1.OwnerReference
```

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,64 @@
...
+	tis := newTisMeta()
+	tis.GenerateName = fmt.Sprintf("%s-", gname)
+
+	return tis
+}
+
+func (tis *tisMeta) config(ci ComponentInstaller) {
+	tis.Labels = ci.GetLabels(context.Background())
+	tis.Annotations = ci.GetAnnotations(context.Background())
+	tis.OwnerReferences = *ci.GetOwnerReferences(context.Background())
```
nikhil-thomas, line 63: the `*` could be a code smell. can we match the type in the struct and the return type of the function.

in addition, let us make it an array as mentioned in comments above.

nikhil-thomas, line 63: can we do
```suggestion
	tis.OwnerReferences = ci.GetOwnerReferences(context.Background())
```

the `*` could be a code smell. can we match the type in the struct and the return type of the function.

in addition, let us make it an array as mentioned in comments above.

pkg/reconciler/common/tektoninstallerset/install_test.go
```diff
@@ -0,0 +1,110 @@
...
+			Annotations:     annotations,
+			OwnerReferences: []metav1.OwnerReference{{}},
+		},
+		Spec: v1alpha1.TektonInstallerSetSpec{
+			Manifests: manifest.Resources(),
+		},
+	}
+
+	if d := cmp.Diff(actual, expected); d != "" {
+		t.Errorf("Actual installerset is different from the expected one %s", diff.PrintWantGot(d))
```
nikhil-thomas, line 108: let us write the PrintWantGot in tektoncd/operator instead of vendoring it from [tektoncd/pipeline PrintWantGot](https://github.com/tektoncd/pipeline/blob/0d8b26a35c2705eba7da98c3c014b6e5e80417cd/test/diff/print.go#L25).

It is a tiny 1 line funtion. 


```plaintext
A little copying is better than a little dependency.
- Go Proverb
```
🤓

nikhil-thomas: or else we can directly use the `fmt.Sprintf("(-want, +got): %s", diff)` format string in this test.

pkg/reconciler/common/tektoninstallerset/install.go
```diff
@@ -0,0 +1,83 @@
...
+
+func makeInstallerSet(manifest *mf.Manifest, mt *tisMeta) *v1alpha1.TektonInstallerSet {
+	return &v1alpha1.TektonInstallerSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            mt.Name,
+			GenerateName:    mt.GenerateName,
+			Labels:          mt.Labels,
+			Annotations:     mt.Annotations,
+			OwnerReferences: []metav1.OwnerReference{mt.OwnerReferences},
```
nikhil-thomas, line 77: let us assume that there will be more than 1 ownerReference in mt.OwnerReferences (array).

so let us do either
```golang
OwnerReferences: mt.OwnerReferences
``` 
or 

```golang
OwnerReferences: append([]metav1.OwnerReference{}, mt.OwnerReferences...}
```

pkg/reconciler/common/tektoninstallerset/defaultinstallerset.go
```diff
@@ -0,0 +1,101 @@
...
+
+	mf "github.com/manifestival/manifestival"
+	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type DefaultInstaller struct {
+	Labels          map[string]string
+	Annotations     map[string]string
+	OwnerReferences metav1.OwnerReference
```
nikhil-thomas, line 30:
```suggestion
	OwnerReferences []metav1.OwnerReference
```
let us make this an array.

pkg/reconciler/common/tektoninstallerset/defaultinstallerset.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (di *DefaultInstaller) AddAnnotationsFromMap(annotations map[string]string) {
+	for k, v := range annotations {
+		di.AddAnnotationsKeyVal(k, v)
+	}
+}
+
+func (di *DefaultInstaller) AddOwnerReferences(ownerRef metav1.OwnerReference) {
+	di.OwnerReferences = ownerRef
```
nikhil-thomas, line 75: let us assume that we add ownerRef to an array so that the installerSet could have more than 1 ownerRef if necessary.

pkg/reconciler/common/tektoninstallerset/defaultinstallerset.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (di *DefaultInstaller) AddReleaseVersionLabel(val string) {
+	di.AddLabelKeyVal(v1alpha1.ReleaseVersionKey, val)
+}
+
+func NewDefaultInstaller() *DefaultInstaller {
+	return &DefaultInstaller{
+		Labels:          map[string]string{},
+		Annotations:     map[string]string{},
+		OwnerReferences: metav1.OwnerReference{},
```
nikhil-thomas, line 98:
```suggestion
		OwnerReferences: []metav1.OwnerReference{},
```
🤓

pkg/reconciler/common/tektoninstallerset/defaultinstallerset_test.go
```diff
@@ -0,0 +1,125 @@
...
+func TestGetLabels(t *testing.T) {
+	di := NewDefaultInstaller()
+
+	di.AddLabelsFromMap(map[string]string{
+		v1alpha1.CreatedByKey: "pipeline",
+	})
+
+	labels := di.GetLabels(context.Background())
+
+	assert.Equal(t, labels[v1alpha1.CreatedByKey], "pipeline")
```
nikhil-thomas, line 39: could we use the `cmp.diff` here also to compare the input map and the output map.

pkg/reconciler/common/tektoninstallerset/defaultinstallerset_test.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+		},
+	}
+
+	ownerRef := *metav1.NewControllerRef(component, component.GetGroupVersionKind())
+	di.AddOwnerReferences(ownerRef)
+
+	ownerReference := di.GetOwnerReferences(context.Background())
+
+	assert.Equal(t, ownerReference.Name, "test-name")
```
nikhil-thomas, line 73:
```suggestion
	if cmp.diff(ownerRef, ownerReference)
```
shall we do this

nikhil-thomas (changes requested):
@PuneetPunamiya This is a very valuable patch ❤️ . Thanks 👍 
Could you address the comments.

vdemeester (approved):

PuneetPunamiya:
> @PuneetPunamiya This is a very valuable patch heart . Thanks +1 Could you address the comments.

@nikhil-thomas Thanks for the review 🙂 I have addressed all the comments, could you please take a look at it once again

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
```diff
@@ -311,53 +312,45 @@ func (r *Reconciler) updateTektonPipelineStatus(ctx context.Context, tp *v1alpha
...
 		return nil, err
 	}
+	dis.AddAnnotationsKeyVal(v1alpha1.LastAppliedHashKey, specHash)
 
-	// create installer set
-	tis := r.makeInstallerSet(tp, manifest, specHash)
-	createdIs, err := r.operatorClientSet.OperatorV1alpha1().TektonInstallerSets().
-		Create(ctx, tis, metav1.CreateOptions{})
+	// Creates installer set with generate name
+	createdIs, err := tisBuilder.CreateInstallerSetWithGenerateName(ctx, dis, v1alpha1.PipelineResourceName)
```
vdemeester, line 347: So, I think I agree with both @nikhil-thomas and @sm43 but I tend to "not" use Builder, at least this way, and prefer "functional methods" instead. Let me try to explain this a bit more in detail.

```go
	// Adds the manifest to the installer
	dis.AddManifest(manifest)

	// Adds the labels to the installer
	dis.AddCreatedByLabel(createdByValue)
	dis.AddTypeLabel(v1alpha1.PipelineResourceName)
	dis.AddReleaseVersionLabel(r.operatorVersion)

	// Adds the ownerRef to the installer
	ownerRef := *metav1.NewControllerRef(tp, tp.GetGroupVersionKind())
	dis.AddOwnerReferences(ownerRef)

	// Adds the annotations to the installer
	dis.AddAnnotationsKeyVal(v1alpha1.TargetNamespaceKey, tp.Spec.TargetNamespace)

	// Hash of TektonPipeline Spec
	specHash, err := hash.Compute(tp.Spec)
	if err != nil {
		return nil, err
	}
	dis.AddAnnotationsKeyVal(v1alpha1.LastAppliedHashKey, specHash)

	// Creates installer set with generate name
	createdIs, err := tisBuilder.CreateInstallerSetWithGenerateName(ctx, dis, v1alpha1.PipelineResourceName)
```

We are replacing a methods with a struct + methods, with the goal to make it re-usable and parts of it shareable (add ownerref or not, AddTypeLabel or not, …). The general idea is that, you create the builder, you call some function, and you call `CreateInstaller…` ; and that function (or the underlying functions) can set sane default.

Although I think that **make sense**, I am not sure the `type` + method approach is the best. 
- it bring new structure just for the sake of it
- it is not super extensible, as to customize the behavior, we need to change the struct `DefaultInstaller`

A more functionnal approach would be something that takes other function, that we can call *transformers* that can mutate a struct, here `TektonInstallerSet` directly. That function can set sane default (aka create a *default* `InstallerSet` and then let the *transformers* do their job. It is more extensible because, the *transforemrs* can come from any place, the same package, or not, inlined functions or not. Anything that is *required* can be passed as an argument *prior* to the transformers

```go
// tisBuilder "github.com/tektoncd/operator/pkg/reconciler/common/tektoninstallerset"
is, err := tisBuilder.Create(ctx, v1alpha1.PipelineResourceName,  
	tisBuilder.WithManifest(manifest),
	tisBuilder.WithCreateByLabel(createdByValue),
	tisBuilder.WithOwnerReferences(*metav1.NewControllerRef(tp, tp.GetGroupVersionKind()),
	withMyVerySpecificThing,
	func(i *v1alpha1.TektonInstallerSet) *v1alpha1.TektonInstallerSet {
		// Inline, on-in-a-lifetime mutation function
	}
	// ...
))
```

vdemeester (commented):
@nikhil-thomas @PuneetPunamiya one "big" comment on the approach taken

pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go
nikhil-thomas: We all agree that with too many reconilers with more than 1 installerSets to be created we do need to make most of the code related to installerSet set creation, uptodate checks, garbage collection resusable and sharable.


The point of this patch is not the struct `DefaultInstaller` but the interface `ComponentInstaller`.

```golang
type ComponentInstaller interface {
	GetManifest(ctx context.Context) (*mf.Manifest, error)
	GetLabels(ctx context.Context) map[string]string
	GetAnnotations(ctx context.Context) map[string]string
	GetOwnerReferences(ctx context.Context) []metav1.OwnerReference
}
```

What you mentioned in the example in your comment is what this patch does.

instead of:

```

// tisBuilder "github.com/tektoncd/operator/pkg/reconciler/common/tektoninstallerset"
is, err := tisBuilder.Create(ctx, v1alpha1.PipelineResourceName,  
	tisBuilder.WithManifest(manifest),
	tisBuilder.WithCreateByLabel(createdByValue),
	tisBuilder.WithOwnerReferences(*metav1.NewControllerRef(tp, tp.GetGroupVersionKind()),
	withMyVerySpecificThing,
	func(i *v1alpha1.TektonInstallerSet) *v1alpha1.TektonInstallerSet {
		// Inline, on-in-a-lifetime mutation function
	}
	// ...
))
```

it does

```
// Creates the installerset using name
func CreateInstallerSetWithName(ctx context.Context, ci ComponentInstaller, name string) (*v1alpha1.TektonInstallerSet, error) {
	newISM := newTisMetaWithName(name)

	return createInstallerSet(ctx, ci, newISM)
}
```

nikhil-thomas: > I am not sure the type + method approach is the best.
>
> it bring new structure just for the sake of it

Now let us address the role of the struct `DefaultInstaller`.

Yes, we can implement the `ComponentInstaller` for each reconciler (~7times) or extent TektonComponent interface. I believe it will result in unnecessary duplication of code and make reconcilers too bloated.

What I observed is that most of the data and functions we use in all most all reconcilers are the same for creating their installerSet. So `DefaultInstaller` is a basic ready to use implementation of ComponentInstaller interface with some additional methods.

The reconcilers where things need to be customized could have their own implementation of ComponentInstaller interface. They can still use `CreateInstallerSetWithName(ctx context.Context, ci ComponentInstaller, name string)` or `CreateInstallerSetWithNamePrefix(ctx context.Context, ci ComponentInstaller, prefix string)`.

nikhil-thomas: I looks like most of concens are about how the package is being used in the reconiler. That is TektonPipeline Reconciler. 

That is something we can improve. What we have here now is simplest change to use the new pacakage retaining the function that is being called inside the current reconciler. (eg: [here](https://github.com/tektoncd/operator/blob/59ef30f9af50336ebe47b10c4fc21ef0dc513fbc/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L176) and [here](https://github.com/tektoncd/operator/blob/59ef30f9af50336ebe47b10c4fc21ef0dc513fbc/pkg/reconciler/kubernetes/tektonpipeline/tektonpipeline.go#L312))

Instead of the builder we can think about a constructor which can take manifests, labelvalues, annotations, ownerReferences etc as function parameters or as options struct params.

However, i think it is too early for those optimizations because, there are already some logic that needs fixing in how we use innstallerset in reconcilers.

for example: the hash annotation we set in installerset for freshness check is that of parent spec (eg: TektonPipeline) not that of installerSet spec. This is incorrect. As any change in parent spec will make all installerSet belonging to parent to be reconciled event when the change in parent is not relevant to an installerSet. Think having to recreate a deploymentPod because deployment strategy was updated in a deployment spec.

vdemeester: I don't really understand your first comment. You compare the function with the tranformers (that replace the function I commented on *and* all the above) with the content of the function ? If I understood correctly, `dis.AddManifest(manifest)` is gonna have to be called for each "reconciler" (`TektonPipeline` here, but all the othe ones) ? If yes, then you can't compared just the call to `CreateInstallerSetWithName` with `tisBuilder.Create(…)` with all the `With…` commands.

My point is that why defining a new struct (`DefaultInstallerSet`) instead of having a function that "define this default `Installerset` and takes transformers with it. I don't know the codebase enough, so my "example" is only based on what I see on the diff. But the general idea is the same. For example, on a very common reconciler, you wouldn't call `tisBuilder.Create(` with any `With`, it would be a very simple call with the only absolute required set of things, the rest would be infered / defaulted. 

Again, it's a implementation detail, and probably a "taste" of things. My personal opinion is that function with transformers on `struct` is usually a better approach than a *builder* struct (aka where the function are attached to a given type, that isn't the final type even).

Maybe we need to define even more clearer what this refactoring brings. From my understanding, we want to get rid of `makeInstallerSet` defined on all reconciler or at least "help" the code of each reconciler to create a `TektonInstallerSet` with the "simplest" possible way. From what I understand as well, the idea is that a long of things could be inferred (the `manifest`, the labels, ownerf, …) without really explicitely calling them. I think this is a good idea and I think I like the idea of the `ComponentInstaller` interface *but* I don't think we need an extract struct. For example, `manifest`, `createdBy`, the operator version, … almost *all* of them are either "global" variables of the reconciler or field on the reconciler. This means, the reconciler struct *itself* could implement most of the `ComponentInstaller` interface instead, and you could use the "functionnal" approach with the interface.

```
// tisBuilder "github.com/tektoncd/operator/pkg/reconciler/common/tektoninstallerset"
// A very simple, common one, no customization whatsoever
// It generates a InstallerSet from defaults value and the rencociler (that is seen as a `ComponentInstaller` from the builder)
is, err := tisBuilder.Create(ctx, r,)
// A more complex example
is, err := tisBuilder.Create(ctx, r,  
	// A custom owner refer for some reason
	tisBuilder.WithOwnerReferences(*metav1.NewControllerRef(tp, tp.GetGroupVersionKind()),
	// A very specific set of things
	withMyVerySpecificThing,
	func(i *v1alpha1.TektonInstallerSet) *v1alpha1.TektonInstallerSet {
		// Inline, on-in-a-lifetime mutation function
	}
	// ...
))
```

With an approach like this, you wouldn't need `DefaultInstaller` struct, only your "contract" and functionnal parameters on the existing struct `TektonInstallerSet`.

vdemeester: > The reconcilers where things need to be customized could have their own implementation of ComponentInstaller interface. They can still use `CreateInstallerSetWithName(ctx context.Context, ci ComponentInstaller, name string)` or `CreateInstallerSetWithNamePrefix(ctx context.Context, ci ComponentInstaller, prefix string)`.

This is what I want to "not have" to be honest. Why redefining a new implementation of a struct, that happen to just "mirror" information the reconciler already has. The reconciler can satisfy the required "interface", and for additional behavior that are very specific to its installerset, it can define and call (or reuse) a function that mutate the `TektonInstallerSet` directly.

nikhil-thomas: @vdemeester @sm43 
Questions:
1. Do we all agree that CreteInstallerSet function (from the package) should take an interface as its argument.
2. Are we on the same page on most of the reconcilers having smilar functional requirements when comes to creation of installersets.
3. DefaultInstaller struct is a reference implementation of ComponentInstallerInterface. Cutomizability comes from implementing ComponentInstaller interface not by modifying DefaultInstaller. In addtion, DefaultInstaller could be embedded in other structs and only the functions that need customization could be reimplemented. What do you guy think about this.

vdemeester: > I looks like most of concens are about how the package is being used in the reconiler. That is TektonPipeline Reconciler.

Not the package but the struct that (`DefaultINstaller`) and the `Create…` function that, in my opinion, bring a bit too much code for what it brings (as highlighted above).

vdemeester: > @vdemeester @sm43 Questions:
> 
> 1. Do we all agree that CreteInstallerSet function (from the package) should take an interface as its argument.
> 2. Are we on the same page on most of the reconcilers having smilar functional requirements when comes to creation of installersets.
> 3. DefaultInstaller struct is a reference implementation of ComponentInstallerInterface. Cutomizability comes from implementing ComponentInstaller interface not by modifying DefaultInstaller. In addtion, DefaultInstaller could be embedded in other structs and only the functions that need customization could be reimplemented. What do you guy think about this.

I agree with 1. and 2., but I disagree with 3. `DefaultInstaller` is not needed in my opinion. I see `ComponentInstaller` as the *way* the `CreteInstallerSet` function can get **require** value from anything that satisfy it (could be the reconciler themselves). The *rest* of the customization, can happen outside `ComponentInstaller`. For me `DefaultInstaller` **and** `tisMeta` struct are not needed at all to satisfy 1. and 2.

nikhil-thomas: > DefaultInstaller is not needed in my opinion. I see ComponentInstaller as the way the CreteInstallerSet function can get required value from anything that satisfy it (could be the reconciler themselves)

this is what we alos thought in the begining. 

but when we reached the `from anything that satisfy it (could be the reconciler themselves)` part, we saw the need for a defaultImplementation.

nikhil-thomas: oh, i see the `With` part is a variadic `With` function type.

```golang
is, err := tisBuilder.Create(ctx, v1alpha1.PipelineResourceName,  
	tisBuilder.WithManifest(manifest),
	tisBuilder.WithCreateByLabel(createdByValue),
	tisBuilder.WithOwnerReferences(*metav1.NewControllerRef(tp, tp.GetGroupVersionKind()),
	withMyVerySpecificThing,
	func(i *v1alpha1.TektonInstallerSet) *v1alpha1.TektonInstallerSet {
		// Inline, on-in-a-lifetime mutation function
	}
	// ...
))
```

that would be better. 🧑‍💻 👍

pkg/reconciler/common/tektoninstallerset/types.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"fmt"
+
+	mf "github.com/manifestival/manifestival"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ComponentInstaller interface {
+	GetManifest(ctx context.Context) (*mf.Manifest, error)
```
vdemeester, line 28: Any reason for passing `context.Context` on those ? They seem to be never used.

nikhil-thomas: in the defaultImplementation no. However it is an extension point for the future. So that in the future ctx could be used in other implementations without having to change the signature of methods in this interface.

nikhil-thomas: for example: platfrom contextKey from ctx.

vdemeester: Ok 👌🏼 (for me, no need for now, so I would just make the interface evolve when we need it, but that's a nit 😝 )

nikhil-thomas:
@vdemeester what do you think about 
```golang
type WithFunc func(*v1alpha1.TektonInstallerset) error

func CreateInstallerSet(m mf.Manifest, withFuncs ...WithFunc) *v1alpha1.TektonInstallerSet
```

no other interfaces or structs. we can provide implement some of the common "withFunc"(s) in the package.


From reconcilers the invocation will be like,
```golang
tispkg.CreateInstallerSet(manifest, 
    WithName("abcd"), // or a Mutually exclusive WithGenerateName("prefix")
    WithLabels(map[string]string, // also WithTypeLabel("label")
    WithAnntations(),
    WithOwnerReference(), ...
```

PR #827, opened 2022-06-07
Checks and updates ownerRef of Hub-db pvc if Hub instance is already created

PuneetPunamiya:
- If Hub instance is already created, then the data in the Hub db should
be preserved

- Hence, this patch checks the ownerRef of hub-db pvc and updates it
to `TektonHub` so while upgrade the pvc is not deleted and the data
is preserved

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/kubernetes/tektonhub/tektonhub.go
+++ b/pkg/reconciler/kubernetes/tektonhub/tektonhub.go
@@ -137,6 +137,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, th *v1alpha1.TektonHub) 
 	}
 	th.Status.MarkPreReconcilerComplete()
 
+	// TODO: remove this after operator openshift-build version 1.8
+	if err := r.checkDbApiPVCOwnerRef(ctx, th); err != nil {
+		return err
+	}
+
 	// Manage DB
 	if err := r.manageDbComponent(ctx, th, hubDir, version); err != nil {
 		return r.handleError(err, th)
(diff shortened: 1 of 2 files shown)


pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -224,6 +224,25 @@ func (r *Reconciler) manageApiComponent(ctx context.Context, th *v1alpha1.Tekton
...
+	if pvc != nil {
+		if !r.checkPVCOwnerRef(pvc, th) {
+			ownerRef := *metav1.NewControllerRef(th, th.GroupVersionKind())
+			pvc.SetOwnerReferences([]metav1.OwnerReference{ownerRef})
+
+			_, err := r.kubeClientSet.CoreV1().PersistentVolumeClaims(th.Spec.GetTargetNamespace()).Update(ctx, pvc, metav1.UpdateOptions{})
+			if err != nil {
+				return err
+			}
+		}
```
nikhil-thomas, line 243: why don't we move this also into the `checkPVCOwnerRef(pvc, th)` function.

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -496,6 +533,24 @@ func (r *Reconciler) transform(ctx context.Context, manifest mf.Manifest, th *v1
...
+}
+
+func (r *Reconciler) checkPVCOwnerRef(pvc *corev1.PersistentVolumeClaim, th *v1alpha1.TektonHub) bool {
+	if len(pvc.GetOwnerReferences()) == 1 {
+		if pvc.GetOwnerReferences()[0].Kind == th.Kind {
+			return true
+		}
+	}
+	return false
+}
```
nikhil-thomas, line 552: add a to be deprecated comment arround this function and each place it is called, as we will not need it in future releases.

something like

```golang
// TODO: remove this after operator <x.y.z>
```

nikhil-thomas:
@PuneetPunamiya do we need this change on main branch. Even if we decide to keep it, let us make sure that we remove it after few releases.

could you create 2 issues:

1. hub e2e tests in operator (hub is the only component without e2e tests now) [refer](https://github.com/tektoncd/operator/tree/main/test/e2e)
2. an issue to track the removal of changes made by this pr.

nikhil-thomas:
/approve
✅ lgtm

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
PuneetPunamiya: yeah we can move that, but then function name should be changed as well, because if we are changing the ownerRef of pvc then on the cluster we are updating the pvc and I guess `checkPVCOwnerRef` is just checking the if it has the `TektonHub` as ownerRef

PR #850, opened 2022-06-21
Add graceful termination to webhook pods

nikhil-thomas:
This patch adds a mechanism to remove webhook defintions created by the webhook pod when it is
being terminated. This eases 2 situations:

1. Operaor CRD instance CRDs can be deleted manually even after webhook pods are removed (eg: OLM uninstalls operator without removing components)
2. When operator in uninstalled the webhook definitions are never removed. Therefore, if the operator is ever installed in a different namespaces there is a timewindow in which the service namespaces are updated in the webhook definition during which all call to webhook fails.

**Note:-** We create mutatingWebhookConfigurations and validatingWebhookConfigurations explicitly using
Kubeapi calls from webhook pod. We do that be cause OLM (operator lifecycle manager) doesnot work well with knative/pkg based webhooks.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
fixes explicit deletion of instances of operator's CRDs (eg: TektonConfig, TektonAddons...) being blocked if the operator is uninstalled before removing the components
```

### DIFF
--- a/cmd/kubernetes/webhook/main.go
+++ b/cmd/kubernetes/webhook/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"os"
 
 	"github.com/tektoncd/operator/pkg/webhook"
@@ -49,6 +50,8 @@ func main() {
 	webhook.CreateWebhookResources(ctx)
 	webhook.SetTypes("kubernetes")
 
+	go gracefulTermination(ctx)
+
 	sharedmain.WebhookMainWithConfig(ctx, serviceName,
 		cfg,
 		certificates.NewController,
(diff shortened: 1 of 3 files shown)


vdemeester (approved):
/meow

nikhil-thomas:
/hold

this works when i scale deployment down to 0 using `kubectl scale deployment` command. But it doesn't work when i delete the deployment using `kubectl delete deployment` command. 

the signals defined in knative/pkg are [here](https://github.com/knative/pkg/blob/7d607d643ee2081f93c3de1ab3731fd5f9a76c1e/signals/signal_posix.go#L27:5)

in this patch, a go routine waits on `<-ctx.Done()` 
the definition of this `Done()` is [this](https://github.com/knative/pkg/blob/7d607d643ee2081f93c3de1ab3731fd5f9a76c1e/signals/signal.go#L63)

the stop channel is set up like [this](https://github.com/knative/pkg/blob/7d607d643ee2081f93c3de1ab3731fd5f9a76c1e/signals/signal.go#L36)

nikhil-thomas:
/hold cancel
the patch works fine. 

What i mentioned in the above comment are delays in reconciliation in installerset reconciler.

PR #863, opened 2022-06-27
Fix chains secret getting overwritten on upgrade

piyush-garg:
This will fix chains' secret getting overwritten over upgrade

Now the secret has been moved out to a separate installerset and
this installerset will not be removed and changed over upgrade.

This installlerset will only be deleted if the namespace
of the tektonchains gets changed.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix chains secret getting overwritten on upgrade
```

### DIFF
--- a/pkg/reconciler/kubernetes/tektonchain/tektonchain.go
+++ b/pkg/reconciler/kubernetes/tektonchain/tektonchain.go
@@ -57,13 +57,22 @@ type Reconciler struct {
 var _ tektonchainreconciler.Interface = (*Reconciler)(nil)
 var _ tektonchainreconciler.Finalizer = (*Reconciler)(nil)
 
-const createdByValue = "TektonChain"
+const (
+	createdByValue          = "TektonChain"
+	secretChainInstallerset = "chain-secret"
+)
 
 var (
 	ls = metav1.LabelSelector{
 		MatchLabels: map[string]string{
 			v1alpha1.CreatedByKey:     createdByValue,
 			v1alpha1.InstallerSetType: v1alpha1.ChainResourceName,
+		},
+	}
+	secretLs = metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			v1alpha1.CreatedByKey:     createdByValue,
+			v1alpha1.InstallerSetType: secretChainInstallerset,
 		},
 	}
 )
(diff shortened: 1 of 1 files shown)


pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -233,6 +244,64 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonChain
...
+	if err != nil {
+		return err
+	}
+	if existingSecretInstallerSet == "" {
+		tc.Status.MarkInstallerSetNotAvailable("Chain Secret InstallerSet not available")
+		createdIs, err := r.createSecretInstallerSet(ctx, tc)
+		if err != nil {
+			return err
+		}
+		return r.updateTektonChainStatus(tc, createdIs)
```
sm43, line 262: lgtm 
here we are overriding installer set name in status of TektonChains
there is only one field for installer set name in CR and now we have 2 for chains
may be in future we could remove that field or add all installerset names in map

piyush-garg: Yes, I also was worrying the same, then checked the pipeline reconciler where we have three installerset like pre, post, and main. There was also the same thing so I kept it as it is for now.

If you want i can remove this logic for now and in the future can do the changes for array/map.

sm43: yeah if we can keep the core installer set in status for now would be okay I think
and skip the secret one

vdemeester (approved):

concaf:
/hold
testing this locally

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
piyush-garg: updated the PR with this

sm43:
LGTM

piyush-garg:
In this scenario, I did not choose the approach of applying secret directly without installerset and keeping the ChainsCR as owner, because in the case of chain CR is edited to install in a different namespace, we need to handle explicitly the case of secret deletion etc. So to achieve that, created a secret in a separate installerset and this installerset will be deleted only on the change of namespace for now else not, and removing chains will also delete the secret.

concaf (approved):
/approve

PR #641, opened 2022-03-01
cmd/fetch-payload: a go command to fetch payload

vdemeester:
# Changes

This is meant to replace `./hack/fetch-releases.sh` in a more flexible
way. In a gist, you can use different components yaml and fetch from
them depending on the need.

On additional goal is to be able to support fetch from PRs and/or
local folders.

```bash
$ go run ./cmd/fetch-payload -filename components.yaml -target foo/ -platform openshift
skip: component tekton-dasbhoard on platform openshift
Download https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml to foo/cmd/openshift/kodata/tekton-pipeline/0.0.0-latest/00-release.yaml
Download https://raw.githubusercontent.com/openshift-pipelines/pipelines-as-code/release-0.5.3/release-0.5.3.yaml to foo/cmd/openshift/kodata/tekton-addon/pipelines-as-code/0.5.3/00-release.yaml
skip: component tekton-results on platform openshift
Download https://storage.googleapis.com/tekton-releases/chains/latest/release.yaml to foo/cmd/openshift/kodata/tekton-chains/0.0.0-latest/00-release.yaml
Download https://storage.googleapis.com/tekton-releases/triggers/latest/interceptors.yaml to foo/cmd/openshift/kodata/tekton-trigger/0.0.0-latest/01-interceptors.yaml
Download https://storage.googleapis.com/tekton-releases/triggers/latest/release.yaml to foo/cmd/openshift/kodata/tekton-trigger/0.0.0-latest/00-release.yaml

$ go run ./cmd/fetch-payload -filename components.yaml -target foo/
Download https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml to foo/cmd/kubernetes/kodata/tekton-pipeline/0.0.0-latest/00-release.yaml
Download https://storage.googleapis.com/tekton-releases/triggers/latest/interceptors.yaml to foo/cmd/kubernetes/kodata/tekton-trigger/0.0.0-latest/01-interceptors.yaml
Download https://storage.googleapis.com/tekton-releases/chains/latest/release.yaml to foo/cmd/kubernetes/kodata/tekton-chains/0.0.0-latest/00-release.yaml
Download https://storage.googleapis.com/tekton-releases/triggers/latest/release.yaml to foo/cmd/kubernetes/kodata/tekton-trigger/0.0.0-latest/00-release.yaml
[...]

PR #898, opened 2022-07-07
Components file as source of truth + tooling

vdemeester:
# Changes

This brings a new configuration file, on the root, to serve as the
source of truth for component version for the operator. Any script
should use this file as a source of truth.

In addition, this bring the start of a tool to quickly get the version
of each component (example of usage in `./test/config.sh`) as well as
checking if there is new version of the component (`check` and
`check-bugfix`).

The next step here is to enhance this tool to be able to update the
`components.yaml` on its own (depending if we want to bump to the next
bugfix release or the next minor/major one). From there, we will be
able to automate the updates of those, a bit à-la-dependabot.

This takes over https://github.com/tektoncd/operator/pull/641 in a simpler way as it doesn't touch anything related to fetching the payload.

/kind misc

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```


components.yaml
```diff
@@ -0,0 +1,26 @@
...
+  version: v0.9.0
+results:
+  github: tektoncd/results
+  version: v0.4.0
+hub:
+  github: tektoncd/hub
+  version: v1.8.0
+pipelines-as-code:
+  github: openshift-pipelines/pipelines-as-code
+  version: 0.10.1
```
vdemeester, line 26: Note: this doesn't support `stable` or `latest` or `nightly` 🙃 The idea behind it is really to do something similar to what `dependabot` does: we always use released version, and an automated tooling is looking at new releases and propose PRs to bump to those.

sm43:
this is for figuring out the latest version right?
and fetch release script will still be used to fetch the relase yamls?
don't we need to similar changes in `makefile` as did in `test/config.sh`?

vdemeester:
> this is for figuring out the latest version right? and fetch release script will still be used to fetch the relase yamls? don't we need to similar changes in `makefile` as did in `test/config.sh`?

@sm43 Yes, this is to "figure out the latest version" (with some rules) and to update the version in case of. And yes, the fetch release scripts will still be used. Yes we need, I forgot there was something in the `Makefile` 😝

vdemeester:
Note that fetch-script will be updated long term to read the `components.yaml` instead of relying on environment variables.

vdemeester:
@sm43 well, I ended up adapting `fetch-release.sh` to read the components file.

vdemeester:
https://github.com/tektoncd/plumbing/pull/1134 to build a "nightly"

concaf (commented):
This is awesome! 🎉 
Can we also make CSV template use the components.yaml as well?

cmd/operator-tool/check.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+	g, ctx := errgroup.WithContext(context.Background())
+	for name, component := range components {
+		// Force scope
+		name := name
+		component := component
+
+		g.Go(func() error {
```
pradeepitm12, line 39: :+1:

PR #921, opened 2022-07-25
Updates the hub api secret by setting the AUTH_BASE_URL from Hub status

PuneetPunamiya:
Currently if user creates the Hub api secret, then Auth url was not set,
and hence Hub login was failing

Hence, this patch updates the value of AUTH_BASE_URL from the hub api
secret from the Hub status when user creates the secret

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/openshift/tektonhub/extension.go
+++ b/pkg/reconciler/openshift/tektonhub/extension.go
@@ -404,6 +404,20 @@ func (oe openshiftExtension) SetAuthBaseURL(ctx context.Context, th *v1alpha1.Te
 			return err
 		}
 		th.Status.SetAuthRoute(fmt.Sprintf("https://%s", authRoute))
+
+		if secret.Data == nil || string(secret.Data["AUTH_BASE_URL"]) != th.Status.AuthRouteUrl {
+
+			if secret.StringData == nil {
+				secret.StringData = make(map[string]string)
+			}
+
+			secret.StringData["AUTH_BASE_URL"] = th.Status.AuthRouteUrl
+
+			_, err = oe.kubeClientSet.CoreV1().Secrets(th.Spec.GetTargetNamespace()).Update(ctx, secret, metav1.UpdateOptions{})
+			if err != nil {
+				return err
+			}
+		}
 	} else {
 		th.Status.SetAuthRoute("")
 	}


pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -404,6 +404,14 @@ func (oe openshiftExtension) SetAuthBaseURL(ctx context.Context, th *v1alpha1.Te
 			return err
 		}
 		th.Status.SetAuthRoute(fmt.Sprintf("https://%s", authRoute))
+
+		secret.StringData = make(map[string]string)
```
sm43, line 408: if one of auth is configured for eg. GitHub this code will always be executed right?
 even if the data is already updated in the secret? 
can we check if url is not set then update or else move on?

PuneetPunamiya: so you mean check the value of the url, if it is empty then only update the secret ?

sm43: yes if url is already set in the secret we can skip the update call

PuneetPunamiya: yeah sounds good 👍🏻

PuneetPunamiya: will update the  patch

PuneetPunamiya: updated the patch!

pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -404,6 +404,16 @@ func (oe openshiftExtension) SetAuthBaseURL(ctx context.Context, th *v1alpha1.Te
 			return err
 		}
 		th.Status.SetAuthRoute(fmt.Sprintf("https://%s", authRoute))
+
+		if string(secret.Data["AUTH_BASE_URL"]) == "" {
```
sm43, line 408:
```suggestion
		if string(secret.Data["AUTH_BASE_URL"]) != th.Status.AuthRouteUrl {
```

sm43 (changes requested):

pkg/reconciler/openshift/tektonhub/extension.go
sm43, line 408: will secret.Data[""] break if the map is empty?

sm43 (changes requested):

pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -404,6 +404,16 @@ func (oe openshiftExtension) SetAuthBaseURL(ctx context.Context, th *v1alpha1.Te
 			return err
 		}
 		th.Status.SetAuthRoute(fmt.Sprintf("https://%s", authRoute))
+
+		if string(secret.Data["AUTH_BASE_URL"]) != th.Status.AuthRouteUrl {
+			secret.StringData = make(map[string]string)
```
piyush-garg, line 409: why are we initializing it, we should do it if it is nill not every time

also if we are initializing it now, then if will give Nil Pinter Exception I think

PuneetPunamiya: have updated the patch,please take a look at it once

pkg/reconciler/openshift/tektonhub/extension.go
```diff
@@ -404,6 +404,20 @@ func (oe openshiftExtension) SetAuthBaseURL(ctx context.Context, th *v1alpha1.Te
 			return err
 		}
 		th.Status.SetAuthRoute(fmt.Sprintf("https://%s", authRoute))
+
+		if string(secret.Data["AUTH_BASE_URL"]) != th.Status.AuthRouteUrl {
```
piyush-garg, line 408: we want to put it in data or stringData?

PuneetPunamiya: So with the current patch we have put in stringData but while updating/creating the secret, it gets encoded and added in the data

PuneetPunamiya: so should we put directly in data ?
i.e. instead of `secret.StringData["AUTH_BASE_URL"] = th.Status.AuthRouteUrl` it will be `secret.Data["AUTH_BASE_URL"] = []byte(th.Status.AuthRouteUrl)`

PR #953, opened 2022-08-02
fix secret detach from SA

pradeepitm12:
issue:
  operator creates SA which is used to run pr/tr, where user could
  attach a secret.
  All the SA create by operator have ownerRef of InsterllerSet
  any time there is a version update all the SA get re-creted and
  any of external added secretes to SA get removed.

fix:
  we have removed the ownerRef from the SAs created, which do not
  re-creates the SA with an upgrade.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/openshift/tektonconfig/extension.go
+++ b/pkg/reconciler/openshift/tektonconfig/extension.go
@@ -70,7 +70,7 @@ func (oe openshiftExtension) PreReconcile(ctx context.Context, tc v1alpha1.Tekto
 
 	// below code helps to retain state of pre-existing SA at the time of upgrade
 	if existingSAWithOwnerRef(r.tektonConfig) {
-		if err := removeOwnerRefFromPreExistingSA(ctx, r.kubeClientSet); err != nil {
+		if err := changeOwnerRefOfPreExistingSA(ctx, r.kubeClientSet, *config); err != nil {
 			return err
 		}
 		tcLabels := config.GetLabels()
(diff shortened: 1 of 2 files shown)


pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -132,3 +145,29 @@ func (oe openshiftExtension) Finalize(ctx context.Context, comp v1alpha1.TektonC
...
 }
+
+func removeOwnerRefFromPreExistingSA(ctx context.Context, kc kubernetes.Interface) error {
+	allSAs, err := kc.CoreV1().ServiceAccounts("").List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	for _, sa := range allSAs.Items {
+		if sa.Name == "pipeline" && !nsRegex.MatchString(sa.Namespace) {
+			sa.SetOwnerReferences([]metav1.OwnerReference{})
```
sm43, line 156: should we replace it with TektonConfig ref 🤔 
so they will be deleted when TC is deleted?
WDYT

pradeepitm12: This makes sense to me, its anyway created by operator and should be deleted with it.

pradeepitm12: PR Updated..

vdemeester (approved):

PR #1122, opened 2022-10-27
Adds short names to Operator CRDs

sm43:

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/config/base/300-operator_v1alpha1_config_crd.yaml
+++ b/config/base/300-operator_v1alpha1_config_crd.yaml
@@ -26,6 +26,8 @@ spec:
     listKind: TektonConfigList
     plural: tektonconfigs
     singular: tektonconfig
+    shortNames:
+    - tc
   preserveUnknownFields: false
   scope: Cluster
   versions:
--- a/config/base/300-operator_v1alpha1_hub_crd.yaml
+++ b/config/base/300-operator_v1alpha1_hub_crd.yaml
@@ -26,6 +26,8 @@ spec:
     listKind: TektonHubList
     plural: tektonhubs
     singular: tektonhub
+    shortNames:
+    - th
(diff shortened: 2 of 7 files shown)


piyush-garg:
Any reason to no to do for chain and result ?

vdemeester:
any reason to do that ? `tc`, `tp` are not really.. conveying any meaning 😅

sm43:
@piyush-garg couldn't think of shortnames for them. tc take for tektonconfig and tr is for taskrun i guess, didn't wanted to conflict.

@vdemeester those help while development a bit 🙃 even if they doesn't help much for the end user

sm43:
@vdemeester do you think we shouldn't add this? 😄 
I could make my life easier with some bash .. so I am too strong adding this 
but still this would be helpful 😝

vdemeester:
> @vdemeester do you think we shouldn't add this? smile I could make my life easier with some bash .. so I am too strong adding this but still this would be helpful stuck_out_tongue_closed_eyes

I don't really have strong opinions. I tend to "dislike" most shortnames because they can conflict with each others and they tend to be so short you don't really know what it's about (e.g. `tp` could mean *a lot of thing*).

They are somewhat useful, only when interactively with `kubectl`, and when you don't have completion setup. In bash script, it's usually better to use long names (and long command line) to be more descriptive and clearer when reading.

But in any case, I am fine getting this in, I am just not a fan of it 😝.

/approve

piyush-garg:
i am also not sure if we should merge this if it not giving any value to end users and also inconsistent as some of the CRDs does not have short names (chains and results)

sm43:
well not useful to end user and the short names could be conflicting
so I will make my life easier with bash for now 😄  
/close

PR #1123, opened 2022-10-27
fix deletion of resource which is breaking the pod

sm43:
manifestival is breaking while deleting a resource: during deletion it first fetch the resource, if there occurs any error it still tries to access the resources which is causing
```
panic: runtime error: invalid memory address or nil pointer dereferenc
```
https://github.com/tektoncd/operator/blob/main/vendor/github.com/manifestival/manifestival/manifestival.go#L175 here when error is not nil, current is nil and then further it tries to access the resource kind which breaks the pod.

so we write our own delete func, which first fetch the resource, if not error occurs then calls delete.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
@@ -52,13 +52,15 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, installerSet *v1alpha1.Te
 		return err
 	}
 
-	// Delete all resources except CRDs and Namespace as they are own by owner of
-	// TektonInstallerSet
-	// They will be deleted when the component CR is deleted
-	deleteManifests = deleteManifests.Filter(mf.Not(mf.Any(namespacePred, mf.CRDs, pvcPred)))
-	err = deleteManifests.Delete()
+	// Delete all resources except CRDs, Namespace & PVC as they are own by owner of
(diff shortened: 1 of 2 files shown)


pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
```diff
@@ -52,13 +52,15 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, installerSet *v1alpha1.Te
...
 	}
 
-	// Delete all resources except CRDs and Namespace as they are own by owner of
-	// TektonInstallerSet
-	// They will be deleted when the component CR is deleted
-	deleteManifests = deleteManifests.Filter(mf.Not(mf.Any(namespacePred, mf.CRDs, pvcPred)))
-	err = deleteManifests.Delete()
+	// Delete all resources except CRDs, Namespace & PVC as they are own by owner of
+	// TektonInstallerSet. They will be deleted when the component CR is deleted
+	installer := installer{
```
piyush-garg, line 57: why we need the installer struct, cant we just do based on manifest?

sm43: yeah its just breaking down code in a func.
so `installer` creates resource so deletes resource too.

vdemeester (approved):

PR #1147, opened 2022-11-08
Exposes resolvers configurations

sm43:
# Changes
this allows user to enable/disable resolvers through TektonConfig and provide
configurations for each of the resolver.
to enable/disable resolvers, following are the fields
```
    enable-bundles-resolver: true
    enable-cluster-resolver: true
    enable-git-resolver: true
    enable-hub-resolver: true
```
to provide configurations for individual providers, following fields are defined
which are map[string]string
and after defining **only valid fields will be copied and invalid fields will be ignored.**
NOTE: we currently **_don't_** **validate** or **default** configurations for individual resolvers
```
     bundles-resolver-config:
        default-service-account: abc
      cluster-resolver-config:
        default-namespace: test
      git-resolver-config:
        server-url: localhost.com
      hub-resolver-config:
        default-tekton-hub-catalog: cat
```

the Tektonconfig would look like below
```
kind: TektonConfig
metadata:
  name: config
spec:
  pipeline:
    bundles-resolver-config:
      default-service-account: abc
    cluster-resolver-config:
      default-namespace: test
    disable-creds-init: false
    embedded-status: full
    enable-api-fields: stable
    enable-bundles-resolver: true
    enable-cluster-resolver: true
    enable-custom-tasks: false
    enable-git-resolver: true
    enable-hub-resolver: true
    enable-tekton-oci-bundles: false
    git-resolver-config:
      server-url: localhost.com
    hub-resolver-config:
      default-tekton-hub-catalog: cat
    metrics.pipelinerun.duration-type: histogram
    metrics.pipelinerun.level: pipeline
    metrics.taskrun.duration-type: histogram
    metrics.taskrun.level: task
    require-git-ssh-secret-known-hosts: false
    running-in-environment-with-injected-sidecars: true
    send-cloudevents-for-runs: false
```

# Submitter Checklist

review them:

# Release Notes


savitaashture (approved):

test/e2e/kubernetes/tektonresultdeployment_test.go
```diff
@@ -37,6 +37,7 @@ import (
 
 // TestTektonResultDeployment verifies the TektonResult creation, deployment recreation, and TektonResult deletion.
 func TestTektonResultDeployment(t *testing.T) {
+	t.Skip()
```
sm43, line 40: https://github.com/tektoncd/operator/issues/1148
need to check in results

vdemeester (commented):
I wonder if we could have done something like:

```
resolvers:
  git:
    enable: true
    server-url: https://…
  hub:
    enable: true
    …
```

But I think we can think about this for a v1beta1 version of the operator CRD or something.

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
```diff
@@ -80,6 +80,20 @@ func (p *Pipeline) setDefaults() {
 		p.MetricsTaskrunLevel = config.DefaultTaskrunLevel
 	}
 
+	// Resolvers
```
piyush-garg, line 83: are we enabling by default?

sm43: yes

vdemeester: yes starting in 0.41 they are enabled by default.

sm43:
> One suggestion
> 
> instead of
> 
> ```
>     bundles-resolver-config:
>       default-service-account: abc
>     enable-bundles-resolver: true
> ```
> 
> we should do
> 
> ```
>     bundles-resolver:
>         enable: true
>         default-service-account: abc
> ```
> 
> or something similar, I feel the config of the resolver should be nested there only, not in a separate field



> 

@piyush-garg 
yeah it make sense but it wouldn't work with existing transformers directly 😄 
should we try that for v1beta1 api?

piyush-garg:
> > One suggestion
> > instead of
> > ```
> >     bundles-resolver-config:
> >       default-service-account: abc
> >     enable-bundles-resolver: true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > we should do
> > ```
> >     bundles-resolver:
> >         enable: true
> >         default-service-account: abc
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or something similar, I feel the config of the resolver should be nested there only, not in a separate field
> 
> > 
> 
> @piyush-garg yeah it make sense but it wouldn't work with existing transformers directly smile should we try that for v1beta1 api?

okay, lets target for v1beta1 then

/lgtm

PR #1149, opened 2022-11-08
Update pipeline-scc to not allow privilege escalation

vdemeester:
# Changes

This SCC shouldn't allow privilege escalation at all.

# Submitter Checklist

review them:

# Release Notes

```release-note
On Openshift, the pipelines-scc doesn't allow privilege escalation anymore
```

### DIFF
--- a/cmd/openshift/operator/kodata/openshift/00-prereconcile/openshift-pipelines-scc.yaml
+++ b/cmd/openshift/operator/kodata/openshift/00-prereconcile/openshift-pipelines-scc.yaml
@@ -13,7 +13,7 @@ allowHostIPC: false
 allowHostNetwork: false
 allowHostPID: false
 allowHostPorts: false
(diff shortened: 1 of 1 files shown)


adambkaplan:
@vdemeester would this be a breaking change for users? Are there any Red Hat-supported tasks which require this in their pod security context?

vdemeester:
> @vdemeester would this be a breaking change for users? Are there any Red Hat-supported tasks which require this in their pod security context?

So as far as I tried, I do not see this breaking users. I don't think it is going to be a breaking change. In what we ship, there is no need for this at least, and all the example I tried were working fine.

acornett21:
@vdemeester This does in fact break users. Any user that is launching `podman` or similar tool in a task, this breaks. Can this change be rolled back and a new version of the OpenShift Operator released? Let me know if there are any other questions. Thanks.

PR #1195, opened 2022-11-30
Update ClusterInterceptor ClusterRoleBinding after the namespace removal

savitaashture:
# Changes

**Issue:** openshift-pipelines-clusterinterceptors ClusterRoleBinding not updated after namespace removal
**Fixes:** Operator rbac will keep on watching on the namespaces which are there in `Subjects` of  openshift-pipelines-clusterinterceptors ClusterRoleBinding and whenever namespace get deleted rbac will update  ClusterRoleBinding by removing the namespace 

**Before:**
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2022-11-30T10:57:38Z"
  name: openshift-pipelines-clusterinterceptors
  ownerReferences:
  - apiVersion: operator.tekton.dev/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: TektonInstallerSet
    name: rhosp-rbac-z4v56
    uid: f591469b-125d-43c1-ba9a-7f1c95ef98d7
  resourceVersion: "151362"
  uid: 291c40fa-1390-4e90-a391-4e8856ca5a3f
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: openshift-pipelines-clusterinterceptors
subjects:
- kind: ServiceAccount
  name: pipeline
  namespace: default
- kind: ServiceAccount
  name: pipeline
  namespace: openshift
- kind: ServiceAccount
  name: pipeline
  namespace: test
```

**After:**
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: "2022-11-30T10:57:38Z"
  name: openshift-pipelines-clusterinterceptors
  ownerReferences:
  - apiVersion: operator.tekton.dev/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: TektonInstallerSet
    name: rhosp-rbac-z4v56
    uid: f591469b-125d-43c1-ba9a-7f1c95ef98d7
  resourceVersion: "151522"
  uid: 291c40fa-1390-4e90-a391-4e8856ca5a3f
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: openshift-pipelines-clusterinterceptors
subjects:
- kind: ServiceAccount
  name: pipeline
  namespace: default
- kind: ServiceAccount
  name: pipeline
  namespace: openshift

```
Signed-off-by: Savita Ashture sashture@redhat.com

# Submitter Checklist

review them:

# Release Notes


sm43:
> From a "performance" perspective, I am a bit worried. Looking at the code, we would execute this every second (right ?). And this would generate a relatively decent amount of API calls (because we are using the clients) : 1 for the `ClusterRoleBinding` and then one for each namespaces referenced in the `ClusterRoleBinding`.
> 
> I wonder if there would be better alternative(s):
> 
>     * We may want to make use of the _global resync tick_ or something to not execute this every second
> 
>     * We may want to use informers instead of client to use cached result instead of query the API too much
> 
>     * We may want to have a specific reconciler (or something) that watch namespaces and reconcile the `ClusterRoleBinding` with it
> 
> 
> cc @sm43 @concaf

one way could be moving `ensureClusterRoleBindings` out of the for loop
so we know which namespace we want to have in clusterrolebinding 
so we check if they exist at once
we could use hash to detect change... hashing the ns array

sm43:
> From a "performance" perspective, I am a bit worried. Looking at the code, we would execute this every second (right ?). And this would generate a relatively decent amount of API calls (because we are using the clients) : 1 for the `ClusterRoleBinding` and then one for each namespaces referenced in the `ClusterRoleBinding`.
> 
> I wonder if there would be better alternative(s):
> 
>     * We may want to make use of the _global resync tick_ or something to not execute this every second
> 
>     * We may want to use informers instead of client to use cached result instead of query the API too much
> 
>     * We may want to have a specific reconciler (or something) that watch namespaces and reconcile the `ClusterRoleBinding` with it
> 
> 
> cc @sm43 @concaf


well about the optimization, yeah we could do that too, using listers to fetch and client only to update
having a seperate reconciler would be my proposal
as change in ns trigger TC reconciler which does a lot other than rbac

savitaashture:
> From a "performance" perspective, I am a bit worried. Looking at the code, we would execute this every second (right ?). And this would generate a relatively decent amount of API calls (because we are using the clients) : 1 for the `ClusterRoleBinding` and then one for each namespaces referenced in the `ClusterRoleBinding`.
> 
> I wonder if there would be better alternative(s):
> 
> * We may want to make use of the _global resync tick_ or something to not execute this every second
> * We may want to use informers instead of client to use cached result instead of query the API too much
> * We may want to have a specific reconciler (or something) that watch namespaces and reconcile the `ClusterRoleBinding` with it
> 
> cc @sm43 @concaf

* We may want to use informers instead of client to use cached result instead of query the API too much
Can try this because it avoids too many API calls

savitaashture:
@vdemeester @sm43 
Updated PR
PTAL
Thank you

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -598,13 +605,67 @@ func (r *rbac) ensureClusterRoleBindings(ctx context.Context, sa *corev1.Service
 		return r.updateClusterRoleBinding(ctx, viewCRB, sa)
 	}
 
+	go func() {
```
vdemeester, line 608: Isn't this at risk of running multiple times ? (and thus create multiple infinite loop — that could be a leak)

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -598,13 +605,67 @@ func (r *rbac) ensureClusterRoleBindings(ctx context.Context, sa *corev1.Service
...
+			if _, ok := nsMap[rb.Subjects[i].Namespace]; !ok {
+				rb.Subjects = removeIndex(rb.Subjects, i)
+				if _, err := rbacClient.ClusterRoleBindings().Update(ctx, rb, metav1.UpdateOptions{}); err != nil {
+					logger.Error(err, "failed to update "+clusterInterceptors+" crb")
+					return err
+				}
+				logger.Infof("successfully removed namespace and updated %s ", clusterInterceptors)
+			}
+		}
+	}
```
vdemeester, line 661: We could also optimize this to just to one update call (remove all from the array and then set it once)

savitaashture: Updated

pkg/reconciler/openshift/tektonconfig/rbac.go
savitaashture: running this in goroutine because delete events are not coming 
:thinking:

sm43: wdym delete events are not coming? if a namespace is deleted reconciler is not getting triggered?

savitaashture: Yes it was my mistake 
Actually events are coming but i added my changes under namespaces range and because of this condition https://github.com/tektoncd/operator/blob/main/pkg/reconciler/openshift/tektonconfig/rbac.go#L191-L193 from second iteration the call is not going further

I have updated now and its working as expected

PTAL

vdemeester (approved):

PR #1389, opened 2023-03-18
Added PipelineRun artifacts to Chains spec

sabre1041:
# Changes

Added `artifacts.pipelinerun.*` to `Chains` spec

# Submitter Checklist

review them:

# Release Notes

```release-note
Added PipelineRun artifacts to Chains spec
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonchain_validation.go
+++ b/pkg/apis/operator/v1alpha1/tektonchain_validation.go
@@ -68,6 +68,28 @@ func (tcs *TektonChainSpec) ValidateChainConfig(path string) (errs *apis.FieldEr
 		}
 	}
 
+	if tcs.ArtifactsPipelineRunFormat != "" {
+		if tcs.ArtifactsPipelineRunFormat != "in-toto" {
+			errs = errs.Also(apis.ErrInvalidValue(tcs.ArtifactsPipelineRunFormat, path+".artifacts.pipelinerun.format"))
+		}
+	}
+
+	if tcs.ArtifactsPipelineRunStorage != "" {
+		input := strings.Split(tcs.ArtifactsPipelineRunStorage, ",")
+		for i, v := range input {
+			input[i] = strings.TrimSpace(v)
+			if !allowedArtifactsStorage.Has(input[i]) {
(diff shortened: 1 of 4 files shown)


PuneetPunamiya:
Thank you for the pr 🙂 
Could you also please update the following fields in the docs https://github.com/tektoncd/operator/blob/main/docs/TektonChain.md

piyush-garg:
@sabre1041 Can you please update the docs and also run codegen script in `hack` directory once.

piyush-garg:
Also if there are any value constraints for these fields, you can add the validation here too https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonchain_validation.go#L48

sabre1041:
@piyush-garg Updated based on your feedback

PuneetPunamiya:
Thanks, could you also please add test for the validation as well

pkg/apis/operator/v1alpha1/tektonchain_validation.go
```diff
@@ -68,6 +68,28 @@ func (tcs *TektonChainSpec) ValidateChainConfig(path string) (errs *apis.FieldEr
 		}
 	}
 
+	if tcs.ArtifactsPipelineRunFormat != "" {
+		if tcs.ArtifactsPipelineRunFormat != "in-toto" {
```
piyush-garg, line 72: the `slsa/v1` is also allowed i guess

sabre1041: Its allowed for backwards compatibility. should it also be included?

PuneetPunamiya:
@sabre1041 sorry one more last comment, can you please add validation test as well for the fields you have added 🙂

pkg/apis/operator/v1alpha1/tektonchain_validation.go
PuneetPunamiya: We have a follow up for it which addresses the newly added values and some more new fields

piyush-garg: @sabre1041 it is like that because that was the only supported value when these config were added

PR #1434, opened 2023-04-10
Change the value on upgrade for embedded status

piyush-garg:
As now we have changed the default value of embedded status to both so during upgrade we should change value to both as some functionalities are not working on v1 resources because some fields have been removed in them

This make sures that we change value during minor version upgrade, not all

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/shared/tektonconfig/tektonconfig.go
@@ -19,6 +19,7 @@ package tektonconfig
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	mf "github.com/manifestival/manifestival"
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -231,6 +232,12 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonConfig)
 		tc.Status.MarkPostInstallFailed(v1alpha1.UpgradePending)
 		tc.Status.MarkNotReady("Upgrade Pending")
 	}
+
+	// TODO: remove later, as it is for handling specific case of one upgrade
+	if ok && !strings.HasPrefix(ver, getPatchVersionTrimmed(r.operatorVersion)) {
+		tc.Spec.Pipeline.EmbeddedStatus = "both"
```
jkandasa, line 238: later time user may change the value, but I guess this logic will reset the `tc.Spec.Pipeline.EmbeddedStatus` to `both`, in the future release. example: `0.69 != 0.70`

jkandasa:
while ago I was asked upgrade code to manage the `TectonConfig` CR changes.
But we are adding upgrade code in the actual logic. 
We may consider to include upgrade logic separately in the future.  
We can manage upgrade changes to a specific version and will be out of actual logic.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
piyush-garg: This a lag which we introduced in 0.62 and changing the value in 0.65, and this will be removed in 0.67 i guess

piyush-garg: Its a short term thing, we need to do otherwise the v1 resources are not working as expected after upgrade

jkandasa (approved):
I am fine, if we create GH issue and track it for the release `0.67` to remove the changes.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
piyush-garg: may be we can add check like if the value is `full` the only do change, otherwise not

concaf (approved):
/lgtm

PR #1445, opened 2023-04-12
Set "disable-ha=true" because HA is not enabled.

yocum137:
HA tekton is not yet enabled, no need to enable it, now.  

The default has been incorrectly set to "false," i.e., it's _enabling_ HA. This is impacting customer pipelines.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/vendor/knative.dev/pkg/injection/sharedmain/main.go
+++ b/vendor/knative.dev/pkg/injection/sharedmain/main.go
@@ -195,7 +195,7 @@ func MainWithContext(ctx context.Context, component string, ctors ...injection.C
 	}
 
 	// TODO(mattmoor): Remove this once HA is stable.
-	disableHighAvailability := flag.Bool("disable-ha", false,
+	disableHighAvailability := flag.Bool("disable-ha", true,
 		"Whether to disable high-availability functionality for this component.  This flag will be deprecated "+
 			"and removed when we have promoted this feature to stable, so do not pass it without filing an "+
 			"issue upstream!")


vdemeester (changes requested):
Hey @yocum137, thanks for the PR. but :
- This is on a vendor dependency, we do not edit directly vendor code (as they would be 
overwritten next time we update it). 
- This would only affect the tektoncd/operator controller and none of the operand (so the pipeline operator, trigger operator, …)

The required change is, more likely, to update the deployment of the tektoncd/operator controller (and webhook) as well as having a *transformer* to updates this on the operand's controllers.

I am still a bit fuzzy in how it impacts customer pipelines. What usually impacts customer pipelines is the number of PipelineRun (and other objects) that are on the cluster at a given time, *and* some performance flags on the controller. The *leader-election* thing is usually not adding any overhead.

khrm (commented):
/ok-to-test

khrm:
Tests should fail because we don't modify vendor code as Vincent said. Just checking whether it happens here like in the case of pipelines or triggers.

khrm:
We have disable-ha property but we don't expose it in the operator. 

https://github.com/tektoncd/pipeline/blob/bd7a02f4c7785cc22faa7f260c1e2380126b6d28/cmd/controller/main.go#L62

But this setting isn't recommended, instead, just run a single replica of the controller.

jkandasa:
@yocum137 `disable-ha` support added in https://github.com/tektoncd/operator/pull/1458

PR #1534, opened 2023-06-01
Exposes `MONGO_SERVER_URL` as env in Tekton Chain

PuneetPunamiya:
- This patch exposes `MONGO_SERVER_URL` env variable that will be added to the chains controller deployment

- The field can be added as
   ```yaml
    chain:
       controllerEnvs:
       - name: MONGO_SERVER_URL      # This is the only field supported at the moment and is optional and  and when added by user, it is added as env to Chains controller
         value: #value               # This can be provided same as env field 
      ....
   ```
- fixes - [SRVKP-3112](https://issues.redhat.com/browse/SRVKP-3112)

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
New field added under tektonChains spec, called `controllerEnvs` with this, we can supply `MONGO_SERVER_URL` as a env variable that will be added to the chains controller deployment. Please refer TektonChains documentation for the detailed information. 
```


pkg/apis/operator/v1alpha1/tektonchain_types.go
```diff
@@ -62,8 +62,15 @@ type TektonChainSpec struct {
 	Config Config `json:"config,omitempty"`
 }
 
-// Chain defines the field to provide chain configuration
 type Chain struct {
+	ChainProperties `json:",inline"`
+	// The params to customize different components of Pipelines
+	// +optional
+	Params []Param `json:"params,omitempty"`
```
jkandasa, line 69: If possible add a test to verify `params` field

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +513,36 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+		}
+
+		for _, p := range params {
+			if p.Name == "MONGO_SERVER_URL" {
+				env := corev1.EnvVar{
+					Name:  p.Name,
+					Value: p.Value,
+				}
+
+				d.Spec.Template.Spec.Containers[0].Env = append(d.Spec.Template.Spec.Containers[0].Env, env)
```
jkandasa, line 536: It is not a good idea to go with container index. go with container name

pkg/apis/operator/v1alpha1/tektonchain_types.go
PuneetPunamiya: Since there are no restricted values and need to access the deployment to see the env, hence I think it is fine if we not add a test for this

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +513,40 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+		}
+
+		d := &appsv1.Deployment{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, d)
+		if err != nil {
+			return err
+		}
+
+		for _, p := range params {
+			if p.Name == "MONGO_SERVER_URL" {
```
piyush-garg, line 530: why we are hardcoding this ?

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -391,8 +394,9 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
 		common.InjectOperandNameLabelOverwriteExisting(v1alpha1.OperandTektoncdChains),
 		common.DeploymentImages(chainImages),
 		common.AddConfiguration(instance.Spec.Config),
-		common.AddConfigMapValues(ChainsConfig, instance.Spec.Chain),
+		common.AddConfigMapValues(ChainsConfig, instance.Spec.Chain.ChainProperties),
 		common.AddDeploymentRestrictedPSA(),
+		AddMongoServerUrl(instance.Spec.Chain.Params),
```
piyush-garg, line 399: what to do with other key value pairs of param field?

concaf: i agree, there are no other fields in param field right now - which should continue being the case
i feel that instead of a new params field, this can go in as a separate field that gets exposed as the env var inside the chains controller pod

concaf: something like
```
spec:
  mongoServerUrl: xxx://xxx:xxx
  artifacts.taskrun.format: in-toto
  artifacts.taskrun.storage: tekton,oci (comma separated values)
  artifacts.taskrun.signer: x509

```

jkandasa: mongo server url might have password/token. Is it ok to keep it in TektonConfig CR?

piyush-garg: instead of this, should we add field like `envs` or `controllerEnvs` which is an array of key value pairs. whatever specified here will be added as env to chains controller.

adding a field which is getting added to controller or other field are getting added to configmap, this is not consistent, and give an wrong impression/UX

pkg/apis/operator/v1alpha1/tektonchain_types.go
```diff
@@ -62,8 +62,13 @@ type TektonChainSpec struct {
 	Config Config `json:"config,omitempty"`
 }
 
-// Chain defines the field to provide chain configuration
 type Chain struct {
+	ChainProperties `json:",inline"`
+	ControllerEnvs  []ControllerEnv `json:"controllerEnvs,omitempty"`
```
jkandasa, line 67: IMHO, we can use `[]corev1.EnvVar` in the place of `[]ControllerEnv`,
With this we can supply the value from secret too. also supports plan `key`, `value` approach.
[EnvVar Doc](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#envvar-v1-core)
```
impoert "corev1 "k8s.io/api/core/v1"

ControllerEnvs []corev1.EnvVar `json:"controllerEnvs,omitempty"`

```

an example to pass a secret via `controllerEnvs`,
```
controllerEnvs:
  - name: MONGO_SERVER_URL
    valueFrom:
      secretKeyRef:
        name: custom_secret_name
        key: mongo_server_url
```

pkg/apis/operator/v1alpha1/tektonchain_validation.go
```diff
@@ -48,7 +48,18 @@ func (tc *TektonChain) Validate(ctx context.Context) (errs *apis.FieldError) {
...
 	}
 
-	return errs.Also(tc.Spec.ValidateChainConfig("spec"))
+	return errs.Also(tc.Spec.ValidateControllerEnv(), tc.Spec.ValidateChainConfig("spec"))
+}
+
+func (tcs *TektonChainSpec) ValidateControllerEnv() (errs *apis.FieldError) {
+	if tcs.ControllerEnvs != nil {
+		for _, v := range tcs.ControllerEnvs {
+			if v.Name != "MONGO_SERVER_URL" {
```
jkandasa, line 57: Keep a list of supported Envs some where and use that. do not hard code here.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +513,41 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+		if controllerEnvs != nil {
+			for _, v := range controllerEnvs {
+				env := corev1.EnvVar{
+					Name:  v.Name,
+					Value: v.Value,
+				}
+
+				for i, c := range d.Spec.Template.Spec.Containers {
+					if c.Name == "tekton-chains-controller" {
+						d.Spec.Template.Spec.Containers[i].Env = append(d.Spec.Template.Spec.Containers[i].Env, env)
```
jkandasa, line 538: If a existing env found with the same name, replace that with new env.
otherwise append the new env.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +513,41 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+
+		if controllerEnvs != nil {
+			for _, v := range controllerEnvs {
+				env := corev1.EnvVar{
+					Name:  v.Name,
+					Value: v.Value,
+				}
+
+				for i, c := range d.Spec.Template.Spec.Containers {
+					if c.Name == "tekton-chains-controller" {
```
jkandasa, line 537: IMHO, keep the name of container as constant in top of the file.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +513,41 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
 	}
 	return v1alpha1.RECONCILE_AGAIN_ERR
 }
+
+func AddControllerEnv(controllerEnvs []v1alpha1.ControllerEnv) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "Deployment" {
```
jkandasa, line 519: also include  `len(controllerEnvs) == 0`, so we can skip immediately, if there is no elements found in `controllerEnvs`
```
if u.GetKind() != "Deployment"  || len(controllerEnvs) == 0 {
```

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +513,41 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+			return nil
+		}
+
+		d := &appsv1.Deployment{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, d)
+		if err != nil {
+			return err
+		}
+
+		if controllerEnvs != nil {
```
jkandasa, line 529: with the above command we can remove this `if` block.
and If I remember correctly, `range` is `nil` safe.

pkg/apis/operator/v1alpha1/tektonchain_types.go
PuneetPunamiya: Yeah makes sense, but not I'm confused, what to add in the docs

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
PuneetPunamiya: I guess deployment should give an error if key with same name are added 🤔

pkg/apis/operator/v1alpha1/tektonchain_types.go
piyush-garg: yup we should take the struct from containers, it will be beneficial to not hardcode in CR

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
piyush-garg: but we dont want to error right. we will error in validation if two variable are defined with same key.

here we want to supply the one which is there on CR and not fail the installation

docs/TektonConfig.md
```diff
@@ -178,6 +178,55 @@ pipeline:
...
+
+Refer to [properties](https://github.com/tektoncd/operator/blob/main/docs/TektonChain.md#chain-config) section in TektonChain for available options
+
+Example:
+
+```yaml
+chain:
+  targetNamespace: tekton-pipelines
+  controllerEnvs:
+    - name: MONGO_SERVER_URL      # This field is optional and when added by user, it is added as env to Chains controller
```
piyush-garg, line 194: we need to mention that this is the only key supported for the moment

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
piyush-garg: also this will make the installation error if value is updated in CR

docs/TektonConfig.md
PuneetPunamiya: yes 👍🏻

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
PuneetPunamiya: Yeah got it 👍🏻

docs/TektonChain.md
```diff
@@ -69,6 +69,9 @@ metadata:
   name: chain
 spec:
   targetNamespace: tekton-pipelines
+  controllerEnvs:
+    - name: MONGO_SERVER_URL      # This is the only field supported at the moment and is optional and  and when added by user, it is added as env to Chains controller
```
jkandasa, line 73: repeated `and`

docs/TektonConfig.md
```diff
@@ -178,6 +178,55 @@ pipeline:
...
+
+Refer to [properties](https://github.com/tektoncd/operator/blob/main/docs/TektonChain.md#chain-config) section in TektonChain for available options
+
+Example:
+
+```yaml
+chain:
+  targetNamespace: tekton-pipelines
+  controllerEnvs:
+    - name: MONGO_SERVER_URL      # This is the only field supported at the moment and is optional and  and when added by user, it is added as env to Chains controller
```
jkandasa, line 194: repeated `and`

pkg/apis/operator/v1alpha1/tektonchain_validation.go
```diff
@@ -48,7 +49,18 @@ func (tc *TektonChain) Validate(ctx context.Context) (errs *apis.FieldError) {
...
 
-	return errs.Also(tc.Spec.ValidateChainConfig("spec"))
+	return errs.Also(tc.Spec.ValidateControllerEnv(), tc.Spec.ValidateChainConfig("spec"))
+}
+
+func (tcs *TektonChainSpec) ValidateControllerEnv() (errs *apis.FieldError) {
+	if tcs.ControllerEnvs != nil {
+		for _, v := range tcs.ControllerEnvs {
+			if !allowedControllerEnvs.Has(v.Name) {
+				errs = errs.Also(apis.ErrInvalidKeyName(v.Name, "supported key is `MONGO_SERVER_URL`"))
```
jkandasa, line 59: IMHO, 
may be, no need to touch in the future, if we include supported fields dynamically.
```
errs = errs.Also(apis.ErrInvalidKeyName(v.Name, fmt.Sprintf("supported keys are %s",strings.Join(allowedControllerEnvs.List(), ","))))
```

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +515,49 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+			return nil
+		}
+
+		d := &appsv1.Deployment{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, d)
+		if err != nil {
+			return err
+		}
+
+		for _, v := range controllerEnvs {
```
jkandasa, line 531: first find the container and iterate the `controllerEnvs` inside the container, it can reduce the number of loops.

jkandasa:
@PuneetPunamiya please update the `Submitter Checklist`, `Release Notes`, and related ticket details, If you do not have GH issue, can be from outside from GitHub too.


image

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +515,49 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
 	}
 	return v1alpha1.RECONCILE_AGAIN_ERR
 }
+
+func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "Deployment" || len(controllerEnvs) == 0 {
```
jkandasa, line 521: Please check the deployment name too.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +517,51 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+		}
+
+		d := &appsv1.Deployment{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, d)
+		if err != nil {
+			return err
+		}
+
+		for i, c := range d.Spec.Template.Spec.Containers {
+			existingEnv := c.Env
```
jkandasa, line 534: `existingEnv := c.Env` can be under `if c.Name == chainsContainerName {`

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +517,51 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+			if c.Name == chainsContainerName {
+				appendNewEnv := true
```
jkandasa, line 536: `appendNewEnv := true` should be inside of `for` loop
```
for exisitingEnvIndex, env := range existingEnv {
  appendNewEnv := true
...
```

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +517,51 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+						}
+						if env.Name == newEnv.Name {
+							existingEnv[exisitingEnvIndex] = newEnv
+							appendNewEnv = false
+							break
+						}
+						// If not found append the new env
+						if appendNewEnv {
+							d.Spec.Template.Spec.Containers[i].Env = append(d.Spec.Template.Spec.Containers[i].Env, newEnv)
+						}
```
jkandasa, line 553: these lines should be outside of `for _, v := range controllerEnvs` loop.
```
// If not found append the new env
if appendNewEnv {
  d.Spec.Template.Spec.Containers[i].Env = append(d.Spec.Template.Spec.Containers[i].Env, newEnv)
}
```

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -509,3 +517,51 @@ func (r *Reconciler) markUpgrade(ctx context.Context, tc *v1alpha1.TektonChain)
...
+
+		unstrObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(d)
+		if err != nil {
+			return err
+		}
+
+		u.SetUnstructuredContent(unstrObj)
+		return nil
+	}
+}
```
jkandasa, line 567: @PuneetPunamiya I removed my earlier comments on this setion, I see more number of changes required, hence I re-arranged  the lines on this `func` as follows,
```go

func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
	return func(u *unstructured.Unstructured) error {
		if u.GetKind() != "Deployment" || len(controllerEnvs) == 0 || u.GetName() != chainsDeploymentName {
			return nil
		}

		d := &appsv1.Deployment{}
		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, d)
		if err != nil {
			return err
		}

		for i, c := range d.Spec.Template.Spec.Containers {
			if c.Name != chainsContainerName {
				continue
			}

			existingEnv := c.Env
			for _, v := range controllerEnvs {
				newEnv := corev1.EnvVar{
					Name:      v.Name,
					Value:     v.Value,
					ValueFrom: v.ValueFrom,
				}
				appendNewEnv := true
				for existingEnvIndex, env := range existingEnv {
					// Check for the key, if found replace it
					if env.Name == newEnv.Name {
						existingEnv[existingEnvIndex] = newEnv
						appendNewEnv = false
						break
					}
				}
				// If not found append the new env
				if appendNewEnv {
					existingEnv = append(existingEnv, newEnv)
				}
			}

			// update the changes into the actual container
			d.Spec.Template.Spec.Containers[i].Env = existingEnv
			break
		}

		unstrObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(d)
		if err != nil {
			return err
		}

		u.SetUnstructuredContent(unstrObj)
		return nil
	}
}

```

PR #1547, opened 2023-06-06
support custom secret in tekton results

jkandasa:
# Changes

For now there is no way to supply `S3` credentials into results api pod via secrets. This PR optionally injects those references via environment variable into `tekton-results-api` deployment.
As these values are injected as environment variable takes the higher precedence.

There is an issue in results repository about this: https://github.com/tektoncd/results/issues/469, https://github.com/tektoncd/results/issues/432

# Submitter Checklist

review them:

# Release Notes

```release-note
To get `S3` credentials from secret, a new property called `secret_name` added under TektonResult spec. User can create a secret with `S3` credentials and update the secret name in `secret_name` property.
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonresult_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonresult_types.go
@@ -83,6 +83,9 @@ type ResultsAPIProperties struct {
 	S3SecretAccessKey     string `json:"s3_secret_access_key,omitempty"`
 	S3MultiPartSize       int64  `json:"s3_multi_part_size,omitempty"`
 	LoggingPVCName        string `json:"logging_pvc_name"`
+	// name of the secret used to get S3 credentials and
+	// pass it as environment variables to the "tekton-results-api" deployment under "api" container
+	SecretName string `json:"secret_name,omitempty"`
 }
 
 // TektonResultStatus defines the observed state of TektonResult
--- a/pkg/reconciler/kubernetes/tektonresult/transform.go
+++ b/pkg/reconciler/kubernetes/tektonresult/transform.go
@@ -31,6 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"knative.dev/pkg/ptr"
 )
 
 const (
@@ -41,6 +42,20 @@ const (
 	configMetrics    = "tekton-results-config-observability"
 	configPostgresDB = "tekton-results-postgres"
 	pvcLoggingVolume = "tekton-logs"
+	apiContainerName = "api"
+)
+
+var (
+	// allowed property secret keys
+	allowedPropertySecretKeys = []string{
+		"S3_BUCKET_NAME",
+		"S3_ENDPOINT",
+		"S3_HOSTNAME_IMMUTABLE",
+		"S3_REGION",
+		"S3_ACCESS_KEY_ID",
+		"S3_SECRET_ACCESS_KEY",
+		"S3_MULTI_PART_SIZE",
+	}
 )
 
 // transform mutates the passed manifest to one with common, component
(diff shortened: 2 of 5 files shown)


pkg/reconciler/kubernetes/tektonresult/transform_test.go
```diff
@@ -121,3 +122,47 @@ S3_ACCESS_KEY_ID=secret
...
+					envFound = true
+					assert.Equal(t, propertyKey, _env.ValueFrom.SecretKeyRef.Key)
+					assert.Equal(t, secretName, _env.ValueFrom.SecretKeyRef.Name)
+					assert.Equal(t, true, *_env.ValueFrom.SecretKeyRef.Optional)
+				}
+			}
+			assert.Equal(t, true, envFound, fmt.Sprintf("property not found in env:%s", propertyKey))
+		}
+	}
+	assert.Equal(t, true, containerFound, "container not found")
```
PuneetPunamiya, line 167: This sounds a little confusing

pkg/apis/operator/v1alpha1/tektonresult_types.go
```diff
@@ -83,6 +83,9 @@ type ResultsAPIProperties struct {
 	S3SecretAccessKey     string `json:"s3_secret_access_key,omitempty"`
 	S3MultiPartSize       int64  `json:"s3_multi_part_size,omitempty"`
 	LoggingPVCName        string `json:"logging_pvc_name"`
+	// name of the secret used to get S3 credentials and
+	// pass it as environment variables to the "tekton-results-api" deployment under "api" container
+	SecretName string `json:"secret_name"`
```
PuneetPunamiya, line 88: I think we should start using a string pointer for a new field, because in case in future if we don't need that field, we can set the field directly to nil in defaulting and will also help in upgrade

PuneetPunamiya: Also I think we should add omitempty in the json tag

pkg/reconciler/kubernetes/tektonresult/transform_test.go
jkandasa: to verify the expected container found on the test,
```
if container.Name != apiContainerName {
```

pkg/apis/operator/v1alpha1/tektonresult_defaults.go
```diff
@@ -18,7 +18,10 @@ package v1alpha1
 
 import (
 	"context"
+	"strings"
 )
 
 func (tp *TektonResult) SetDefaults(ctx context.Context) {
+	// remove white space in the given secret name
+	tp.Spec.SecretName = strings.TrimSpace(tp.Spec.SecretName)
```
PuneetPunamiya, line 26: I think we don't need this

pkg/apis/operator/v1alpha1/tektonresult_types.go
jkandasa: added `omitempty`

pkg/apis/operator/v1alpha1/tektonresult_defaults.go
jkandasa: removed

jkandasa:
cc @piyush-garg @PuneetPunamiya @khrm

piyush-garg:
Issue on operator https://github.com/tektoncd/results/issues/432

pkg/reconciler/kubernetes/tektonresult/transform.go
```diff
@@ -210,3 +226,70 @@ func updateApiConfig(p interface{}) mf.Transformer {
...
+			// update only allowed properties
+			for _, propertyKey := range allowedPropertySecretKeys {
+				newEnv := corev1.EnvVar{
+					Name: propertyKey,
+					ValueFrom: &corev1.EnvVarSource{
+						SecretKeyRef: &corev1.SecretKeySelector{
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: props.SecretName,
+							},
+							Key:      propertyKey,
```
piyush-garg, line 264: Will there be any issues assume if a key is missed in secret ?

jkandasa: @piyush-garg I have added `Optional: ptr.Bool(true)`, So it will not give any issue, if the key not present in the secret.

piyush-garg: ohk, perfect

PR #1587, opened 2023-06-19
Handles chains config map during upgrade and adds provision to enable/disable chains

PuneetPunamiya:
- Handles chains config map during upgrade 

  - Previously when user installed chains by adding any of the
supported fields and then when upgraded to the latest version,
where chains is installed through TektonConfig, the data in the
TektonChain CR is overwritten

  - And also the data in the chains-config confimap was not deleted
because the manifestival package appends the latest data, instead of
replacing it

  - Hence, fixing this by creating separate `chains-config` configMap installerset

---
- Adds a provision to enable/disable chains

  - Previously, when operator was installed chains was installed
by default, but there can be a case where user doesn't needs
chains to installed

  - Hence, this patch provides a field with `disabled` where user
can provide a true/false value where user can decide, if we
wants to install chains or not

# Changes

# Submitter Checklist

review them:

# Release Notes

* NOTE:- If user has installed chains in `tekton-pipelines` or `openshift-pipelines` i.e `targetNamespace` then only it's previous data will be preserved on upgrade

### DIFF
--- a/pkg/reconciler/kubernetes/tektonchain/tektonchain.go
+++ b/pkg/reconciler/kubernetes/tektonchain/tektonchain.go
@@ -72,6 +72,7 @@ var _ tektonchainreconciler.Finalizer = (*Reconciler)(nil)
 const (
 	createdByValue          = "TektonChain"
 	secretChainInstallerset = "chain-secret"
+	configChainInstallerset = "chain-config"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -230,8 +237,10 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonChain
 			if tc.Spec.GetTargetNamespace() == pipelineNamespace {
 				manifest = manifest.Filter(mf.Not(mf.ByKind("Namespace")))
 			}
-			// remove secret from this installerset as this installerset will be deleted on upgrade
-			manifest = manifest.Filter(mf.Not(mf.ByKind("Secret")))
+			// remove secret and `chains-config` configMap from this installerset as this installerset will be deleted on upgrade
```
piyush-garg, line 240: this shoule be - `remove secret from this installerset as this installerset will be deleted on upgrade and `chains-config` configMap  as it will be in separate installerset`

vdemeester (approved):

PR #1654, opened 2023-08-24
Add support for External DB in Tekton Results

khrm:
# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonresult_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonresult_types.go
@@ -90,8 +90,8 @@ type ResultsAPIProperties struct {
 	SecretName         string `json:"secret_name,omitempty"`
 	GCSCredsSecretName string `json:"gcs_creds_secret_name,omitempty"`
 	GCSCredsSecretKey  string `json:"gcs_creds_secret_key,omitempty"`
-
-	// options holds additions fields and these fields will be updated on the manifests
+	IsExternalDB       bool   `json:"is_external_db"`
+	// Options holds additions fields and these fields will be updated on the manifests
 	Options AdditionalOptions `json:"options"`
 }
 
--- a/pkg/reconciler/kubernetes/tektonresult/tektonresult.go
+++ b/pkg/reconciler/kubernetes/tektonresult/tektonresult.go
@@ -229,6 +229,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tr *v1alpha1.TektonResul
 				return err
 			}
 
+			r.manifest = r.manifest.Filter(filterExternalDB(tr.Spec.IsExternalDB))
+
 			// Update the spec hash
 			current := installedTIS.GetAnnotations()
 			current[v1alpha1.LastAppliedHashKey] = expectedSpecHash
--- a/docs/TektonResult.md
+++ b/docs/TektonResult.md
@@ -20,7 +20,7 @@ To install Tekton Result on your cluster follow steps as given below:
     - namespace: `tekton-pipelines`
     - name: `tekton-results-postgres`
     - contains the fields:
-        - `user=root`
+        - `user=<user name>`
         - `password=<your password>`
 
   If you are not using a particular password management strategy, the following
@@ -31,7 +31,7 @@ To install Tekton Result on your cluster follow steps as given below:
    export NAMESPACE="tekton-pipelines"
    kubectl create secret generic tekton-results-postgres --namespace=${NAMESPACE} --from-literal=POSTGRES_USER=result --from-literal=POSTGRES_PASSWORD=$(openssl rand -base64 20)
    ```
-- Generate cert/key pair. 
+- Generate cert/key pair.
   Note: Feel free to use any cert management software to do this!
 
   Tekton Results expects the cert/key pair to be stored in a [TLS Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets).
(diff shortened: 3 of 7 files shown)


pkg/reconciler/kubernetes/tektonresult/transform.go
```diff
@@ -82,6 +84,18 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
 	return common.Transform(ctx, manifest, instance, extra...)
 }
 
+func isDBExternal(p v1alpha1.ResultsAPIProperties) mf.Transformer {
```
jkandasa, line 87: the name of the function say, it returns `bool`, can you please rename it?

pkg/apis/operator/v1alpha1/tektonresult_types.go
```diff
@@ -85,7 +85,8 @@ type ResultsAPIProperties struct {
 	LoggingPVCName        string `json:"logging_pvc_name"`
 	// name of the secret used to get S3 credentials and
 	// pass it as environment variables to the "tekton-results-api" deployment under "api" container
-	SecretName string `json:"secret_name,omitempty"`
+	SecretName   string `json:"secret_name,omitempty"`
+	IsDBExternal bool   `json:"is_db_external"`
```
jkandasa, line 89: IMHO, `IsExternalDB`, looks better

jkandasa (commented):
Thanks for the PR @khrm ,
* can you remove `[WIP]` from commit message?
* with this PR, you are removing internal database deployment. But there is no code to map the external database to results deployment :thinking: 
* Can you add document to the new fields introduced here and how to setup external database, If user want to set external database with results?

pkg/reconciler/kubernetes/tektonresult/transform.go
khrm: Yes, agreed.

pkg/apis/operator/v1alpha1/tektonresult_types.go
khrm: Sure. Will change.

khrm (commented):
@tektoncd/operator-collaborators 

This pr is also ready for review.

pkg/apis/operator/v1alpha1/tektonresult_types.go
khrm: Changed to that.

pkg/reconciler/kubernetes/tektonresult/transform.go
khrm: Changed to filterExternalDB.

vdemeester (approved):

khrm:
There seems to be some problem with the tide. Why is it showing to rebase even when files are different?

pkg/apis/operator/v1alpha1/tektonresult_types.go
```diff
@@ -85,7 +85,8 @@ type ResultsAPIProperties struct {
 	LoggingPVCName        string `json:"logging_pvc_name"`
 	// name of the secret used to get S3 credentials and
 	// pass it as environment variables to the "tekton-results-api" deployment under "api" container
-	SecretName string `json:"secret_name,omitempty"`
+	SecretName   string `json:"secret_name,omitempty"`
+	IsExternalDB bool   `json:"is_db_external"`
```
piyush-garg, line 89: key and json are different

pkg/reconciler/kubernetes/tektonresult/transform.go
```diff
@@ -82,6 +84,18 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
 	return common.Transform(ctx, manifest, instance, extra...)
 }
 
+func filterExternalDB(p v1alpha1.ResultsAPIProperties) mf.Transformer {
```
piyush-garg, line 87: i dont know, what this is doing

piyush-garg:
i am not able to understand this PR, if the flag is set what are we doing in results.

And i dont see how the external db is getting configured in results

pkg/reconciler/kubernetes/tektonresult/transform.go
jkandasa: @khrm I too have same question. Could you please elaborate about this transformer?

pkg/apis/operator/v1alpha1/tektonresult_types.go
jkandasa: @khrm please update json tag too

pkg/reconciler/kubernetes/tektonresult/transform.go
khrm: It removes the statefulset for DB.

khrm (commented):
We don't need to do any additional thing for ExternalDB. This PR just disabled the default statefulset for DB. Instead, the user can supply external DB configurations to Postres and it should work.

khrm:
This is also requesting rebase when no changes required.

pkg/apis/operator/v1alpha1/tektonresult_types.go
khrm: Updating the tag. I have forgotten.

khrm: Done.

pkg/reconciler/kubernetes/tektonresult/transform.go
piyush-garg: if you want to remove this resource from manifest, you should do a filterning for that resource based on flag,

not transform like this

piyush-garg:
> We don't need to do any additional thing for ExternalDB. This PR just disabled the default statefulset for DB. Instead, the user can supply external DB configurations to Postres and it should work.

How user will supply, how it can be configured with results reconciler, i am not getting those part here in this PR.
Also this does not contain any documentation for configuring

khrm (commented):
@piyush-garg I have updated `TektonResult.md`  doc with details.

pkg/reconciler/kubernetes/tektonresult/transform.go
khrm: Filter doesn't remove resources. It uses a deepcopy. 
As manifestival write: 
```
The only way to alter resources in a Manifest is with its Transform method.
```

khrm: Changed to using a filter.

khrm: Putting this to resolved.

khrm (commented):
@piyush-garg @jkandasa Please relook into this.

jkandasa:
LGTM
cc @piyush-garg

pkg/reconciler/kubernetes/tektonresult/tektonresult.go
```diff
@@ -229,6 +229,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tr *v1alpha1.TektonResul
 				return err
 			}
 
+			r.manifest = r.manifest.Filter(filterExternalDB(tr.Spec.IsExternalDB))
```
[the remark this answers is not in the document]
khrm: Yes, we could done something like this but it's not that much efficient. We have three context calls instead of one when we filter but it won't happen unless the code arrives there.

Another thing to note is we lose on unit testing here because we aren't unit testing ReconcileKind. I would prefer to have a separate function in ReconcileKind so that we can unit test.  @jkandasa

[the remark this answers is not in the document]
khrm: So instead of 			
```
if tr.Spec.IsExternalDB {
	 r.manifest = r.manifest.Filter(mf.Not(mf.All(mf.ByKind("StatefulSet"), mf.ByName("tekton-results-postgres"))))
}
```
 We could have 
 ```
 r. manifest = filterExternalDB(r.manifest)
 ```
 And `filterExternalDB` can have those lines.

PR #1675, opened 2023-09-05
introduce upgrade mechanism and add upgrade patch to v1 storage migration

jkandasa:
# Changes

This PR introduces upgrade mechanism. Which will be executed as two part.
* PreUpgrade - triggers before applying manifests
* PostUpgrade - triggers once TektonConfig CR becomes ready state 

marks last applied upgrade reference to `TektonConfig` CR(via status annotation)

In addition to upgrade base code, this PR includes the following fixes

* fixes: [SRVKP-2917](https://issues.redhat.com/browse/SRVKP-2917) - Handle `v1` storage conversion
* fixes: [SRVKP-2935](https://issues.redhat.com/browse/SRVKP-2935) - Upgrade to version 1.9.x results in the following error: Install failed with message: ... status.storedVersions[0]: Invalid value: \"v1alpha1\": must appear in spec.versions
* fixes: [SRVKP-3202](https://issues.redhat.com/browse/SRVKP-3202) - TektonConfig CR displays message " Install failed with message: tektonpipeline pipeline is not in desired state" after pipelines operator upgrade

# Submitter Checklist

review them:

# Release Notes
```release-note
NONE
```


vdemeester (commented):
Few questions:
- Any reason why the generated injection code is removed ? (probably due to the knative.dev/pkg update ?)
- Any reason to have an `upgrade/helper` package instead of just having things in `upgrade` ? (`helper` is a package name we tend to "not use" because it doesn't really mean anything)

Otherwise I like the PR but I think we need to have some unit-tests at least 👼🏼

jkandasa:
> Few questions:
> 
> * Any reason why the generated injection code is removed ? (probably due to the knative.dev/pkg update ?)

Yes, I updated the `knative.dev/pkg` to get `knative.dev/pkg/apiextensions`

> * Any reason to have an `upgrade/helper` package instead of just having things in `upgrade` ? (`helper` is a package name we tend to "not use" because it doesn't really mean anything)

In the future, If we want to call storage migration, we can reuse that function.
Thought of keeping reusable functions inside `helper` package.
Other than that I have no reason.

> 
> Otherwise I like the PR but I think we need to have some unit-tests at least 👼🏼

I was waiting for the initial confirmation about this feature to put effort on unit-tests.
Now I can work on unit-tests and other pending works on this PR.
Thanks @vdemeester :smile:

upgrade/helper/storage_version_migrator.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"go.uber.org/zap"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"knative.dev/pkg/apiextensions/storageversion"
+)
+
+// performs crd storage version upgrade
+// lists all the resources and,
+// keeps only one storage version on the crd
```
piyush-garg, line 31: I don't think, this is what we wanted to keep on CR i.e. only 1 version

What I know about the issue is - the fresh installation and old installation can have different version, but we need to update the CRD by removing the old versions and using the new versions. We can't delete and create again, as it will remove user CRs.

jkandasa: @piyush-garg My understanding is,
* storage version should contain latest version and may contain old version too
* from the `migrate` package [1], first updating all the resources to latest version
* then keeps only the latest version under `status.storedVersions`

[1] - https://github.com/knative/pkg/blob/main/apiextensions/storageversion/migrator.go#L49-L56

piyush-garg: i am not sure if this will work like this, because just changing the version will not work for tekton resources, spec also needs to handled accordingly.

what i was expecting from this story was to handle the change of this field https://github.com/tektoncd/pipeline/blob/main/config/300-pipeline.yaml#L27

piyush-garg:
/approve
/lgtm

we need to properly test this more

PR #1703, opened 2023-09-15
Fix bitbucket clustertriggerbinding when both SSH and HTTPS enabled

khrm:
# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/cmd/openshift/operator/kodata/tekton-addon/addons/01-clustertriggerbindings/bitbucket.yaml
+++ b/cmd/openshift/operator/kodata/tekton-addon/addons/01-clustertriggerbindings/bitbucket.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   params:
     - name: gitrepo-url
-      value: $(body.pullRequest.fromRef.repository.links.clone[0].href)
+      value: $(body.pullRequest.fromRef.repository.links.clone[?(@.name=="ssh")].href)
     - name: pullreq-sha
       value: $(body.pullRequest.fromRef.latestCommit)
     - name: pullreq-state


cmd/openshift/operator/kodata/tekton-addon/addons/01-clustertriggerbindings/bitbucket.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   params:
     - name: gitrepo-url
-      value: $(body.pullRequest.fromRef.repository.links.clone[0].href)
+      value: $(body.pullRequest.fromRef.repository.links.clone[?(@.name=="ssh")].href)
```
jkandasa, line 8: @khrm what happens if `ssh` object not present in clone array?

khrm: It will fail.

khrm: @af-wieda Why is https not working for you with git-cloneeeee?

khrm: Is it because it's sometimes https and sometimes SSH?

khrm: Is it possible in bitbucket-server to not have ssh? I know it's not possible in github.

khrm: Is it possible in bitbucket-server to not have ssh? I know it's not possible in GitHub. @savitaashture

jkandasa:
/hold

currently it is not clear about the clone url. I am holding this PR until we get a proper fix.

khrm:
@jkandasa Issue is that current variable behaves in undefined way.
Index shouldn't be used. We should use name to assign value.

jkandasa:
@khrm 

>Issue is that current variable behaves in undefined way.
Index shouldn't be used. We should use name to assign value.

I agree. But we should confirm gitlab has always `ssh`.

cmd/openshift/operator/kodata/tekton-addon/addons/01-clustertriggerbindings/bitbucket.yaml
af-wieda: Totally missed this thread of comments.

You are correct, the array position is not predictable and will sometimes give https and sometimes ssh which makes the Bitbucket Datacenter ClusterTriggerBinding unusable if the Bitbucket Datacenter instance has HTTPS enabled. We currently have to modify it after each cluster upgrade, which is a workaround I suppose.

SSH is the default for **Bitbucket Datacenter** and it is not possible to turn it off and use only HTTPS.

This does not to my knowledge affect any other Git-server providers as the scope is the default ClusterTriggerBinding for Bitbucket Datacenter.

jkandasa (approved):

jkandasa:
cc @piyush-garg

PR #1865, opened 2023-11-21
Fix to deploy TektonHub in custom namespace

pratap0007:
# Changes
This will add the fix for targetnamespace deletion
it will delete the namespace only if namespace contains
ownerreference of same reconcilation

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix targetnamespace deletion
```

### DIFF
--- a/pkg/reconciler/common/targetnamespace.go
+++ b/pkg/reconciler/common/targetnamespace.go
@@ -57,9 +57,10 @@ func ReconcileTargetNamespace(ctx context.Context, labels map[string]string, tek
 			_targetNamespace := namespace.DeepCopy()
 			targetNamespace = _targetNamespace
 		} else {
-			// delete irrelevant namespaces
+			// delete irrelevant namespaces if the owner is the same component
 			// if deletionTimestamp is not nil, that indicates, the namespace is in deletion state
-			if namespace.DeletionTimestamp == nil {
+			ownerReferenceName := namespace.GetOwnerReferences()[0].Name
+			if namespace.DeletionTimestamp == nil && ownerReferenceName == tektonComponent.GetName() {
 				if err := kubeClientSet.CoreV1().Namespaces().Delete(ctx, namespace.Name, metav1.DeleteOptions{}); err != nil {
 					logger.Errorw("error on deleting a namespace",
 						"namespace", namespace.Name,
@@ -67,8 +68,9 @@ func ReconcileTargetNamespace(ctx context.Context, labels map[string]string, tek
 					)
 					return err
 				}
-			} else {
-				logger.Infof("'%v' namespace is in deletion state", namespace.Name)
+			}
+			if namespace.DeletionTimestamp != nil {
+				logger.Debugf("'%v' namespace is in deletion state", namespace.Name)
 				namespaceDeletionInProgress = true
 			}
 		}
--- a/pkg/reconciler/common/targetnamespace_test.go
+++ b/pkg/reconciler/common/targetnamespace_test.go
@@ -59,6 +59,14 @@ func TestReconcileTargetNamespace(t *testing.T) {
 			err: nil,
 		},
 		{
+			name: "verify-custom-target-namespace-tekton-hub",
+			component: &v1alpha1.TektonHub{
+				ObjectMeta: metav1.ObjectMeta{Name: "hub"},
+				Spec:       v1alpha1.TektonHubSpec{CommonSpec: v1alpha1.CommonSpec{TargetNamespace: "custom"}},
+			},
+			err: nil,
+		},
+		{
 			name: "verify-custom-target-namespace",
 			component: &v1alpha1.TektonConfig{
 				ObjectMeta: metav1.ObjectMeta{Name: "config"},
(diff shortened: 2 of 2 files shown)


vdemeester (approved):
/cc @jkandasa @concaf

jkandasa:
/hold
@vdemeester @pratap0007  we are keeping only `targetNamespace` and deleting all other namespaces which has `labels` that matches to `targetNamespace`, Those was used as target namespace.

With this PR we are removing the deletion part completely. I need sometime to verify this change.

Thanks for the PR @pratap0007

pratap0007:
> /hold @vdemeester @pratap0007 we are keeping only `targetNamespace` and deleting all other namespaces which has `labels` that matches to `targetNamespace`, Those was used as target namespace.
> 
> With this PR we are removing the deletion part completely. I need sometime to verify this change.
> 
> Thanks for the PR @pratap0007

why do we want to delete all other namespaces other than `targetNamespace` ?

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -1035,3 +1030,24 @@ func getConfigDataFromHubURL(th *v1alpha1.TektonHub) (*Data, error) {
 
 	return data, nil
 }
+
+// checks target namespace, creates namespace if not found
+func (r *Reconciler) checkAndCreateNameSpace(ctx context.Context, th *v1alpha1.TektonHub) error {
+	if th.Spec.TargetNamespace != "openshift-pipelines" {
```
jkandasa, line 1036: @pratap0007 in kubernetes we use `tekton-pipelines` and in openshift `openshift-pipelines`

jkandasa: Why do we need this check?

pratap0007: To check the hub target namespace is same as tekton pipelines target namespace or not, if not then it creates the namespace

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -1035,3 +1035,24 @@ func getConfigDataFromHubURL(th *v1alpha1.TektonHub) (*Data, error) {
...
+func (r *Reconciler) checkAndCreateNameSpace(ctx context.Context, th *v1alpha1.TektonHub, tektonConfigNamespace string) error {
+	if th.Spec.TargetNamespace != tektonConfigNamespace {
+		_, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, th.Spec.GetTargetNamespace(), metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				ns := &corev1.Namespace{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: th.Spec.TargetNamespace,
+					},
+				}
```
PuneetPunamiya, line 1049: I think we need to add the ownerRef as well 
cc @jkandasa @piyush-garg

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -175,12 +175,12 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, th *v1alpha1.TektonHub)
...
 	th.SetDefaults(ctx)
 
-	// reconcile target namespace
-	if err := common.ReconcileTargetNamespace(ctx, nil, th, r.kubeClientSet); err != nil {
-		logger.Errorw("error on reconciling targetNamespace",
-			"targetNamespace", th.Spec.GetTargetNamespace(),
-			err,
-		)
+	tc, err := r.operatorClientSet.OperatorV1alpha1().TektonConfigs().Get(ctx, v1alpha1.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
```
jkandasa, line 179: @pratap0007 still it is possible to install Hub without `TektonConfig` right?
if yes, ignore resource not found error.

PuneetPunamiya: Good catch, I think with this would not work if TektonConfig is not there

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -175,12 +175,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, th *v1alpha1.TektonHub)
 
 	th.SetDefaults(ctx)
 
-	// reconcile target namespace
-	if err := common.ReconcileTargetNamespace(ctx, nil, th, r.kubeClientSet); err != nil {
-		logger.Errorw("error on reconciling targetNamespace",
-			"targetNamespace", th.Spec.GetTargetNamespace(),
-			err,
-		)
+	tc, _ := r.operatorClientSet.OperatorV1alpha1().TektonConfigs().Get(ctx, v1alpha1.ConfigResourceName, metav1.GetOptions{})
```
PuneetPunamiya, line 178: I still feel this would not work when TektonConfig is nil

pratap0007: can you please explain the case when TektonConfig could be nil, so I can check once this ?

PuneetPunamiya: Meaning TektonHub is independent of TekttonConfig, so ideally we should not check for the namespace of TektonConfig, we can directly check if the namespace mentioned in TektonHub is present or not and based on that we can directly create that

For ref: https://github.com/tektoncd/operator/commit/51fff501ad3175a5a75721625d7f5384209b1b77#diff-1c3da4772ce1a38a45935ea0814741d91d7803c96d9aef2b0f10b00f6d8b33f3L864-L875

pratap0007: okay, thank you

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -1035,3 +1030,25 @@ func getConfigDataFromHubURL(th *v1alpha1.TektonHub) (*Data, error) {
...
+func (r *Reconciler) createNameSpace(ctx context.Context, th *v1alpha1.TektonHub) error {
+	_, err := r.kubeClientSet.CoreV1().Namespaces().Get(ctx, th.Spec.GetTargetNamespace(), metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			ownerRef := *metav1.NewControllerRef(th, th.GroupVersionKind())
+
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            th.Spec.TargetNamespace,
+					OwnerReferences: []metav1.OwnerReference{ownerRef},
```
jkandasa, line 1044: I have a little concern on owner reference.

There is a possibility: 
* `hub` installed earlier than `pipeline` on `tekton-pipelines` namespace
* later `pipeline` installed on `tekton-pipelines` namespace (still the owner reference will be `hub`)
* now, if we delete the `hub` CR, that will remove the `tekton-pipelines` namespace along with `pipeline`

@piyush-garg @PuneetPunamiya

piyush-garg: Yes namesapce will be deleted and recreated

PR #1890, opened 2023-12-05
Update feature flags for pipeline

piyush-garg:
This will update feature flags of pipeline component to be compatible with latest 0.54 release of pipeline

# Submitter Checklist

review them:

# Release Notes

```release-note
Update feature flags for pipeline
```


pkg/apis/operator/v1alpha1/tektonpipeline_validation.go
```diff
@@ -59,11 +62,29 @@ func (p *PipelineProperties) validate(path string) (errs *apis.FieldError) {
 		}
 	}
 
+	if p.MaxResultSize != nil {
+		if *p.MaxResultSize >= 1572864 {
```
jkandasa, line 66: @piyush-garg can you add some reference about `1572864`, from where did we get this

pkg/apis/operator/v1alpha1/tektonpipeline_types.go
```diff
@@ -107,7 +107,18 @@ type PipelineProperties struct {
 
 	// ScopeWhenExpressionsToTask Deprecated: remove in next release
 	ScopeWhenExpressionsToTask *bool `json:"scope-when-expressions-to-task,omitempty"`
-	PipelineMetricsProperties  `json:",inline"`
+
+	EnforceNonfalsifiability  string `json:"enforce-nonfalsifiability,omitempty"`
+	EnableKeepPodOnCancel     *bool  `json:"keep-pod-on-cancel,omitempty"`
+	ResultExtractionMethod    string `json:"results-from,omitempty"`
```
jkandasa, line 113: json tag looks different than the variable name, is that exoected?
`ResultExtractionMethod != results-from`

jkandasa (commented):
@piyush-garg thanks for the PR.

Can you please add a documentation about the new variables?

pkg/apis/operator/v1alpha1/tektonpipeline_validation.go
piyush-garg: This field has limit of value equal to 1.5MB as mentioned here https://github.com/tektoncd/pipeline/blob/main/docs/additional-configs.md#enabling-larger-results-using-sidecar-logs

And I took the value from here https://github.com/tektoncd/pipeline/blob/main/pkg/apis/config/feature_flags.go#L346

pkg/apis/operator/v1alpha1/tektonpipeline_types.go
piyush-garg: I followed what pipeline has done and what we're doing for other fields https://github.com/tektoncd/pipeline/blob/main/pkg/apis/config/feature_flags.go#L125

pkg/apis/operator/v1alpha1/tektonpipeline_validation.go
jkandasa: Thanks @piyush-garg 
It will be nicer, if we add those supporting docs as comment on top of the block.

vdemeester (commented):
@jkandasa @concaf do we want to keep those in since or do we want to deprecate them and rely on additionnal fields in the future ?

jkandasa:
> @jkandasa @concaf do we want to keep those in since or do we want to deprecate them and rely on additionnal fields in the future ?

IMHO, We can rely on `options` (additional fields). No need to work on new fields

@piyush-garg WDYT?

piyush-garg:
> > @jkandasa @concaf do we want to keep those in since or do we want to deprecate them and rely on additionnal fields in the future ?
> 
> IMHO, We can rely on `options` (additional fields). No need to work on new fields
> 
> @piyush-garg WDYT?

Sure we should more move towards `options`. For now, we can have this for consistency and in future release, we need to mark them deprecated and provide migration during upgrade that whatever specified will be moved to options so smooth upgrade and no action items on user.

jkandasa (approved):

concaf (approved):
/lgtm

PR #2003, opened 2024-02-15
Fix the results reconciler's manifest reference 

khrm:
Earlier reconciler's manifest value was being changed every cycle. This fixes is so that, we have same manifest every reconciler. We take value from reconciler at the start of reconcile cycle and transform it.

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/reconciler/kubernetes/tektonresult/filter.go
+++ b/pkg/reconciler/kubernetes/tektonresult/filter.go
@@ -26,10 +26,11 @@ const (
 	servicePostgresDB = "tekton-results-postgres-service"
 )
 
-func (r *Reconciler) filterExternalDB(tr *v1alpha1.TektonResult) {
+func filterExternalDB(tr *v1alpha1.TektonResult, manifest mf.Manifest) *mf.Manifest {
 	if tr.Spec.IsExternalDB {
-		r.manifest = r.manifest.Filter(mf.Not(mf.All(mf.ByKind("StatefulSet"), mf.ByName(statefulSetDB))))
-		r.manifest = r.manifest.Filter(mf.Not(mf.All(mf.ByKind("ConfigMap"), mf.ByName(configPostgresDB))))
-		r.manifest = r.manifest.Filter(mf.Not(mf.All(mf.ByKind("Service"), mf.ByName(servicePostgresDB))))
+		manifest = manifest.Filter(mf.Not(mf.All(mf.ByKind("StatefulSet"), mf.ByName(statefulSetDB))))
+		manifest = manifest.Filter(mf.Not(mf.All(mf.ByKind("ConfigMap"), mf.ByName(configPostgresDB))))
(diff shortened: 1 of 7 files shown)


jkandasa:
@khrm did you test this locally with `1.13.1` to `1.14.0` (with this PR) upgrade path on a OpenShift environment?
Does everything works ok?

jkandasa (approved):

pkg/reconciler/kubernetes/tektonresult/filter.go
```diff
@@ -26,10 +26,11 @@ const (
 	servicePostgresDB = "tekton-results-postgres-service"
 )
 
-func (r *Reconciler) filterExternalDB(tr *v1alpha1.TektonResult) {
+func (r *Reconciler) filterExternalDB(tr *v1alpha1.TektonResult, manifest mf.Manifest) mf.Manifest {
```
piyush-garg, line 29: why we are passing this, can we take it from reconcile?

pkg/reconciler/kubernetes/tektonresult/installerset.go
```diff
@@ -26,10 +26,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func (r *Reconciler) createInstallerSet(ctx context.Context, tr *v1alpha1.TektonResult) (*v1alpha1.TektonInstallerSet, error) {
+func (r *Reconciler) createInstallerSet(ctx context.Context, tr *v1alpha1.TektonResult, manifest mf.Manifest) (*v1alpha1.TektonInstallerSet, error) {
 
-	r.filterExternalDB(tr)
-	if err := r.transform(ctx, &r.manifest, tr); err != nil {
+	r.filterExternalDB(tr, manifest)
```
piyush-garg, line 31: it seems like here we are not receiving the value returned

pkg/reconciler/kubernetes/tektonresult/filter.go
khrm: We were passing it because we didn't want to modify r.Manifest.

pkg/reconciler/kubernetes/tektonresult/installerset.go
khrm: We were changing the value, so the receiver wasn't needed.

khrm: Now I have changed the function.

pkg/reconciler/kubernetes/tektonresult/transform.go
```diff
@@ -68,6 +68,7 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp
 	resultImgs := common.ToLowerCaseKeys(common.ImagesFromEnv(common.ResultsImagePrefix))
 
 	targetNs := comp.GetSpec().GetTargetNamespace()
+	manifest = filterExternalDB(instance, *manifest)
```
khrm, line 71: Calling this here changes the address.

PR #2015, opened 2024-02-21
Disallow to update targetNamespace in Hub CR

pratap0007:
This patch disallow to update the targetNamespace in
the Hub CR 

# Changes

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonhub_validation.go
+++ b/pkg/apis/operator/v1alpha1/tektonhub_validation.go
@@ -45,6 +45,14 @@ func (th *TektonHub) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return nil
 	}
 
+	// disallow to update the targetNamespace
+	if apis.IsInUpdate(ctx) {
+		existingTC := apis.GetBaseline(ctx).(*TektonHub)
+		if existingTC.Spec.GetTargetNamespace() != th.Spec.GetTargetNamespace() {
+			errs = errs.Also(apis.ErrGeneric("doesn't allow to update targetNamespace, delete existing TektonHub object and create the updated TektonHub object", "spec.targetNamespace"))
+		}
+	}
+
 	// execute common spec validations
 	errs = errs.Also(th.Spec.CommonSpec.validate("spec"))
 
--- a/pkg/apis/operator/v1alpha1/tektonhub_validation_test.go
+++ b/pkg/apis/operator/v1alpha1/tektonhub_validation_test.go
@@ -22,6 +22,7 @@ import (
 
 	"gotest.tools/v3/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/apis"
 )
 
 func Test_ValidateTektonConfig_InvalidHubParam(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jkandasa:
@pratap0007 We are not allowing changing namespace in `TektonConfig` CR.
Why should we have to allow in `TektonHub` CR?

jkandasa:
> There was an use case, user wanted to install Hub in a different namespace

As we do it for TektonConfig, user can delete the existing `TektonHub` CR and create new `TektonHub` CR with different namespace.

pkg/reconciler/kubernetes/tektonhub/installerset.go
```diff
@@ -41,21 +41,32 @@ func (r *Reconciler) checkIfInstallerSetExist(ctx context.Context, oc clientset.
...
+		if ok && existingNamespace != targetNamespace {
+			deleteInstallerSets = true
+		}
+
 		version, ok := ctIs.Annotations[v1alpha1.ReleaseVersionKey]
+		if ok && version != relVersion {
+			// release version doesn't exist or is different from expected
+			// mark the deleteInstallerSets true
+			deleteInstallerSets = true
+		}
```
PuneetPunamiya, line 56: May be we can club this as we do for other components https://github.com/tektoncd/operator/blob/main/pkg/reconciler/kubernetes/tektonchain/tektonchain.go#L236-L244 and we might not even need that flag `deleteInstallerSets`

pratap0007:
> > There was an use case, user wanted to install Hub in a different namespace
> 
> As we do it for TektonConfig, user can delete the existing `TektonHub` CR and create new `TektonHub` CR with different namespace.

yes, we can do but this condition is handled for all components like chains has here https://github.com/tektoncd/operator/blob/main/pkg/reconciler/kubernetes/tektonchain/tektonchain.go#L244 , so we should have like this in Hub component, WDYT ?

jkandasa:
>@jkandasa yes, we can do but this condition is handled for all components like chains has here https://github.com/tektoncd/operator/blob/main/pkg/reconciler/kubernetes/tektonchain/tektonchain.go#L244 , so we should have like this in Hub component, WDYT ?

@vdemeester @piyush-garg WDYT about this PR?
IMHO, We are not allowing to change namespace via TektonConfig CR. I want to see the same for other components CR too.

piyush-garg:
> > @jkandasa yes, we can do but this condition is handled for all components like chains has here https://github.com/tektoncd/operator/blob/main/pkg/reconciler/kubernetes/tektonchain/tektonchain.go#L244 , so we should have like this in Hub component, WDYT ?
> 
> @vdemeester @piyush-garg WDYT about this PR? IMHO, We are not allowing to change namespace via TektonConfig CR. I want to see the same for other components CR too.

I agree here, as we are now disabling changing targetNamespace in tektonconfig, same we should do for HUB

pratap0007:
if we are not going to support to change the  targetNamespace in tektonconfig,then shall I remove changes related to the targetNamespace from this PR ?

piyush-garg:
> if we are not going to support to change the targetNamespace in tektonconfig,then shall I remove changes related to the targetNamespace from this PR ?

we mean like we can add the same logic for tektonhub validation like we did for tektonconfig, to not change namespace value in upgrade/ecit etc

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
```diff
@@ -150,6 +151,29 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, original *v1alpha1.Tekton
 	if err := r.extension.Finalize(ctx, original); err != nil {
 		logger.Error("Failed to finalize platform resources", err)
 	}
+
```
piyush-garg, line 154: Do we need to handle all this? Can't we better do add ownerReference in job and it gets automatically handled?

pratap0007:
> > if we are not going to support to change the targetNamespace in tektonconfig,then shall I remove changes related to the targetNamespace from this PR ?
> 
> we mean like we can add the same logic for tektonhub validation like we did for tektonconfig, to not change namespace value in upgrade/ecit etc
Okay, then I update the PR

pkg/reconciler/kubernetes/tektonhub/tektonhub.go
pratap0007: > Do we need to handle all this? Can't we better do add ownerReference in job and it gets automatically handled?

Job has already ownerReference of `db-migration` installerset and while deleting hub config cr it deletes the db-migration installerset and removes ownerReference from the job

piyush-garg: i think we can handle this better, see this https://github.com/kubernetes/client-go/issues/495

pkg/apis/operator/v1alpha1/tektonhub_validation.go
```diff
@@ -45,6 +45,14 @@ func (th *TektonHub) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return nil
 	}
 
+	// disallow to update the targetNamespace
+	if apis.IsInUpdate(ctx) {
+		existingTC := apis.GetBaseline(ctx).(*TektonHub)
+		if existingTC.Spec.GetTargetNamespace() != th.Spec.GetTargetNamespace() {
+			errs = errs.Also(apis.ErrGeneric("Doesn't allow to update targetNamespace, delete existing TektonHub config and create the updated TektonHub config", "spec.targetNamespace"))
```
piyush-garg, line 52: s/config/object

It will confuse with config object otherwise

pkg/apis/operator/v1alpha1/tektonhub_validation.go
```diff
@@ -45,6 +45,14 @@ func (th *TektonHub) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return nil
 	}
 
+	// disallow to update the targetNamespace
+	if apis.IsInUpdate(ctx) {
+		existingTC := apis.GetBaseline(ctx).(*TektonHub)
+		if existingTC.Spec.GetTargetNamespace() != th.Spec.GetTargetNamespace() {
+			errs = errs.Also(apis.ErrGeneric("Doesn't allow to update targetNamespace, delete existing TektonHub object and create the updated TektonHub object", "spec.targetNamespace"))
```
jkandasa, line 52: @pratap0007 please start the error with lowercase

jkandasa (approved):
/lgtm

PR #2033, opened 2024-02-26
Fix buildah clustertask to support long image names

jordigilh:
# Changes

Fixes a problem in the buildah clustertask where when provided with a long image name (e.g. `image-registry.openshift-image-registry.svc:5000/test/serverless-workflow-mtaanalysis:1.0`), it caused the generated script to wrap the line which is interpreted by the shell as a new command, rather than part of the previous command:

Currently, the script generates this `push` command:
```
buildah --storage-driver=vfs push \
 --tls-verify=true \
--digestfile /tmp/image-digest \
image-registry.openshift-image-registry.svc:5000/test/serverless-workflow-mtaanalysis:1.0
 \
docker://image-registry.openshift-image-registry.svc:5000/test/serverless-workflow-mtaanalysis:1.0
```

Which, when executed, it fails with this error:
```
Writing manifest to image destination
/tekton/scripts/script-0-4dsn8: line 35: docker://image-registry.openshift-image-registry.svc:5000/test/serverless-workflow-mtaanalysis:1.0: No such file or directory
```

The proposal fix here is to capture the value of the image in an environment variable and use it in the `push` command, avoiding the wrapping and ensuring it is not affected by any image length.

```
IMAGE=$(params.IMAGE)
buildah --storage-driver=$(params.STORAGE_DRIVER) push \
  $(params.PUSH_EXTRA_ARGS) --tls-verify=$(params.TLSVERIFY) \
  --digestfile /tmp/image-digest $IMAGE docker://$IMAGE
```

# Submitter Checklist

review them:

- [X] Run `make test lint` before submitting a PR

# Release Notes
```release-note
Fix buildah clustertask to support long image names
```

### DIFF
--- a/cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
+++ b/cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
@@ -108,13 +108,12 @@ spec:
           exit 1
         fi
       fi
-
+      IMAGE=$(params.IMAGE)
       buildah --storage-driver=$(params.STORAGE_DRIVER) push \
         $(params.PUSH_EXTRA_ARGS) --tls-verify=$(params.TLSVERIFY) \
-        --digestfile /tmp/image-digest $(params.IMAGE) \
-        docker://$(params.IMAGE)
+        --digestfile /tmp/image-digest $IMAGE docker://$IMAGE
       cat /tmp/image-digest | tee $(results.IMAGE_DIGEST.path)
-      echo -n "$(params.IMAGE)" | tee $(results.IMAGE_URL.path)
+      echo -n "$IMAGE" | tee $(results.IMAGE_URL.path)
     volumeMounts:
     - name: varlibcontainers
       mountPath: /var/lib/containers
--- a/cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
+++ b/cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
@@ -79,7 +79,7 @@ spec:
     workingDir: $(workspaces.source.path)
     script: |
       ENTITLEMENT_VOLUME=""
-
+      IMAGE=$(params.IMAGE)
       if [[ "$(workspaces.rhel-entitlement.bound)" == "true" ]]; then
         ENTITLEMENT_VOLUME="--volume /tmp/entitlement:/etc/pki/entitlement"
       fi
@@ -87,7 +87,7 @@ spec:
       buildah bud --storage-driver=$(params.STORAGE_DRIVER) \
         $ENTITLEMENT_VOLUME $(params.BUILD_EXTRA_ARGS) \
         --format=$(params.FORMAT) --tls-verify=$(params.TLSVERIFY) \
-        -f $(params.DOCKERFILE) -t $(params.IMAGE) $(params.CONTEXT)
+        -f $(params.DOCKERFILE) -t $IMAGE $(params.CONTEXT)
 
       [[ "$(params.SKIP_PUSH)" == "true" ]] && echo "Push skipped" && exit 0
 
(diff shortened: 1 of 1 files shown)


cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
```diff
@@ -108,13 +108,12 @@ spec:
           exit 1
         fi
       fi
-
+      IMAGE=$(params.IMAGE)
       buildah --storage-driver=$(params.STORAGE_DRIVER) push \
         $(params.PUSH_EXTRA_ARGS) --tls-verify=$(params.TLSVERIFY) \
-        --digestfile /tmp/image-digest $(params.IMAGE) \
-        docker://$(params.IMAGE)
+        --digestfile /tmp/image-digest $IMAGE docker://$IMAGE
```
jkandasa, line 114: @jordigilh thanks for the PR, I see we use `$(params.IMAGE)` on line number 90. can you please update that entry too?

vdemeester:
/hold
The `buildah` Task is going to be maintained from [`openshift-pipelines/task-containers`](https://github.com/openshift-pipelines/task-containers), so the change would need to be there. Though I am not entirely sure it is affected there.

jordigilh:
> /hold The `buildah` Task is going to be maintained from [`openshift-pipelines/task-containers`](https://github.com/openshift-pipelines/task-containers), so the change would need to be there. Though I am not entirely sure it is affected there.

Should I also raise a PR there?
https://github.com/openshift-pipelines/task-containers/blob/66681544bedfa02e2d5d37075f1e2150f23809d1/scripts/buildah-bud.sh#L89-L92
The issue with the long name could also be found there.

cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
jordigilh: Done

jordigilh:
> > > /hold The `buildah` Task is going to be maintained from [`openshift-pipelines/task-containers`](https://github.com/openshift-pipelines/task-containers), so the change would need to be there. Though I am not entirely sure it is affected there.
> > 
> > 
> > Should I also raise a PR there? https://github.com/openshift-pipelines/task-containers/blob/66681544bedfa02e2d5d37075f1e2150f23809d1/scripts/buildah-bud.sh#L89-L92 The issue with the long name could also be found there.
> 
> ah superb 👼🏼 yeah that would be perfect @jordigilh 🤗

I stand corrected. It's not needed because it's a script itself and the parameters are passed as environment values directly, so the problem I faced here won't happen there. I got carried away when I saw a similar structure to the `push` command. Sorry for the confusion 🙏 

```bash
_buildah push ${PARAMS_PUSH_EXTRA_ARGS} \
    --digestfile="${digest_file}" \
    ${PARAMS_IMAGE} \
    docker://${PARAMS_IMAGE}
```

cmd/openshift/operator/kodata/tekton-addon/addons/02-clustertasks/source_local/buildah/buildah-task.yaml
```diff
@@ -79,15 +79,15 @@ spec:
     workingDir: $(workspaces.source.path)
     script: |
       ENTITLEMENT_VOLUME=""
-
+      IMAGE=$(params.IMAGE)
```
vdemeester, line 82: We could even pass this as an `env` to reduce "attack" vector

vdemeester (approved):

piyush-garg:
> > > > /hold The `buildah` Task is going to be maintained from [`openshift-pipelines/task-containers`](https://github.com/openshift-pipelines/task-containers), so the change would need to be there. Though I am not entirely sure it is affected there.
> > > 
> > > 
> > > Should I also raise a PR there? https://github.com/openshift-pipelines/task-containers/blob/66681544bedfa02e2d5d37075f1e2150f23809d1/scripts/buildah-bud.sh#L89-L92 The issue with the long name could also be found there.
> > 
> > 
> > ah superb 👼🏼 yeah that would be perfect @jordigilh 🤗
> 
> I stand corrected. It's not needed because it's a script itself and the parameters are passed as environment values directly, so the problem I faced here won't happen there. I got carried away when I saw a similar structure to the `push` command. Sorry for the confusion 🙏
> 
> ```shell
> _buildah push ${PARAMS_PUSH_EXTRA_ARGS} \
>     --digestfile="${digest_file}" \
>     ${PARAMS_IMAGE} \
>     docker://${PARAMS_IMAGE}
> ```

as the clustertask will be deprecated and removed soon, we are not adding any changes, and as you mentioned tasks are not affected and as they are already shipped, we are not going with this change. Thanks for the efforts @jordigilh

PR #2085, opened 2024-03-27
[main] ignore unknown fields on webhook call

jkandasa:
# Changes

* over a time, we removed some of the fields from the CR, If a user upgrades their tekton operator from a very old version to most recent version, there is a huge chance to face unknown field error on a webhook validation. To resolve these kind of issues, we can allow unknown fields. I do not see any issue on allowing unknown fields. Even though, if we allow the unknown fields, the unknown fields are not stored on the CR. Those are all removed on the fly.

##### Sample error:
```
admission webhook "webhook.operator.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "disable-working-directory-overwrite"
```

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -64,8 +64,8 @@ func NewDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher
 			return ctx
 		},
 
-		// Whether to disallow unknown fields.
-		true,
+		// allow unknown fields
+		false,
 	)
 }
 
@@ -86,8 +86,8 @@ func NewValidationAdmissionController(ctx context.Context, cmw configmap.Watcher
 			return ctx
 		},
 
-		// Whether to disallow unknown fields.
-		true,
+		// allow unknown fields
+		false,
 	)
 }
 


vdemeester (commented):
We are ok that this will mean a user can make a typo and the webhook won't disallow it ? 
This is done so that if we remove a field, we don't fail on upgrade ? If that's the case, maybe we should aim for not removing any field (but make them ignored) maybe ?

This is also probably due to the fact that we are still serving v1alpha1 for the operator's CRDs

jkandasa:
@vdemeester 
>  This is done so that if we remove a field, we don't fail on upgrade ? If that's the case, maybe we should aim for not removing any field (but make them ignored) maybe ?

This looks good to me. Shall I close this PR?

vdemeester:
> @vdemeester
> 
> > This is done so that if we remove a field, we don't fail on upgrade ? If that's the case, maybe we should aim for not removing any field (but make them ignored) maybe ?
> 
> This looks good to me. Shall I close this PR?

I think so yes 👼🏼

PR #2118, opened 2024-04-18
Put back the namespace to reconcile list if a resource cant be created for…

jkhelil:
… this namespace

# Changes

Put back the namespace to the reconcile list if a resource creation is failing for this namespace, so we continue reconciling even if an error happens when creating a resource one namespace.

# Submitter Checklist

review them:

# Release Notes

```release-note
Put back the namespace to the reconcile list if a resource creation is failing for this namespace, so we continue reconciling even if an error happens when creating a resource one namespace.
```

### DIFF
--- a/pkg/reconciler/openshift/tektonconfig/rbac.go
+++ b/pkg/reconciler/openshift/tektonconfig/rbac.go
@@ -280,6 +280,18 @@ func (r *rbac) getNamespacesToBeReconciled(ctx context.Context) ([]corev1.Namesp
 	return namespaces, nil
 }
 
+// AddNsToBeReconciledOnError removes label openshift-pipelines.tekton.dev/namespace-reconcile-version: <release-version>
+// on a given namespace from namespace labels to force the namepace to be reconciled again
+func (r *rbac) AddNsToReconcileOnError(ctx context.Context, ns *corev1.Namespace) error {
+	labels := ns.GetLabels()
+	delete(labels, namespaceVersionLabel)
+	ns.SetLabels(labels)
+	if _, err := r.kubeClientSet.CoreV1().Namespaces().Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
+		return fmt.Errorf("failed to update namespace %s: %v", ns.Name, err)
+	}
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


jkhelil:
@jkandasa @piyush-garg can i have a first rview here please ?
thank you

pkg/reconciler/openshift/tektonconfig/rbac.go
```diff
@@ -396,35 +408,57 @@ func (r *rbac) createResources(ctx context.Context) error {
 	for _, ns := range namespacesToBeReconciled {
 		logger.Infow("Inject CA bundle configmap in ", "Namespace", ns.GetName())
 		if err := r.ensureCABundles(ctx, &ns); err != nil {
-			return fmt.Errorf("failed to ensure ca bundles presence in namespace %s, %w", ns.Name, err)
+			logger.Errorf("failed to ensure ca bundles presence in namespace %s: %v", ns.Name, err)
+			if err = r.AddNsToReconcileOnError(ctx, &ns); err != nil {
+				return err
```
jkandasa, line 413: @jkhelil When we return from the loop on error on a specific namespace issue, we fail to reconcile further namespaces until we fix the error on the specific namespace. So IMHO, we can continue to proceed with further namespaces, even though we see error on a namespace.

jkhelil: hey @jkandasa thats why we continue juste after the check. here we return from AddNsToReconcileOnError wich is removing the label on the namespace on etcd. so if something is failing there, we dont continue otherwise the namespace will be considered as reconciled without having the prerequisite resources.

jkandasa: >  we return from AddNsToReconcileOnError wich is removing the label on the namespace on etcd. so if something is failing there, we dont continue

@jkhelil I meant this

jkandasa (approved):

pkg/reconciler/openshift/tektonconfig/rbac.go
piyush-garg: I did not get this logic, we are not adding the label, untill we are reaching the end of this function, which means everything is created.

I think what we need is more checks whether the namespace needs to be reonciled

PR #2145, opened 2024-05-09
Bound SA Token to Tekton Result API server in OpenShift

khrm:
This enables workload identity federation.

# Changes

# Submitter Checklist

review them:

# Release Notes


khrm (commented):
/kind misc

jkandasa (approved):

khrm:
@savitaashture Probably, we can get this merge. Can you review this?

savitaashture:
LGTM

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -27,13 +27,21 @@ import (
...
 )
 
 const (
 	// manifests console plugin yaml directory location
 	routeRBACYamlDirectory  = "static/tekton-results/route-rbac"
 	internalDBYamlDirectory = "static/tekton-results/internal-db"
+	deploymentAPI           = "tekton-results-api"
+	apiContainerName        = "api"
+	boundSAVolume           = "bound-sa-token"
+	boundSAPath             = "/var/run/secrets/openshift/serviceaccount"
```
savitaashture, line 44: Is this the default path which exist on all the openshift clusters?
Any doc reference to show about this path usage ?

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -141,3 +150,79 @@ func filterAndTransform() client.FilterAndTransform {
 		return manifest, nil
 	}
 }
+
+// injectBoundSAToken adds a sa token projected volume to the Results Deployment
+func injectBoundSAToken(comp v1alpha1.TektonComponent) mf.Transformer {
```
savitaashture, line 155: can we add it in the doc about this injection of SA ?

khrm: It's being mentioned in release note. For now it's enough. We planned to update install section of Results. We can add then.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: `The path to the service account token inside the pod. By convention, this is /var/run/secrets/openshift/serviceaccount/token for OpenShift Container Platform components.`

https://docs.openshift.com/container-platform/4.11/authentication/managing_cloud_provider_credentials/cco-mode-sts.html

Same in Google GCS.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: TBH, it doesn't change have any UX change for end user, that's why I didn't add it.

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -76,6 +84,7 @@ func (oe openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Tr
 		occommon.RemoveRunAsUser(),
 		occommon.RemoveRunAsGroup(),
 		occommon.ApplyCABundles,
+		injectBoundSAToken(comp),
```
piyush-garg, line 87: i think we should get the value of field required for this before and only pass the field required in transformer function

piyush-garg: or add this function based on the condition

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -27,13 +27,21 @@ import (
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/pkg/logging"
 )
 
 const (
 	// manifests console plugin yaml directory location
 	routeRBACYamlDirectory  = "static/tekton-results/route-rbac"
 	internalDBYamlDirectory = "static/tekton-results/internal-db"
+	deploymentAPI           = "tekton-results-api"
```
piyush-garg, line 41: this does not seems to be used

khrm: I think I removed it by mistake. It was used earlier. Let me re-add it in the condition.

pkg/reconciler/openshift/tektonresult/extension.go
piyush-garg, line 155: this will be added for all the cases. Can there will be cases where users does not want to do this?

khrm: Yes, it will be added for all users. It's a conscious decision on my part.
It doesn't matter whether the user requires this or not. This is something which is used when we have WIF. Even if it's not WIF, then we are just getting a SA token mounted.
We just don't allow removing this token which is provided by OpenShift in all clusters. Or we can have just a configuration to enable or disable this but every configuration adds overhead towards default installation.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: We already have a condition in injectBoundSAToken and we would need it anyway to apply only to a particular deployment. I feel having a condition introduces a bad coupling between functions.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: I have added it back in the condition.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: So I am passing properties now.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: We support only AWS or Google. So both have the same convention for `OpenShift Container Platform components`. 
I tested this in a CU cluster.

We will need to pass properties and modify when we add Azure but we don't support Azure anyway.

khrm:
@piyush-garg Please re-review.

pkg/reconciler/openshift/tektonresult/extension.go
khrm: https://docs.openshift.com/container-platform/4.11/authentication/managing_cloud_provider_credentials/cco-mode-sts.html

khrm: In this case results API server requires s3 or GCS bucket for logging.

jkandasa: thanks @piyush-garg @khrm 
I am fine with this changes

piyush-garg: as discussed on chat, this will not impact and results should work fine for both type of usersm using WIF or not

PR #2164, opened 2024-05-21
Move clustertask and clustertriggerbinding role to component

piyush-garg:
This will move clustertask and clustertriggerbinding view permissions for all system authenticated users on openshift to be installed along with component installation instead of doing it with addon resources.

This will help users to move to basic or lite profile without permission issues, also if user disable default clustertasks installation then also permission issue will not happen

Also clustertask permissions were getting created as part of all installersets like custom and versioned installerset, and later in new versioned installerset also if upgraded. This was creating race between the different installerset reconiler to update resource because of the different installerset name was getting added to ownerReference resulting in different yaml and sha value, so this will also fix that as we moved the permissions out of addon installerset and there will be single instance. I have changed the name of resources to not conflict with old name and we can users to delete old clustertask installerset if they are not using old clustertasks

# Submitter Checklist

review them:

# Release Notes

```release-note
Move clustertask and clustertriggerbinding role to component
```


savitaashture:
Should we do this for https://github.com/tektoncd/operator/tree/main/cmd/openshift/operator/kodata/tekton-addon/addons/07-ecosystem ?

jkandasa (approved):
LGTM

piyush-garg:
> Should we do this for https://github.com/tektoncd/operator/tree/main/cmd/openshift/operator/kodata/tekton-addon/addons/07-ecosystem ?

these roles are specific to getting task only in openshift-pipelines, while the clustertask and clustertriggerbinding are different case as these are clustertwide resources and end up in breaking profiles, UI etc

> 
> > I have changed the name of resources to not conflict with old name and we can users to delete old clustertask installerset if they are not using old clustertasks
> 
> Should we indicate this in RN ? and is this a breaking change for existing users ?

changing the name of the role/rolebinding which is not exposed to user, should not be a breaking change for users

PuneetPunamiya:
Great findings 🙌🏻
/lgtm

PR #2179, opened 2024-05-30
ArtifactsPipelineRunEnableDeepInspection making this field to accept both bool and string

rupalibehera:
# Changes
https://issues.redhat.com/browse/SRVKP-4387

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonchain_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonchain_types.go
@@ -80,10 +80,10 @@ type ChainProperties struct {
 	ArtifactsTaskRunSigner  string  `json:"artifacts.taskrun.signer,omitempty"`
 
 	// pipelinerun artifacts config
-	ArtifactsPipelineRunFormat               string  `json:"artifacts.pipelinerun.format,omitempty"`
-	ArtifactsPipelineRunStorage              *string `json:"artifacts.pipelinerun.storage,omitempty"`
-	ArtifactsPipelineRunSigner               string  `json:"artifacts.pipelinerun.signer,omitempty"`
-	ArtifactsPipelineRunEnableDeepInspection *bool   `json:"artifacts.pipelinerun.enable-deep-inspection,omitempty"`
+	ArtifactsPipelineRunFormat               string    `json:"artifacts.pipelinerun.format,omitempty"`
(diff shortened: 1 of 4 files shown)


rupalibehera:
@PuneetPunamiya does this make sense ?

rupalibehera:
I have tested this with upgrading and it is working fine.

jkandasa:
@rupalibehera can you please rebase 2 commits to one?

jkandasa (approved):

piyush-garg:
@rupalibehera @PuneetPunamiya can you please give detail about why this change in required in commit message, and also is it required for all the three fields?

rupalibehera:
> @rupalibehera can you please rebase 2 commits to one?

done

piyush-garg:
Also can we add some unit tests for this?

rupalibehera:
> @rupalibehera @PuneetPunamiya can you please give detail about why this change in required in commit message, and also is it required for all the three fields?

some boolean fields in tekton config for chains were string and some were boolean, while adding those it was like a guess work if it should be bool or string hence I have updated it, it was required only for one field https://issues.redhat.com/browse/SRVKP-4387 but I added for all others as well

rupalibehera:
> Also can we add some unit tests for this?

we do not have any existing unit tests for the same

piyush-garg:
> > @rupalibehera @PuneetPunamiya can you please give detail about why this change in required in commit message, and also is it required for all the three fields?
> 
> some boolean fields in tekton config for chains were string and some were boolean, while adding those it was like a guess work if it should be bool or string hence I have updated it, it was required only for one field https://issues.redhat.com/browse/SRVKP-4387 but I added for all others as well

AFAIK, we do this when a field which was supposed to be boolean before, now is string, so to support upgrade, which change it like this to support both type/values. If this is the case here, please do this for the field which has got this change. Also, we should add validation defaulting etc accordingly. Also, good to have unit tests if you can add.

If this is not the case, and there is some other use case for this, can you please give details for that. Also, would be great if you can put all those details in commit msg and PR

rupalibehera:
@piyush-garg @jkandasa I am just changing ArtifactsPipelineRunEnableDeepInspection field now and dropped the other fields.

rupalibehera:
@piyush-garg @jkandasa , can we please merge this without the unit tests, there are very minimal changes now

PR #2275, opened 2024-08-06
Add default config properties to chains configMap

jkhelil:
# Changes
- Add chains option to allow users to opt-in to generate a chains signing secret and generate and ecdsa key pair as signing secret
- Addition of Sane Default Configuration Properties:
This update introduces sensible default configuration properties to the Chains controller, enhancing its initial setup and performance.
- Storage Backend Update:
Previously, the Chains controller used Tekton storage, where payloads and signatures were stored in the annotations of Tekton objects (such as TaskRuns and PipelineRuns). This approach significantly increased the size of these objects. To address this issue, this pull request changes the default storage backend to OCI, thereby reducing the object size and improving overall efficiency.

By default chains controller uses storage type tekton for its differents added annotation

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/docs/TektonChain.md
+++ b/docs/TektonChain.md
@@ -45,6 +45,14 @@ It is recommended to install the component through [TektonConfig](./TektonConfig
 ## Chain Config
 
 There are some fields which you can define on Tekton Chains CR to configure the behaviour of the chains
+If nothing is set, the operator sets these default properties:
+  artifacts.taskrun.format: in-toto
+  artifacts.taskrun.storage: oci
+  artifacts.oci.storage: oci
+  artifacts.oci.format: simplesigning
+  artifacts.pipelinerun.format: in-toto
+  artifacts.pipelinerun.storage: oci
+
 
 ### Properties (Mandatory)
 
--- a/docs/TektonChain.md
+++ b/docs/TektonChain.md
@@ -71,7 +71,6 @@ values then those will be set for the particular field.
 Details of the field can be found in [Tekton Chains Config][chains-config]
 
 Chains CR will look like this after providing all the fields.
-
 ```yaml
 apiVersion: operator.tekton.dev/v1alpha1
 kind: TektonChain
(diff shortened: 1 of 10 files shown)


docs/TektonChain.md
```diff
@@ -45,6 +45,14 @@ It is recommended to install the component through [TektonConfig](./TektonConfig
...
 
 There are some fields which you can define on Tekton Chains CR to configure the behaviour of the chains
+If nothing is set, the operator sets these default properties:
+  artifacts.taskrun.format: in-toto
+  artifacts.taskrun.storage: oci
+  artifacts.oci.storage: oci
+  artifacts.oci.format: simplesigning
+  artifacts.pipelinerun.format: in-toto
+  artifacts.pipelinerun.storage: oci
+
```
PuneetPunamiya, line 55: I think we should have this default values which are defined here https://github.com/tektoncd/chains/blob/main/config/100-deployment.yaml#L45-L52

Since we are creating x509 secret keys, does it makes sense to add the default signer as x509 ?

docs/TektonConfig.md
```diff
@@ -248,6 +248,7 @@ Example:
 chain:
   disabled: false                  # - `disabled` : if the value set as `true`, chains feature will be disabled (default: `false`)
   targetNamespace: tekton-pipelines
+  generateSigningSecret: true # default value: false
```
PuneetPunamiya, line 251: Are we handling the scenario where if we change the user change the field value from `true` to `false` `generateSigningSecret` in that case are we deleting the key/values from the secret ?

PuneetPunamiya, line 251: After setting this value, signing-secrets were not updated

pkg/reconciler/kubernetes/tektonchain/transform.go
```diff
@@ -23,8 +23,18 @@ import (
...
 )
 
+var defaultChainProperties v1alpha1.ChainProperties = v1alpha1.ChainProperties{
+	ArtifactsTaskRunFormat:      "in-toto",
+	ArtifactsTaskRunStorage:     ptr.String("oci"),
+	ArtifactsPipelineRunFormat:  "in-toto",
+	ArtifactsPipelineRunStorage: ptr.String("oci"),
+	ArtifactsOCIFormat:          "simplesigning",
+	ArtifactsOCIStorage:         ptr.String("oci"),
+}
```
PuneetPunamiya, line 36: I see this values are added in the chains-config configMap, but since TektonConfig is the source of truth, I think these values should be reflected there as well

docs/TektonConfig.md
jkhelil: No actually, it doesnt update the signing key. I dont think we want to do so,the property itself generates a signing key, the way operator works with installerset is that
- if generateSigningSecret is true, the installerset is created with the secret data(in the yaml manifest)
- if generateSigningSecret is false, the installerset is created with an empty secret data. Now if the installerset is already there, it is not updated. this is common function for all the component handled by the operator, the user should remove the installerset to recreate an empty one.

jkhelil: You need to delete the chains secret installerset. This is the way the operator handles the objects that are generated by the installersets 
@piyush-garg can you confirm please, that in order to update an object after a config change, we have to delete the installerset, in order to the reconciler to restart it is creation operation

pkg/reconciler/kubernetes/tektonchain/transform.go
jkhelil: You mean by default create a tektonConfig with these default, i think that make sens

docs/TektonConfig.md
PuneetPunamiya: I think reconciler should take care of that and user shouldn't have to go and delete the installerset

pkg/reconciler/kubernetes/tektonchain/transform.go
PuneetPunamiya: Yeah, since for user, TektonConfig is the source of truth

PR #2299, opened 2024-09-02
Add default config properties to chains configMap

jkhelil:
# Changes
- Add chains option to allow users to opt-in to generate a chains signing secret and generate and ecdsa key pair as signing secret
- Addition of Sane Default Configuration Properties:
This update introduces sensible default configuration properties to the Chains controller, enhancing its initial setup and performance.
- Storage Backend Update:
Previously, the Chains controller used Tekton storage, where payloads and signatures were stored in the annotations of Tekton objects (such as TaskRuns and PipelineRuns). This approach significantly increased the size of these objects. To address this issue, this pull request changes the default storage backend to OCI, thereby reducing the object size and improving overall efficiency.
By default chains controller uses storage type tekton for its differents added annotation

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/docs/TektonChain.md
+++ b/docs/TektonChain.md
@@ -45,6 +45,14 @@ It is recommended to install the component through [TektonConfig](./TektonConfig
 ## Chain Config
 
 There are some fields which you can define on Tekton Chains CR to configure the behaviour of the chains
+If nothing is set, the operator sets these default properties:
+  artifacts.taskrun.format: in-toto
+  artifacts.taskrun.storage: oci
+  artifacts.oci.storage: oci
+  artifacts.oci.format: simplesigning
+  artifacts.pipelinerun.format: in-toto
+  artifacts.pipelinerun.storage: oci
+
 
 ### Properties (Mandatory)
 
--- a/docs/TektonChain.md
+++ b/docs/TektonChain.md
@@ -71,7 +71,6 @@ values then those will be set for the particular field.
 Details of the field can be found in [Tekton Chains Config][chains-config]
 
 Chains CR will look like this after providing all the fields.
-
 ```yaml
 apiVersion: operator.tekton.dev/v1alpha1
 kind: TektonChain
@@ -80,17 +79,18 @@ metadata:
 spec:
   disabled: false
   targetNamespace: tekton-pipelines
+  generateSigningSecret: true # default value: false
   controllerEnvs:
     - name: MONGO_SERVER_URL      # This is the only field supported at the moment which is optional and when added by user, it is added as env to Chains controller
       value: #value               # This can be provided same as env field of container
-  artifacts.taskrun.format: in-toto
-  artifacts.taskrun.storage: tekton,oci (comma separated values)
+  artifacts.taskrun.format: in-toto # default value: in-toto
+  artifacts.taskrun.storage: tekton,oci (comma separated values) # default value: oci
   artifacts.taskrun.signer: x509
   artifacts.oci.storage: oci (comma separated values)
   artifacts.oci.format: simplesigning
   artifacts.oci.signer: x509
-  artifacts.pipelinerun.format: in-toto
-  artifacts.pipelinerun.storage: tekton,oci (comma separated values)
+  artifacts.pipelinerun.format: in-toto # default value: in-toto
+  artifacts.pipelinerun.storage: tekton,oci (comma separated values) # default value: oci
   artifacts.pipelinerun.signer: x509
   artifacts.pipelinerun.enable-deep-inspection: #value (boolean - true/false)
   storage.gcs.bucket: #value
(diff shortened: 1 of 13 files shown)


jkhelil:
@PuneetPunamiya got my old PR closed, changed the branch name
Anyway, I took in account your  useful comments, and did fixes here
Now whenever we change the geenratedSigningSecret input, it deletes the installerset
However i am still seeing an issue, with tektonconfig defaults, those are lending correctly on the chains configmap, but once i update one of them in tektonconfig, they dont get updated on chains configmap. Wondering if this is correct place where we should put them, @piyush-garg ?

pkg/apis/operator/v1alpha1/tektonconfig_defaults.go
```diff
@@ -94,4 +94,24 @@ func (tc *TektonConfig) SetDefaults(ctx context.Context) {
...
+	}
+	if tc.Spec.Chain.ArtifactsPipelineRunStorage == nil {
+		tc.Spec.Chain.ArtifactsPipelineRunStorage = ptr.String("oci")
+	}
+	if tc.Spec.Chain.ArtifactsOCIFormat == "" {
+		tc.Spec.Chain.ArtifactsOCIFormat = "simplesigning"
+	}
+	if tc.Spec.Chain.ArtifactsOCIStorage == nil {
+		tc.Spec.Chain.ArtifactsOCIStorage = ptr.String("oci")
+	}
```
jkandasa, line 116: @jkhelil these defaulting will not impact existing chains CR, 
However, if user updates a field on a TektonConfig CR, not related to chains also changes the chains behavior.
Will that impact the existing chains instance?

pkg/reconciler/common/transformers.go
```diff
@@ -940,3 +940,49 @@ func ReplaceNamespace(newNamespace string) mf.Transformer {
...
+
+		// Convert unstructured to Secret
+		secret := &corev1.Secret{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, secret)
+		if err != nil {
+			return err
+		}
+
+		// Update the Secret's data only if it is nil or empty
+		if secret.Data == nil || len(secret.Data) == 0 {
```
jkandasa, line 965: @jkhelil I believe `len` should return `0` on `nil` value too. If so, we can remove the `nil` check

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -46,6 +52,9 @@ const (
 	chainsContainerName = "tekton-chains-controller"
 	// Deployment Name
 	chainsDeploymentName = "tekton-chains-controller"
+
+	// secret installer set additional Annotation
+	SecretTISSigningAnnotation = "operator.tekton.dev/generated-signing-secret"
```
jkandasa, line 57: if we do not use `SecretTISSigningAnnotation` outside package, we can make it `private`

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -469,6 +478,27 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonChain
 		}
 		return nil
 	}
+	// if generatedSigningSecret has been changed for chainsCR, then update the Tekton Chain Secret InstallerSet
+	secretInstallerSetSigningKey, _ := strconv.ParseBool(installedSecretTIS.Annotations[SecretTISSigningAnnotation])
```
jkandasa, line 482: Does `SecretTISSigningAnnotation` present always in the annotations? on the existing running instance too?

jkandasa (commented):
@jkhelil Please confirm the upgrade happens smoothly from the previous version of chains

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
jkhelil: No it is added if the secret is generated

pkg/apis/operator/v1alpha1/tektonconfig_defaults.go
jkhelil: yes i guess, it is going to set defaults, but thats what we expect, we do want to set these defaults

piyush-garg: this defaulting should be done in chains CR, and from here we should be just calling that

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
jkandasa: If the annotation `secretTISSigningAnnotation` not present, return empty string ``, that leads bool parse error,
```
false strconv.ParseBool: parsing "": invalid syntax
```

pkg/apis/operator/v1alpha1/tektonchain_defaults_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/test/diff"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/ptr"
+)
+
+func test_SetDefaults_Chain(t *testing.T) {
```
jkandasa, line 29: @jkhelil can you camelCase the function name? ==> `testSetDefaultsChain`

error:
```
pkg/apis/operator/v1alpha1/tektonchain_defaults_test.go:29:6: func `test_SetDefaults_Chain` is unused (unused)
func test_SetDefaults_Chain(t *testing.T) {
```

pkg/apis/operator/v1alpha1/tektonchain_defaults_test.go
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright 2022 The Tekton Authors
```
jkandasa, line 2: year should be 2024

docs/TektonChain.md
```diff
@@ -113,6 +121,14 @@ spec:
   transparency.enabled: #value (boolean - true/false)
   transparency.url: #value
 ```
+- `disabled` : if the value set as `true`, chains feature will be disabled (default: `false`)
+- `generateSigningSecret`: If set to true, the operator will generate an ECDSA key pair (x509.pem private key and x509-pub.pem public key) and store it in the "signing-secrets" secret in the "tekton-pipelines" namespace. the secret is used by chains controller to sign tekton artifacts(tasjruns, pipelineruns). it is important to mention that:
```
jkandasa, line 125: typo: `tasjruns`

jkandasa (approved):

PR #2361, opened 2024-09-30
WIP add support for pipelines controller statefulset ordinals

jkhelil:
# Changes

# Submitter Checklist

review them:

# Release Notes


khrm (commented):
One problem I can see is that from from statefulset to deployment is broken. It's not the issue with this PR but the operator. 
https://github.com/tektoncd/operator/issues/2372

khrm (commented):
One problem I can see is that from from statefulset to deployment is broken. It's not the issue with this PR but the operator. 
https://github.com/tektoncd/operator/issues/2372

So two controllers will be competing in this case when we shift back to deployment from statefulset.

jkhelil:
> One problem I can see is that from from statefulset to deployment is broken. It's not the issue with this PR but the operator. #2372
> 
> So two controllers will be working on same events in this case when we shift back to deployment from the statefulset.

fixed now, the switch should work properly between the two modes

pkg/apis/operator/v1alpha1/tektonpipeline_types.go
```diff
@@ -201,6 +203,12 @@ type PipelinePerformanceLeaderElectionConfig struct {
 	Buckets *uint `json:"buckets,omitempty"`
 }
 
+// allow to configure pipelines controller ha mode to statefulset ordinals
+type PipelinePerformanceStatefulsetOrdinalsConfig struct {
+	//if is true, enable StatefulsetOrdinals mode
+	StatefulsetOrdinals *bool `json:"statefulset-ordinals"`
```
PuneetPunamiya, line 209:
```suggestion
	StatefulsetOrdinals *bool `json:"statefulset-ordinals, omitempty"`
```

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
```diff
@@ -150,6 +150,17 @@ func (p *Pipeline) setDefaults() {
 		p.EnableGitResolver = ptr.Bool(true)
 	}
 
+	// Statefulset Ordinals
+	// if StatefulSet Ordinals mode, buckets should be equal to replicas
+	if p.Performance.StatefulsetOrdinals == nil {
+		p.Performance.StatefulsetOrdinals = ptr.Bool(false)
```
PuneetPunamiya, line 156: I think we might not need to set this to default value false when it is nil

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
```diff
@@ -150,6 +150,17 @@ func (p *Pipeline) setDefaults() {
...
+	} else if *p.Performance.StatefulsetOrdinals {
+		if p.Performance.Replicas != nil && *p.Performance.Replicas > 1 {
```
PuneetPunamiya, line 158: If `p.Performance.Replicas != nil` then we might have to set the buckets to 1

pkg/reconciler/common/transformers.go
```diff
@@ -58,6 +59,11 @@ const (
 
 	runAsNonRootValue              = true
 	allowPrivilegedEscalationValue = false
+
+	tektonPipelinesControllerName                      = "tekton-pipelines-controller"
+	tektonPipelinesServiceName                         = "tekton-pipelines-controller"
+	tektonPipelinesControllerStatefulServiceName       = "STATEFUL_SERVICE_NAME"
+	tektonPipelinesControllerStatefulControllerOrdinal = "STATEFUL_CONTROLLER_ORDINAL"
```
PuneetPunamiya, line 66: Pipelines specific things we can keep it in pipelines reconciler and the function can take these values as parameters so that we can use it for other components in future

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
jkhelil: in statefulset mode, bucket should be equal to replicas, not to 1

pkg/reconciler/common/transformers.go
jkhelil: well, these constant are use in the transformers function in this file, so i think we should keep them here

pkg/apis/operator/v1alpha1/tektonpipeline_types.go
jkhelil: fixed

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
jkhelil: yes true i think we can omit this

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
jkandasa: @jkhelil I guess @PuneetPunamiya try to say bucket should be `1`, when `p.Performance.Replicas == nil`

PuneetPunamiya: yeah sorry for the confusion 
+1 on what @jkandasa said

pkg/reconciler/common/transformers.go
PuneetPunamiya: Yeah, that make sense, but let's say in future we want to add for other components as well, then we might not be able to reuse the transformer function

pkg/apis/operator/v1alpha1/tektonpipeline_defaults.go
jkhelil: When replicas is not specified (it's nil), the default behavior of a StatefulSet is to create a single pod,  in StatefulSet's Knative mode, the default behavior is to create one bucket, so there is no need to enforce that mode, it works by default.

However, when replicas is explicitly set (not nil), I found that if we don’t enforce the bucket assignment to equal to replicas, the election mode and StatefulSet mode can interfere with each other. This results in some pods being in election mode while others are in StatefulSet mode, leading to inconsistent behavior.

jkhelil:
@PuneetPunamiya PTAL

PuneetPunamiya:
Thanks @jkhelil 🤙🏻 
/lgtm

PR #2516, opened 2025-01-16
fix: avoid panic when reconciling namespace without owner reference

l-qing:
# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/common/common.go
+++ b/pkg/reconciler/common/common.go
@@ -43,7 +43,8 @@ func PipelineReady(informer informer.TektonPipelineInformer) (*v1alpha1.TektonPi
 		}
 		return nil, err
 	}
-	if ppln.GetStatus() != nil && strings.Contains(ppln.GetStatus().GetCondition(apis.ConditionReady).Message, v1alpha1.UpgradePending) {
+	readyCondition := ppln.GetStatus().GetCondition(apis.ConditionReady)
+	if readyCondition != nil && strings.Contains(readyCondition.Message, v1alpha1.UpgradePending) {
 		return nil, v1alpha1.DEPENDENCY_UPGRADE_PENDING_ERR
 	}
 	if !ppln.Status.IsReady() {
--- a/pkg/reconciler/common/common.go
+++ b/pkg/reconciler/common/common.go
@@ -43,14 +43,25 @@ func PipelineReady(informer informer.TektonPipelineInformer) (*v1alpha1.TektonPi
 		}
 		return nil, err
 	}
-	readyCondition := ppln.GetStatus().GetCondition(apis.ConditionReady)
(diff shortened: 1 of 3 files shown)


pkg/reconciler/common/common.go
```diff
@@ -43,7 +43,8 @@ func PipelineReady(informer informer.TektonPipelineInformer) (*v1alpha1.TektonPi
 		}
 		return nil, err
 	}
-	if ppln.GetStatus() != nil && strings.Contains(ppln.GetStatus().GetCondition(apis.ConditionReady).Message, v1alpha1.UpgradePending) {
+	readyCondition := ppln.GetStatus().GetCondition(apis.ConditionReady)
+	if readyCondition != nil && strings.Contains(readyCondition.Message, v1alpha1.UpgradePending) {
```
l-qing, line 47: In my environment, sometimes the Pipeline has not yet been reconciled, resulting in the status being empty. At this point, entering here may trigger a panic.

pkg/reconciler/common/targetnamespace.go
```diff
@@ -56,7 +56,7 @@ func ReconcileTargetNamespace(ctx context.Context, labels map[string]string, ann
 		if namespace.Name == tektonComponent.GetSpec().GetTargetNamespace() && namespace.DeletionTimestamp == nil {
 			_targetNamespace := namespace.DeepCopy()
 			targetNamespace = _targetNamespace
-		} else {
+		} else if len(namespace.GetOwnerReferences()) > 0 {
```
l-qing, line 59: I manually created the ns, which resulted in entering here, causing a panic.

I understand that we should not panic under any circumstances.

pratap0007:
I think, if you create namespace manually it adds owner reference so in which case a target namespace would not have owner reference?

pkg/reconciler/common/common.go
jkhelil: @l-qing we need to keep the check ppln.GetStatus() != nil, otherwise it will panic on ppln.GetStatus().GetCondition, and refactor as you suggest ?

l-qing: Although it's not possible for nil here, it is indeed necessary to optimize it to reduce cognitive load.

https://github.com/tektoncd/operator/blob/f547f0924df772c81e7960959615deec57e1798d/pkg/apis/operator/v1alpha1/tektonpipeline_types.go#L30-L44

l-qing:
> pkg/reconciler/common/targetnamespace.go

I'm not very sure about this.

In theory, if the initial namespace had another owner, the Tekton owner would not be added later. 
If another owner is removed owner and namespace later, is it possible to enter here?

vdemeester (approved):

PR #2610, opened 2025-02-28
Preserve the existing TektonResult CR configuration

pratap0007:
Now TektonResult is installed by the TektonConfig, TektonConfig config was creating TektonResult CR with default configuration and also not preserving the exsiting TektonResult configuration, now this patch fixes to preserve the TektonResult configuration

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Preserve the existing TektonResult CR configuration
```

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/upgrade/pre_upgrade.go
+++ b/pkg/reconciler/shared/tektonconfig/upgrade/pre_upgrade.go
@@ -103,3 +103,36 @@ func removeDeprecatedAddonParams(ctx context.Context, logger *zap.SugaredLogger,
 	_, err = operatorClient.OperatorV1alpha1().TektonConfigs().Update(ctx, tcCR, metav1.UpdateOptions{})
 	return err
 }
+
+// previous version of the Tekton Operator does not install the TektonResult via TektonConfig
+// in the new version, we are supporting to installing and manage TektonResult via TektonConfig
+// so if TektonResult CR exists on the cluster then needs to copy the TektonResult CR configuration to the TektonConfig CR
+func copyResultConfigToTektonConfig(ctx context.Context, logger *zap.SugaredLogger, k8sClient kubernetes.Interface, operatorClient versioned.Interface, restConfig *rest.Config) error {
+	// get the TekonResult CR
+	trCR, err := operatorClient.OperatorV1alpha1().TektonResults().Get(ctx, v1alpha1.ResultResourceName, metav1.GetOptions{})
+	if err != nil {
+		if apierrs.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	// get the TekonConfig CR
+	tcCR, err := operatorClient.OperatorV1alpha1().TektonConfigs().Get(ctx, v1alpha1.ConfigResourceName, metav1.GetOptions{})
+	if err != nil {
+		if apierrs.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	// copy the existing TektonResult CR configuration  to the TektonConfig CR
+	tcCR.Spec.Result.ResultsAPIProperties = trCR.Spec.ResultsAPIProperties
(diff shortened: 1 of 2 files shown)


pkg/reconciler/shared/tektonconfig/upgrade/pre_upgrade.go
```diff
@@ -103,3 +103,34 @@ func removeDeprecatedAddonParams(ctx context.Context, logger *zap.SugaredLogger,
...
+		return err
+	}
+
+	// copy the existing TektonResult CR configuration  to the TektonConfig CR
+	tcCR.Spec.Result.ResultsAPIProperties = trCR.Spec.ResultsAPIProperties
+	tcCR.Spec.Result.LokiStackProperties = trCR.Spec.LokiStackProperties
+
+	_, err = operatorClient.OperatorV1alpha1().TektonConfigs().Update(ctx, tcCR, metav1.UpdateOptions{})
+	if err == nil {
+	}
```
PuneetPunamiya, line 134:
```suggestion
	if err != nil {
	  return err 
	}
	
	return nil
```

Shall we do this instead ?

pkg/reconciler/shared/tektonconfig/upgrade/pre_upgrade.go
```diff
@@ -103,3 +103,34 @@ func removeDeprecatedAddonParams(ctx context.Context, logger *zap.SugaredLogger,
...
+
+	// get the TekonResult CR
+	trCR, err := operatorClient.OperatorV1alpha1().TektonResults().Get(ctx, v1alpha1.ResultResourceName, metav1.GetOptions{})
+	if err != nil {
+		if apierrs.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
```
PuneetPunamiya, line 126: It would be nice if we first get Tekton Result and then get Tetkton Config, reason being if Tekton Result is not installed in previous versions then in that case we can one api call in this function to get TetkonConfig ?

cc @jkhelil @pratap0007

pkg/reconciler/shared/tektonconfig/upgrade/pre_upgrade.go
```diff
@@ -103,3 +103,33 @@ func removeDeprecatedAddonParams(ctx context.Context, logger *zap.SugaredLogger,
 	_, err = operatorClient.OperatorV1alpha1().TektonConfigs().Update(ctx, tcCR, metav1.UpdateOptions{})
 	return err
 }
+
+// previous version of the Tekton Operator does not install the TektonResult via
```
savitaashture, line 107:
```suggestion
// previous version of the Tekton Operator does not install the TektonResult via TektonConfig
```

It would be good if you cover the unit test

but yeah you do it in a followup PR

pkg/reconciler/shared/tektonconfig/upgrade/pre_upgrade.go
```diff
@@ -103,3 +103,33 @@ func removeDeprecatedAddonParams(ctx context.Context, logger *zap.SugaredLogger,
...
+		}
+		return err
+	}
+
+	// copy the existing TektonResult CR configuration  to the TektonConfig CR
+	tcCR.Spec.Result.ResultsAPIProperties = trCR.Spec.ResultsAPIProperties
+	tcCR.Spec.Result.LokiStackProperties = trCR.Spec.LokiStackProperties
+
+	_, err = operatorClient.OperatorV1alpha1().TektonConfigs().Update(ctx, tcCR, metav1.UpdateOptions{})
+	return err
```
PuneetPunamiya, line 134: It would be nice to check the error here as well if not nil then return error

savitaashture: I don't think its required because this is the last statement of the function so when you return err it is returning actual error if there are any or nil 

so we don't need to add extra check

But yeah as you have updated that's fine

savitaashture:
LGTM

PR #2635, opened 2025-03-14
Set kubernetes min version

jkhelil:
# Changes
- Fixes typo: use assertNoError instead of assertNoEror.
- Propagates the KUBERNETES_MIN_VERSION environment variable to component deployments if set at the operator pod level, using a transformer function.

fixes https://github.com/tektoncd/pipeline/issues/8529

# Submitter Checklist

review them:

# Release Notes

### DIFF
--- a/config/openshift/base/operator.yaml
+++ b/config/openshift/base/operator.yaml
@@ -48,6 +48,8 @@ spec:
             drop:
               - ALL
         env:
+        - name: KUBERNETES_MIN_VERSION
+          value: "v1.0"
         - name: SYSTEM_NAMESPACE
           valueFrom:
             fieldRef:
--- a/pkg/reconciler/common/transformer_inject_envar_test.go
+++ b/pkg/reconciler/common/transformer_inject_envar_test.go
@@ -1,0 +1,137 @@
+/*
+Copyright 2025 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package common
+
+import (
+	"path"
+	"testing"
+
+	mf "github.com/manifestival/manifestival"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestDeploymentEnvVars(t *testing.T) {
+	t.Run("inject environment variables into deployments", func(t *testing.T) {
+		envVars := []corev1.EnvVar{
+			{
+				Name:  "KUBERNETES_MIN_VERSION",
+				Value: "v1.0",
+			},
+			{
+				Name: "SECRET_VAR",
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: "test-secret",
+						},
+						Key: "secret-key",
+					},
+				},
+			},
+		}
+
+		testData := path.Join("testdata", "test-inject-envvar.yaml")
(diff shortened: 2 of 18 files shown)


pkg/reconciler/common/transformer_inject_envvar.go
```diff
@@ -0,0 +1,87 @@
...
+	if minVersion, exists := os.LookupEnv(KUBERNETES_MIN_VERSION_ENV_KEY); exists {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  KUBERNETES_MIN_VERSION_ENV_KEY,
+			Value: minVersion,
+		})
+	}
+	return DeploymentEnvVars(envVars)
+}
+
+func DeploymentEnvVars(envVars []corev1.EnvVar) mf.Transformer {
```
vdemeester, line 44: It probably doesn't need to be exported does it ?

vdemeester (approved):
One question, otherwise looks good.

pkg/reconciler/common/transformer_inject_envvar.go
```diff
@@ -0,0 +1,87 @@
...
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "Deployment" {
+			return nil
+		}
```
aThorp96, line 48: Since the remote-resolvers and pipelines-controller both can run as StatefulSets, do we need to support `u.GetKind() == "StatefulSet"` here or would that belong in a separate function like `StatefulSetEnvVars()`?

pkg/reconciler/common/transformer_inject_envvar.go
```diff
@@ -0,0 +1,87 @@
...
+
+	mf "github.com/manifestival/manifestival"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	KUBERNETES_MIN_VERSION_ENV_KEY = "KUBERNETES_MIN_VERSION"
```
aThorp96, line 30: Nitpick: For consistency and self-documentation, can we use the const [exported from](https://github.com/knative/pkg/blob/b7bbf4be5dbd9a8b8b60e06094bd8930cf241357/version/version.go#L31) `knative.dev/pkg/version`?

pkg/reconciler/common/transformer_inject_envvar.go
jkhelil: Well, in. theory, yes, but until now we assumed all components are deployd as deployment. Statefulsets are kind of transformers that comes at the end of all other transformations, to adress your point we may need to review all the transformers functions which means adding a bench of tests unit and e2e tests, but probably it is useless for now.

pkg/reconciler/common/transformer_inject_envar_test.go
```diff
@@ -0,0 +1,137 @@
...
+				Name: "SECRET_VAR",
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: "test-secret",
+						},
+						Key: "secret-key",
+					},
+				},
+			},
```
savitaashture, line 44: Naive query 
this extra env is to check how env's are getting added when there are more than one ??

config/openshift/base/operator.yaml
```diff
@@ -48,6 +48,8 @@ spec:
             drop:
               - ALL
         env:
+        - name: KUBERNETES_MIN_VERSION
+          value: "v1.0"
```
PuneetPunamiya, line 52: Should we also for k8s specific platform https://github.com/tektoncd/operator/blob/main/config/kubernetes/base/operator.yaml

PuneetPunamiya (commented):
I wander if we need this pr, can't we achieve this through options ?

pkg/reconciler/common/transformer_inject_envar_test.go
jkhelil: yes

config/openshift/base/operator.yaml
jkhelil: I think there is a bit a consensus to not do that upstream for now,  I would say

jkhelil:
> I wander if we need this pr, can't we achieve this through options ?

We want to force this min version, for openshift operators, so we cant let it optional

PR #2657, opened 2025-03-25
Update tekton task versions using a gh workflow that internally runs a script : Fix SRVKP-7119

infernus01:
# Changes

Added a script at path hack/openshift/update-tekton-task-versions.sh and a gh workflow at path .github/workflows/update-tektoncd-task-versions.yml that runs weekly that internally runs the script which modifies the fetch-tektoncd-catalog-tasks.sh file (TEKTON_ECOSYSTEM_TASKS field) whenever a newly added version directory is found under the p branch of the openshift-pipleines/tektoncd-catalog repo at path tasks/task-*.


jkhelil:
@infernus01 Is there any inconvenience in running this on a daily basis?

.github/workflows/update-tektoncd-task-versions.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Update Task Versions
+        id: update
+        run: |
+          chmod +x hack/openshift/update-tekton-task-versions.sh
+          ./hack/openshift/update-tekton-task-versions.sh \
+            --catalog-path ../tektoncd-catalog \
+            --script-path hack/openshift/fetch-tektoncd-catalog-tasks.sh
+          echo "has_changes=$(git diff --name-only | grep -q fetch-tektoncd-catalog-tasks.sh && echo 'true' || echo 'false')" >> $GITHUB_OUTPUT
+
+      # - name: Set up Git config
```
jkhelil, line 34: do we need the comment ?

infernus01:
And yes we could run this on daily basis as well, I was earlier worried about those unnecessary PRs it would create, and I guess, have handled that in the commit changes step.

hack/openshift/update-tekton-task-versions.sh
```diff
@@ -0,0 +1,54 @@
...
+    latest_version=$(find "$resource_dir" -maxdepth 1 -type d | grep -v "^$resource_dir$" | sort -V | tail -n 1 | xargs basename)
+
+    if [[ -z "$latest_version" ]]; then
+      echo "Skipping $resource_name (no versions found)"
+      continue
+    fi
+
+    # Update the script with the latest version
+    # Using perl for cross-platform compatibility (works on both Linux and macOS)
+    perl -i -pe "s|(\\['$resource_name'\\]=)['\"][^'\"]*['\"]|\1'$latest_version'|" "$FETCH_SCRIPT"
```
PuneetPunamiya, line 38: Not sure though just had a thought, can it be achieved through `sed`

hack/openshift/update-tekton-task-versions.sh
```diff
@@ -0,0 +1,54 @@
...
+
+    echo "Updated $resource_name to version $latest_version"
```
PuneetPunamiya, line 40: Even if there is no update in the version we still log that the version for that resource is updated

PuneetPunamiya: It would be nice if we can have a logic where we log only the resources whose versions have been changed

hack/openshift/update-tekton-task-versions.sh
infernus01: Yes, we can use `sed`. The only thing to keep in mind is that `sed -i` works differently on macOS and Linux. macOS requires a backup file, while Linux doesn’t.
I can update the script to use `sed` in a way that works on both. Would you prefer always creating a backup file (which works everywhere), or should I add a check for macOS and handle it separately?

However, `perl` worked on both without any extra checks.

hack/openshift/update-tekton-task-versions.sh
infernus01: Good catch! I’ll update the script so it only logs when a version actually changes. Thanks for pointing it out!

.github/workflows/update-tektoncd-task-versions.yml
```diff
@@ -0,0 +1,66 @@
...
+          fetch-depth: 0
+
+      - name: Clone TektonCD Catalog (p branch)
+        run: |
+          git clone --branch p https://github.com/openshift-pipelines/tektoncd-catalog.git ../tektoncd-catalog
+
+      - name: Update Task Versions
+        id: update
+        run: |
+          chmod +x hack/openshift/update-tekton-task-versions.sh
```
aThorp96, line 28: Any reason we don't have the script checked in with the executable permission?

.github/workflows/update-tektoncd-task-versions.yml
```diff
@@ -0,0 +1,66 @@
...
+          ./hack/openshift/update-tekton-task-versions.sh \
```
aThorp96, line 29: Is this script openshift specific? If so, we should probably indicate this in the name.

.github/workflows/update-tektoncd-task-versions.yml
```diff
@@ -0,0 +1,66 @@
...
+          author: Shubham Bhardwaj <shubbhar@redhat.com>
+          title: "Update Tekton task versions"
+          body: |
+            This PR automatically updates the task versions in the `fetch-tektoncd-catalog-tasks.sh` script.
+
+            The versions are pulled from the latest available versions in the [tektoncd-catalog](https://github.com/openshift-pipelines/tektoncd-catalog) repository.
+          branch: update-tekton-versions
+          base: main
+          labels: automation
+          delete-branch: true
```
aThorp96, line 66: blocking: We should probably avoid vendoring actions which are somewhat simple to script. In this case the `ubuntu-latest` [image ships with the github CLI](](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools), `gh`. We can use that CLI in a script instead and reduce the risk of a supply-chain attack and avoid providing a secret like `GITHUB_TOKEN` to 3rd-party code. The action `peter-evans/create-pull-request` is verified, so it's a little safer than just any action, but for it's generally advisable to BIY instead of vendoring if possible. **If we have to vendor an Action, we should freeze the revision** as is done [here](https://github.com/tektoncd/operator/blob/6cc783d1f87787e39392f954d9359ccced57b8b3/.github/workflows/golangci-lint.yaml#L18) since Github's model defaults to using floating tags (i.e. the Action author can change the action and update the tag, and users automatically get the update without opting in or getting notified)

Here's an example of why it's important to be careful about vendoring Github Actions:
https://lobste.rs/s/4ko499/popular_github_action_tj_actions_changed

hack/openshift/update-tekton-task-versions.sh
```diff
@@ -0,0 +1,58 @@
...
+  echo "Error: tektoncd-catalog repo is missing! Clone it first."
+  exit 1
+fi
+
+# Backup the original fetch script before modifying
+cp "$FETCH_SCRIPT" "$FETCH_SCRIPT.bak"
+
+# Track if any updates were made
+updated=false
+
```
aThorp96, line 18: Optional nitpick: it'd be a bit cleaner to use a return code instead of a global variable. E.g.:
```bash
update_versions() {
  if [[ "some indicator of not updated" ]]; then
    return 1
  fi
}

if update_versions; then
    echo "updated"
else
    echo "not updated"
fi
```

.github/workflows/update-tektoncd-task-versions.yml
infernus01: As we discussed on a call, I have updated the peter-evans/create-pull-request version from v5 to v7's SHA.

.github/workflows/update-tektoncd-task-versions.yml
```diff
@@ -0,0 +1,65 @@
...
+          git add hack/openshift/fetch-tektoncd-catalog-tasks.sh
+          if ! git diff --cached --quiet; then
+            git commit -m "Update Tekton task versions"
+          else
+            echo "No changes to commit. Skipping."
+          fi
+
+      - name: Create Pull Request
+        if: steps.update.outputs.has_changes == 'true'
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e
```
aThorp96, line 51:
```suggestion
        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e  # v7
```

aThorp96 (approved):
/lgtm 

One non-functional suggestion

vdemeester (approved):

PR #2670, opened 2025-04-02
change generateSigningSecret to generate a cosign key pair instead of…

mbpavan:
… ECDSA

# Changes

Operator should create cosing key/pair when generateSigningSecret is enbled
As an admin trying to enhance artifact signing capabilities, I want to automatically generate a cosign key pair when generateSigningSecret is enabled so that I can ensure secure and verifiable artifact signatures.

Actually the key pair signing key was of type ECDSA, and for making the process of signing and verifying more easy,we want to move the generated key to cosign key

Organizations require robust methods for signing and verifying artifacts to ensure their integrity and provenance. Cosign provides a lightweight mechanism for container image signing and verification. Currently, operators  creates an ECDSA key pairs, which introduces operational complexity (revocation, rotation)

# Submitter Checklist

review them:

- [ x] Run `make test lint` before submitting a PR

# Release Notes


mbpavan:
Below are the testing done 
1. In Tektonconfig set generateSigningSecret to true. The secret signing-secrets in tekton-pipelines ns should create cosign.pub and cosign.key.
2. In In Tektonconfig set generateSigningSecret to false. The secret signing-secrets in tekton-pipelines ns should not have cosign.pub and cosign.key.
3. In Tektonconfig set generateSigningSecret to true. The secret signing-secrets in tekton-pipelines ns should create cosign.pub and cosign.key. Upgrade. Once the upgrade is complete, we should still see cosign.pub and cosign.key in the signing-secrets secret. 
4. Create secret signing-secrets in tekton-pipelines ns with x509 pub and private key(without this change). Add these changes and then upgrade the operator. The secret signing-secrets in tekton-pipelines ns  should now have cosign.pub and cosign.key instead of the x509 onces.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -504,9 +501,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonChain
 	if err != nil {
 		secretInstallerSetSigningKey = false
 	}
-	if secretInstallerSetSigningKey != tc.Spec.GenerateSigningSecret {
-		manifest := r.manifest
-		manifest = manifest.Filter(mf.ByKind("Secret"))
+	// Check for x509 keys
```
PuneetPunamiya, line 504: Can you please add more details in the comment like why we are checking for x509 keys

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -680,44 +677,33 @@ func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
...
-		Type:  "PUBLIC KEY",
-		Bytes: publicKeyBytes,
-	})
-
-	return map[string][]byte{
-		"x509.pem":     privateKeyPEM,
-		"x509-pub.pem": publicKeyPEM,
+// isX509SigningSecret checks if any Secret resource in the installer set manifest
+// contains x509 keys
+func isX509SigningSecret(manifests []unstructured.Unstructured) bool {
```
PuneetPunamiya, line 696: I think we can add a TODO here to remove this after like 2/3 releases

PuneetPunamiya (approved):
Thanks 🤙🏻

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -680,44 +677,33 @@ func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
...
-	// Generate ECDSA key pair
-	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
-	if err != nil {
-		logger.Error("Error generating private key:", err)
-		return nil
-	}
 
-	// Convert private key to PKCS8
-	privateKeyBytes, err := x509.MarshalPKCS8PrivateKey(privateKey)
+	keys, err := cosign.GenerateKeyPair(nil)
```
jkhelil, line 681: can you add a comment, why we use nil, (we dont input a password for private key encryption because it will need a user secret)

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -680,44 +677,33 @@ func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
 
 func GenerateSigningSecrets(ctx context.Context) map[string][]byte {
```
jkhelil, line 678: looks like we dont need to export this: s/GenerateSigningSecrets/generateSigningSecrets

PuneetPunamiya: +1

PuneetPunamiya:
@mbpavan please check if upgrade is working fine 🙂

mbpavan:
review comments have been addressed. Please have another look @jkhelil @PuneetPunamiya

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -678,46 +676,37 @@ func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
...
+// contains x509 keys
+// TODO: Remove this function when we no longer check if x509 keys are being used.
+func isX509SigningSecret(manifests []unstructured.Unstructured) bool {
+	for _, u := range manifests {
+		if u.GetKind() == "Secret" {
+			data, found, err := unstructured.NestedStringMap(u.Object, "data")
+			if err != nil || !found {
+				continue
+			}
+			if _, exists := data["x509.pem"]; exists {
```
pratap0007, line 706: nit: We can return exists directly without a condition check.

pratap0007:
LGTM

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
jkhelil: @mbpavan What is the purpose of this ? 'backward copatibility', Please describe the logic for it

mbpavan: Hi @jkhelil the purpose of this is is to support a seamless upgrade path. In older versions of the operator, when generateSigningSecret was enabled, an x509-based ECDSA key pair was created. With the move to cosign keys, we need a way to detect if an installation already has a legacy x509 secret. The function  checks the secret manifest for the presence of a key like "x509.pem". If it finds that key, it indicates that the secret is from the older format. This detection means that that we must regenerate the signing secret using cosign keys rather than retaining the x509 keys. This happens for upgrade cases when we already have x509 keys and when we upgrade need to migrate from x509 to cosign. Hence I think Puneets comment to add a TODO to remove this going forward as we do not need this check once we have migrated to cosign.

PuneetPunamiya: Yeah but if user has x509 keys and during upgrade we are changing to cosign, I'm not sure what would be the impact to the end user. One approach which comes to my mind is for now 
 * If user has x509 we keep and use that only 
 * If it's a fresh install then use cosign
 * And add a deprecation notice that we will remove the x509 thingy and after let's say two successive releases we remove x509

mbpavan: @jkhelil any thoughts on this? Do we need to keep the approach of moving from x509 to cosign upon upgrade or keep x509 if the user already has it?

jkhelil: By default there is an empty signing-secret. We should check that in anycase, the signing-secret has data,(not empty), we shouldnt override it through generateSigningSecret. Please highlight this in the doc. That generateSigningSecret would generate a secret for you only if you dont have one., and that the secret now is a cosign key pair.

docs/TektonChain.md
```diff
@@ -122,13 +122,13 @@ spec:
   transparency.url: #value
 ```
 - `disabled` : if the value set as `true`, chains feature will be disabled (default: `false`)
-- `generateSigningSecret`: If set to true, the operator will generate an ECDSA key pair (x509.pem private key and x509-pub.pem public key) and store it in the "signing-secrets" secret in the "tekton-pipelines" namespace. the secret is used by chains controller to sign tekton artifacts(taskruns, pipelineruns). it is important to mention that: 
- * The user should retrieve and store in a safe place the x509-pub.pem public key to verify later artifact attestations.
+- `generateSigningSecret`: If set to true, the operator will generate a cosign key pair (`cosign.key` private key and `cosign.pub` public key) and store them in the `signing-secrets` secret in the `tekton-pipelines` namespace. This secret is used by the Chains controller to sign Tekton artifacts (taskruns, pipelineruns).
```
jkhelil, line 125: generateSigningSecret: When set to true, the operator will generate a Cosign key pair (cosign.key as the private key and cosign.pub as the public key) and store them in the signing-secrets secret within the tekton-pipelines namespace. This secret is used by the Chains controller to sign Tekton artifacts (taskruns, pipelineruns).

docs/TektonChain.md
```diff
@@ -122,13 +122,13 @@ spec:
...
+   Before enabling generateSigningSecret, the signing-secret is empty and does not contain any data. If the signing-secret does not contain data, enabling generateSigningSecret will create a new Cosign key pair (cosign.key and cosign.pub). However, if the secret already contains data, enabling generateSigningSecret should not overwrite the existing secret. It is important to note that:
```
jkhelil, line 126: If the signing-secret is empty, enabling generateSigningSecret will create a new Cosign key pair. However, if the secret already contains data, enabling generateSigningSecret will not overwrite the existing secret.

docs/TektonChain.md
```diff
@@ -122,13 +122,13 @@ spec:
...
- * The user should retrieve and store in a safe place the x509-pub.pem public key to verify later artifact attestations.
+- `generateSigningSecret`: When set to true, the operator will generate a cosign key pair (`cosign.key` as the private key  and `cosign.pub` as the public key) and store them in the signing-secrets secret within the tekton-pipelines namespace. This secret is used by the Chains controller to sign Tekton artifacts (taskruns, pipelineruns).
+   Before enabling generateSigningSecret, the signing-secret is empty and does not contain any data. If the signing-secret is empty, enabling generateSigningSecret will create a new Cosign key pair. However, if the secret already contains data, enabling generateSigningSecret will not overwrite the existing secret.. However, if the secret already contains data, enabling generateSigningSecret will not overwrite the existing secret. However, if the secret already contains data, enabling generateSigningSecret should not overwrite the existing secret. It is important to note that:
```
jkhelil, line 126: this repeated 
However, if the secret already contains data, enabling generateSigningSecret will not overwrite the existing secret. However, if the secret already contains data, enabling generateSigningSecret should not overwrite the existing secret.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -678,46 +676,19 @@ func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
...
-		Type:  "PUBLIC KEY",
-		Bytes: publicKeyBytes,
-	})
-
+	// Return the keys using cosign naming conventions.
 	return map[string][]byte{
-		"x509.pem":     privateKeyPEM,
-		"x509-pub.pem": publicKeyPEM,
+		"cosign.key": keys.PrivateBytes,
+		"cosign.pub": keys.PublicBytes,
```
Roming22, line 692: We currently use the `cosign generate-key-pair k8s://openshift-pipelines/signing-secrets` command to generate the key.
In addition to the `cosign.key` and `cosign.pub` we have a `cosign.password`, which I don't see in this code change.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
lcarva: Could we generate a random password on the user's behalf?

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
```diff
@@ -678,46 +676,19 @@ func AddControllerEnv(controllerEnvs []corev1.EnvVar) mf.Transformer {
...
-	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
-	if err != nil {
-		logger.Error("Error generating private key:", err)
-		return nil
-	}
 
-	// Convert private key to PKCS8
-	privateKeyBytes, err := x509.MarshalPKCS8PrivateKey(privateKey)
+	// Here, nil indicates that no password is provided for the private key encryption, as it requires a user-provided secret.
+	keys, err := cosign.GenerateKeyPair(nil)
```
lcarva, line 683: Could we generate a random password on the user's behalf?

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
jkhelil: generating a random password make sens, it is better than have nothing :), @mbpavan we need to store in cosign.password as @Roming22 is mentionning

mbpavan: Okay, makes sense. Added.

pkg/reconciler/kubernetes/tektonchain/tektonchain.go
mbpavan: Okay, makes sense. Added.

Roming22:
Looks good to me with the latest changes, looking forward to have access to this change.

jkhelil:
@PuneetPunamiya do you lgtm ?

PR #2680, opened 2025-04-09
Add LoggingPluginMultipartRegex for ResultsAPI

enarha:
Added LOGGING_PLUGIN_MULTIPART_REGEX field for setting configmap of Results API.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonresult_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonresult_types.go
@@ -115,6 +115,7 @@ type ResultsAPIProperties struct {
 	LoggingPluginForwarderDelayDuration *uint  `json:"logging_plugin_forwarder_delay_duration,omitempty"`
 	LoggingPluginQueryLimit             *uint  `json:"logging_plugin_query_limit,omitempty"`
 	LoggingPluginQueryParams            string `json:"logging_plugin_query_params,omitempty"`
+	LoggingPluginMultipartRegex         string `json:"logging_plugin_multipart_regex,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


jkhelil:
@enarha can you run _**make generated**_ please

pkg/apis/operator/v1alpha1/tektonresult_types.go
```diff
@@ -115,6 +115,7 @@ type ResultsAPIProperties struct {
 	LoggingPluginForwarderDelayDuration *uint  `json:"logging_plugin_forwarder_delay_duration,omitempty"`
 	LoggingPluginQueryLimit             *uint  `json:"logging_plugin_query_limit,omitempty"`
 	LoggingPluginQueryParams            string `json:"logging_plugin_query_params,omitempty"`
+	LoggingPluginMultipartRegex         string `json:"logging_plugin_multipart_regex,omitempty"`
```
PuneetPunamiya, line 118: Do we need to add any validation test for this ?

PuneetPunamiya: I mean do we need to add any validation for this field, if so we can add it here https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonresult_validation.go

enarha: IMO no. This is not something which applies to the TektonResult CRD, but rather configuration option stored in a ConfigMap. It matches against 3rd party service (e.g Vector) which is not managed by the operator, so it's not something we can match end to end.

PR #2707, opened 2025-04-30
ADD: OPC task as a part of openshift-pipelines namespace

abhishek-ghosh-codes:
# Changes

-     Update config.yaml of Operator to add key and values for opc
-     Add opc task to hack folder of Operator
-     Making sure that image is built by Konflux

### DIFF
--- a/config/openshift/base/operator.yaml
+++ b/config/openshift/base/operator.yaml
@@ -105,6 +105,8 @@ spec:
           value: registry.redhat.io/ubi9/openjdk-17@sha256:cdbd64ba1b7826d09bc180fa19aa6fbc09ddb5bff5e02b7e1351353bbabebb78
         - name: IMAGE_ADDONS_OC
           value: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+        - name: IMAGE_ADDONS_OPC
+          value: registry.redhat.io/openshift-pipelines/pipelines-cli-opc-rhel8@sha256:ae968e107c963f48910f7991419c29d7253d237c0752bbaaf9eca92811007a7a
       - name: openshift-pipelines-operator-cluster-operations  # tektoninstallerset reconciler
         image: ko://github.com/tektoncd/operator/cmd/openshift/operator
         args:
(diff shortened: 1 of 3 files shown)


operatorhub/openshift/config.yaml
```diff
@@ -138,14 +138,21 @@ image-substitutions:
       envKeys:
       - IMAGE_ADDONS_PARAM_KN_IMAGE
       - IMAGE_ADDONS_KN
-- image: registry.redhat.io/rhel9/skopeo@sha256:993f799d58c48151ecae7a84e63860f6c0039e8e18efed5ad67e7752f4b7e408
+- image: registry.redhat.io/openshift-pipelines/pipelines-cli-opc-rhel8
```
savitaashture, line 141:
```suggestion
- image: registry.redhat.io/openshift-pipelines/pipelines-cli-opc-rhel9
```

config/openshift/base/operator.yaml
```diff
@@ -105,6 +105,8 @@ spec:
           value: registry.redhat.io/ubi9/openjdk-17@sha256:cdbd64ba1b7826d09bc180fa19aa6fbc09ddb5bff5e02b7e1351353bbabebb78
         - name: IMAGE_ADDONS_OC
           value: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
+        - name: IMAGE_ADDONS_OPC
+          value: registry.redhat.io/openshift-pipelines/pipelines-cli-opc-rhel8@sha256:ae968e107c963f48910f7991419c29d7253d237c0752bbaaf9eca92811007a7a
```
savitaashture, line 109: Can you remove this 
as this is handled by config.yaml

abhishek-ghosh-codes: Sure! This will be handled by the config.yaml, so we can remove it from here. Thanks for pointing that out.

operatorhub/openshift/config.yaml
abhishek-ghosh-codes: Updated rhel8 to rhel9

savitaashture (approved):

PR #2281, opened 2024-08-08
Operator now supports StepAction alongside Tasks from ecosystem

savitaashture:
PR includes below changes

* Introduction of a separate tektoninstallerset to manage stepaction.
* Support for both versioned and non-versioned stepactions.
* Implementation of image replacement for stepaction.
* Default enabling of enable-step-actions set to true.
* Bump released version for all ecosystem tasks
* Additionally, changes have been made to create StepAction within the openshift-pipelines namespace alongside Tasks. This prepares for future use when Cluster resolver support for stepaction resolution is available. Disclaimer: Currently, the cluster resolver does not support stepaction resolution.

Disclaimer: Currently, the cluster resolver does not support stepaction resolution.

Signed-off-by: Savita Ashture <sashture@redhat.com>

**Verification:**

```
$ kubectl get tektoninstallerset
NAME                                              READY   REASON
addon-custom-resolverstepaction-8nvw8             True    
addon-versioned-resolverstepactions-devel-t2qss   True   
```

```
$ kubectl get stepaction -n openshift-pipelines
NAME                AGE
git-clone           25m
git-clone-devel-0   25m
```

# Changes

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Run `make test lint` before submitting a PR
- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/operator/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

PR #2732, opened 2025-05-21
Fix the rbac for ecosystem tasks and stepactions

khrm:
RBAC wasn't taking effect because we only generate manifest on the basis of stepactions and tasks sub folders. This fixes it.

# Changes

# Submitter Checklist

review them:

# Release Notes


khrm (commented):
/kind bug

savitaashture:
@khrm I have hard time to understand the PR changes
* why are we creating separate role for stepactions
* The current role also gives permission to stepactions why its not taking the effect :thinking:

savitaashture (changes requested):
@khrm I have hard time to understand the PR changes

* why are we creating separate role for stepactions
* The current role also gives permission to stepactions why its not taking the effect 🤔

savitaashture:
> 1. If you look into code, we keep step actions and tasks in separate directory and have a separate manifest. So we should also segregate the role and rolebindings also.
> 
> Before this PR https://github.com/tektoncd/operator/pull/2281/files, roles along with tasks were in the parent directory, So there was no issue then.
> 
> 2. Currently, this role isn't created at all because we don't create any manifests for this role at parent location. So it won't take effect.

I see make sense

Just one point 
shall we keep role.yaml and rolebinding.yaml here https://github.com/khrm/operator/tree/ecosystem-rbac/cmd/openshift/operator/kodata/tekton-addon/addons/06-ecosystem at the root location to avoid duplication

khrm:
@savitaashture We can but nothing wrong in having finer grain RVAC. Both RBAC are different. One for stepaction and another for tasks.

PR #2788, opened 2025-07-04
Add PAC support for Kubernetes/Kind platform via Operator

mbpavan:
# Changes

Prior to this change, the Operator shipped an OpenShiftPipelinesAsCode custom resource and controller that assumed an OpenShift environment. Teams running Tekton on vanilla Kubernetes lacked an integrated “Pipelines as Code” experience in the Operator. Adding this support now. 
# Submitter Checklist

review them:

- [x ] Run `make test lint` before submitting a PR
- [x ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)

# Testing
1. Install tests - Verified fresh installation of the Operator with Pipelines as Code enabled on both OpenShift and Kubernetes clusters. For Kind cluster can see crd, cr, installerset and controllers running successfully with this change.

2. Upgrade tests - Tested upgrading from an existing Kind installation without Pipelines as Code to the new version with Pipelines as Code enabled. Also tested upgrade from openshift with additional controllers to make sure pac and additional controllers exist after the upgrade.

3. Added E2e and unit tests.

# Release Notes

```release-note
PipelinesAsCode can now be enabled on vanilla Kubernetes clusters via spec.platforms.kubernetes.pipelinesAsCode.enable: true. Existing OpenShift pac using spec.platforms.openshift.pipelinesAsCode.enable: true continue without modification.
```
-->


pkg/apis/operator/v1alpha1/tektonconfig_types.go
```diff
@@ -186,4 +186,6 @@ type Platforms struct {
 	// OpenShift allows configuring openshift specific components and configurations
 	// +optional
 	OpenShift OpenShift `json:"openshift,omitempty"`
+	// +optional
+	Kubernetes Kubernetes `json:"kubernetes,omitempty"`
```
pratap0007, line 190: Can't we manage PAC outside of the platform, as it will be supported on both OpenShift and Kubernetes, just like other components such as Results and Chains?

mbpavan: We could, but the original design specifically introduced spec.platforms.openshift.pipelinesAsCode for PAC and even named the CRD as openshiftpipelinesascodes.operator.tekton.dev. I agree with your point, but since this structure is already in place, we’ll unfortunately need to stick with this platform setup to maintain backward compatibility (and upgrade path) for PAC on OpenShift clusters imho. Unforunately this also means that the crd and the code references will also have openshiftpiplelinesascode instead of pipelinesascode. We could move it to just PAC at a later stage I guess.

pkg/apis/operator/v1alpha1/pipelinesascode.go
```diff
@@ -0,0 +1,28 @@
...
+// PipelinesAsCode holds common settings for both Kubernetes and OpenShift.
+// +k8s:openapi-gen=true
+type PipelinesAsCode struct {
+	// Enable or disable pipelines as code by changing this bool
+	// +optional
+	Enable *bool `json:"enable,omitempty"`
+	// PACSettings allows user to configure PAC configurations
+	// +optional
+	PACSettings `json:",inline"`
+}
```
jkhelil, line 28: Any reason, this is not in openshiftpipelineascode_types.go
if it is about naming, we can handle this in one pr to rename all

anithapriyanatarajan:
@mbpavan - Please update https://github.com/tektoncd/operator/blob/main/charts/tekton-operator/templates/kubernetes-crds.yaml so that helm charts for k8s cluster are up to date

pkg/apis/operator/v1alpha1/tektonconfig_types.go
anithapriyanatarajan: @mbpavan is it possible to revisit the cr name. The naming could result unnecessary issues. What would be the impact of accomodating this change. @vdemeester if we plan to adopt PAC upstream, I assume the PAC CRD might require a renaming from `openshiftpipelinesascodes.operator.tekton.dev` to `pipelinesascodes.operator.tekton.dev`

Attached herewith the screenshot of CRDs in a k8s cluster. 

Screenshot From 2025-07-17 11-44-09

anithapriyanatarajan:
doc https://github.com/tektoncd/operator/blob/main/docs/OpenShiftPipelinesAsCode.md needs an update. `custom-console-url ` is applicable for openshift while for tekton dashboard it would be `tekton-dashboard-url`

pkg/apis/operator/v1alpha1/tektonconfig_types.go
mbpavan: I would defer to @vdemeester and @jkhelil on that. I core issue here is the the crd and code was defined to be specific to ocp. If we were to switch to pitpelinesascodes.operator.tekton.dev obviosuly upgrade breaks and backwards compatibility is lost. I think we have have a follow up PR which has pipelinesascodes.operator.tekton.dev cr along with openshiftpipelinesascodes.operator.tekton.dev and after a depreatiation period remove openshiftpipelinesascodes.operator.tekton.dev completely. How does that sound?

jkhelil: My suggestion is to have both CRDS from now, pitpelinesascodes.operator.tekton.dev  the target one, exposes it from tektonconfig as other component using same api
openshiftpipelinesascodes.operator.tekton.dev should still work as it is for some time, we would advise customers to migrate through a migration plan in the future
but definitely i am for having the target crd, target api, from now, in parallell to the existant

mbpavan:
Following an offline discussion, it was agreed that several additional changes are required to support PAC installation via the Operator on Kubernetes:

- Rename the CRD and related code from openshiftpipelinesascodes.operator.tekton.dev to pipelinesascodes.operator.tekton.dev.

- Refactor the codebase to isolate Kubernetes-specific logic (e.g., transformers) into a dedicated kubernetes directory, and introduce an OpenShift extension following the existing component structure.

- Ensure a smooth upgrade path by supporting both CRDs (openshiftpipelinesascodes.operator.tekton.dev and pipelinesascodes.operator.tekton.dev) during the transition phase, with plans to deprecate the former in a future release.

- Modify TektonConfig to eliminate the platform-specific configuration for OpenShift, aligning PAC with the approach used for other platform-agnostic components like Pipelines.

- Perform thorough regression testing on both Kubernetes and OpenShift platforms, given that this involves refactoring existing code and not merely adding Kubernetes support.

Given the above, I will close this PR for now. A future PR will incorporate these broader changes comprehensively.

PR #2809, opened 2025-07-25
add init container support to tekton-operator deployment

cmorinupgrade:
# Changes

  Allow configuring init containers in the Helm chart deployment template when `operator.initContainers.enabled` is set to true.

  Usage example:

  We use this to inject a `initContainer` before Tekton Operator bootup, like so:

```yaml
operator:
  initContainers:
    enabled: true
    containers:
      - name: init-setup
        image: "<image>"
        command: ['sh', '-c', '<custom operation>']
```

  # Submitter Checklist

  review them:

  # Release Notes

  ```release-note
  Add support for configuring init containers in Tekton Operator Helm chart deployment template

### DIFF
--- a/charts/tekton-operator/templates/deployment.yaml
+++ b/charts/tekton-operator/templates/deployment.yaml
@@ -24,6 +24,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.operator.deployment.initContainers.enabled }}
+      initContainers:
+        {{- toYaml .Values.operator.deployment.initContainers.containers | nindent 8 }}
+      {{- end }}
       containers:
         - env:
             - name: SYSTEM_NAMESPACE
--- a/charts/tekton-operator/values.yaml
+++ b/charts/tekton-operator/values.yaml
@@ -48,6 +48,12 @@ operator:
   deployment:
     # Custom labels for the Deployment resource.
     customLabels: ""
+    # Custom init containers for the Deployment resource.
+    # Can be used for pre-deployment setup and validation (e.g. waiting for specific resources to be available prior to starting the operator)
+    # see https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
+    initContainers:
+      enabled: false
+      containers: []
     # Custom labels for the Deployment Pod Template.
     podTemplateCustomLabels: ""
 


charts/tekton-operator/templates/deployment.yaml
```diff
@@ -24,6 +24,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.operator.initContainers.enabled }}
```
anithapriyanatarajan: 1. There is no corresponding entry in values.yaml to enable or disable operator.initContainers.enabled.

2. Init containers are not part of Tekton’s core operator or webhook functionality. Unless this addition supports a clear and justified Tekton use case, it’s best not to include it in the Operator Helm chart. Please clarify the intent and how we can verify the integrity and necessity of the activities performed by these init containers.

cmorinupgrade: Hi @anithapriyanatarajan 👋 

Thanks for the feedback. Last commit addresses both comments.

To expand on the second question, we are running the operator in a ipv6 cluster in EKS, and we are currently iterating with AWS support in order to mitigate an issue we're getting right now at operator bootup:

`2025/07/15 11:48:42 Error reading/parsing logging configuration: timed out waiting for the condition: Get "https://[fd1d:f086:2d51::1]:443/api/v1/namespaces/tekton-pipelines/configmaps/config-logging-triggers": context deadline exceeded`  

The rundown is:
- `config-logging` configmap is failing to fetch, which happens at the very start of Kubernetes workloads that use the knative logging library (which all the Tekton components do use as well).
- this always happens when pods are new upon restart & stabilizes after. ~2 restarts per pod (when the pod network is ready).
- all the pods that restart are all on the same node, which is also a fresh node that karpenter put up as its consolidating/bin-packing the cluster.
- this is due to some of pods starting faster then the networking setup on the node, which have the pods fail to connect to kube api server for few seconds, resulting in restarts.

Although the issue we're currently having is not directly related to the Tekton Operator, I think init containers is a common pattern used to enable flexibility in general.

Hope this helps! Thank you.

anithapriyanatarajan:
@jkhelil - Before merging this PR, could we please evaluate the security implications of allowing arbitrary init containers in the Tekton Operator deployment? While I agree that supporting init containers can be valuable in certain custom scenarios, this requirement appears to be specific to a particular use case and is not directly related to Tekton’s default installables.

Given that, it may be more appropriate to either:

1. Whitelist specific init container images, or
2. Recommend that the requestor maintain a custom version of the Helm chart tailored to their needs, rather than introducing this flexibility into the official Tekton Operator chart.

khrm:
I think this PR is fine. Helm is used by admin. They are well aware of security issues. If some admins want to inject malicious code in his  own cluster, then there are easier ways than using helm.

anithapriyanatarajan:
/release-note-edit ```release-note Support initContainers through Helm chart```

PR #2810, opened 2025-07-28
fix: prevent panic when pruner field is nil during upgrade

l-qing:
Add nil pointer checks to `TektonPruner.IsDisabled()` method to safely handle cases where Pruner or Disabled pointer is nil during upgrade scenarios.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix panic in `TektonPruner` when `Disabled` field is nil during upgrade
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonpruner_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonpruner_types.go
@@ -88,5 +88,9 @@ func (tp *TektonPruner) GetStatus() TektonComponentStatus {
 
 // IsDisabled returns true if the TektonPruner is disabled
 func (p *Pruner) IsDisabled() bool {
+	if p == nil || p.Disabled == nil {
+		// When the Pruner is nil or Disabled is nil, we assume it is the default state.
+		return DefaultPrunerDisabled
+	}
 	return *p.Disabled
 }


pkg/apis/operator/v1alpha1/tektonpruner_types.go
```diff
@@ -88,5 +88,5 @@ func (tp *TektonPruner) GetStatus() TektonComponentStatus {
 
 // IsDisabled returns true if the TektonPruner is disabled
 func (p *Pruner) IsDisabled() bool {
-	return *p.Disabled
+	return p != nil && p.Disabled != nil && *p.Disabled
```
l-qing: In upgrade scenarios, the value of this `Disabled` field may be null.

l-qing:
```sh
panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1f6e8cb]
goroutine 476 [running]:

github.com/tektoncd/operator/pkg/apis/operator/vlalpha1. (*Pruner).IsDisabled (...)
    github.com/tektoncd/operator/pkg/apis/operator/v1alpha1/tektonpruner_types.go:91
github.com/tektoncd/operator/pkg/reconciler/shared/tektonconfig.(*Reconciler).ReconcileKind(0xc000631b90, {0x2b52d70, 0xC001d4e8d0}, 0XC000529808)
    github.com/tektoncd/operator/pkg/reconciler/shared/tektonconfig/tektonconfig.go:185 +0x114b
github.com/tektoncd/operator/pkg/client/injection/reconciler/operator/v1alphal/tektonconfig.(*reconcilerImpl).Reconcile(0xc0000eb5e0, {0x2b52d70, 0 xC001d4e8a0}, {0xc00144f700, 0x6})
    github.com/tektoncd/operator/pkg/client/injection/reconciler/operator/v1alphal/tektonconfig/reconciler.go:236 +0x4d2
knative.dev/pkg/controller.(*Impl).processNextWorkItem(0xc0004f1e60)
    knative.dev/pkg@v0.0.0-20250424013628-d5e74d29daa3/controller/controller.go:540 +0x508
knative.dev/pkg/controller. (*Impl).RunContext.func3()
    knative.dev/pkg@v0.0.0-20250424013628-d5e74d29daa3/controller/controller.go:489+0x53
created by knative.dev/pkg/controller.(*Impl).RunContext in goroutine 457
    knative.dev/pkg@v0.0.0-20250424013628-d5e74d29daa3/controller/controller.qo:487
```

l-qing: Although this sets default values, in upgrade scenarios, the old webhook might respond to the default event first.https://github.com/tektoncd/operator/blob/b8826d6e8a03cacb6b4de87a829042173895bfa0/pkg/apis/operator/v1alpha1/tektonpruner_defaults.go#L27-L35

pratap0007: [this](https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonconfig_defaults.go#L35) take care of setting default values.

anithapriyanatarajan:
@pramodbindal - this validation is needed since we have set the type for `Disabled` as `pointer to bool` rather than `bool`. Ref: https://github.com/tektoncd/operator/blob/main/pkg/apis/operator/v1alpha1/tektonpruner_types.go#L45. having it as a pointer allows the case where the user may leave the field 'nil'. Do we need this behavior?

l-qing:
> having it as a pointer allows the case where the user may leave the field 'nil'. Do we need this behavior?

It seems we don't need a pointer type here, we can just use a bool type directly.

The user cannot forcibly set this field to null because we have a default webhook that sets default values.
The situation where we encounter `null` values occurs during the upgrade process, when the old webhook might have handled the default case. When it gets stored in etcd, that field becomes null (because it simply doesn't know about the newly added field).
This only exists at that specific moment during the upgrade - when the old `tekton-operator-webhook` is still working, and the switch was handled by the old Pod that processed the default webhook.

pkg/apis/operator/v1alpha1/tektonpruner_types.go
l-qing: This is the ideal scenario, where the old `tekton-operator-webhook` has already stopped working, and now all webhook requests are forwarded to the new Pod, which gives you the expected behavior.

The actual situation might be more complex, meaning the old `tekton-operator-webhook` hasn't completely stopped working yet.

pramodbindal: Can you change the condition to something like this.

```
// IsDisabled returns true if the TektonPruner is disabled
func (p *Pruner) IsDisabled() bool {
	if p == nil || p.Disabled == nil {
		return DefaultPrunerDisabled
	}
	return *p.Disabled
}
```

Reason: When this value is not set the we don't want to  return `false` as caller will assume pruner is enabled. Instead we should return default value and let pruner decide what should be the default value.

l-qing: I have made adjustments according to your suggestions and added a line of comments. 😆

PR #2814, opened 2025-07-29
fix(openshift git resolver): mount the trusted-ca-configmap into component's system ca store in Openshift

aThorp96:
# Changes
Mount the trusted-ca-configmap into component's system ca store, allowing the git resolver to use the custom CA bundle.

This is necessary for "legacy applications" (e.g. cURL or git) to use the CA bundles as they do not respect `SSL_CERT_DIR`, when components such as the git-resolver utilize legacy applications. In the Openshift environment the TrustedCAConfigMap has both the default and custom certificates combined. Note that the TrustedCAConfigMap does not contain the Service CA bundle, however that bundle is utilized for the internal image registry and its tooling respects `SSL_CERT_DIR`. Also note that this transformer should not be applied to containers with user-defined images since it both assumes the image is a RHEL (or similar) environment and because it may override a user's image's custom certificate bundle (see #549 and [SRVKP-1497](https://issues.redhat.com/browse/SRVKP-1497))

Resolves [SRVKP-8204](https://issues.redhat.com/browse/SRVKP-8204)

See also:
- Openshift documentation on [certificate injection](https://github.com/openshift/openshift-docs/blob/a8269cf65696fbd08647c8f3b5d065d53a8a1f52/modules/certificate-injection-using-operators.adoc)
- System CA management - [man (8) update-ca-trust](https://man.archlinux.org/man/update-ca-trust.8)
- Openshift Pipelines bug - git resolver not respecting custom certificates - [SRVKP-8204](https://issues.redhat.com/browse/SRVKP-8204)

# Submitter Checklist

review them:

# Release Notes

```release-note
Bug fix: Before this change, in Openshift Pipelines all components' deployents and statefulsets had the Openshift trusted CA bundle mounted at `/tekton-custom-certs/ca-bundle.crt` which was used by programs via the environment variable `SSL_CERT_DIR`. After this change, those deployments and statefulsets now also have the CA bundle mounted at `/etc/pki/ca-trust/extracted/tls-ca-bundle.pem` as if the certificates were installed on the system with `update-ca-trust extract`. This allows binaries such as `git` which do not recognize `SSL_CERT_DIR` to use the custom certificates without any additional configuration. Fixes SRVKP-8204
```

### DIFF
--- a/docs/TektonOperator.md
+++ b/docs/TektonOperator.md
@@ -92,3 +92,18 @@ After installing the resources, `TektonInstallerSet` waits for deployment pods t
 
   As we have extension mechanism where we handle platform specific resources, in case of OpenShift we create additional resources in Pre and Post Reconciler in TektonPipeline. In both the cases we have an `TektonInstallerSet` created, on upgrade or target namespace change we delete the old and create a new `TektonInstallerSet`. 
 
+## Tekton Operator on Openshift
+When the Tekton Operator is [installed](./install.md) for Openshift, the
+Operator configure Tekton in order to cater Tekton the deployment for an
+Openshift cluster.
+
+### Custom CA Certificates
+When installed for Openshift, the Operator will inject the Openshift
+Proxy's custom certificates into the Tekton components' pods system trust
+store. This allows components such as the remote resolvers to connect
+to domains which use self-signed or private certificates for TLS.
+
+Documentation on configuring custom certificates for Openshift Proxy can
+be found in the [Openshift Network Configuration documentation][openshift-proxy-configuration].
+
+[openshift-proxy-configuration]: https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/configuring_network_settings/configuring-a-custom-pki
--- a/pkg/reconciler/common/certificates.go
+++ b/pkg/reconciler/common/certificates.go
@@ -35,8 +35,8 @@ const (
 	ServiceCAKey             = "service-ca.crt"
 )
 
-// newVolumeWithConfigMap creates a new volume with the given ConfigMap
-func newVolumeWithConfigMap(volumeName, configMapName, configMapKey, configMapPath string) corev1.Volume {
+// NewVolumeWithConfigMap creates a new volume with the given ConfigMap
+func NewVolumeWithConfigMap(volumeName, configMapName, configMapKey, configMapPath string) corev1.Volume {
 	return corev1.Volume{
 		Name: volumeName,
 		VolumeSource: corev1.VolumeSource{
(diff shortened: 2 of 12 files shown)


jkhelil:
@aThorp96 
I dont understand the purpose of the PR
```
This is necessary for "legacy applications" (e.g. cURL or git) to use the CA bundles as they do not respect SSL_CERT_DIR. In the Openshift environment the TrustedCAConfigMap has both the default and custom certificates combined. Note that the TrustedCAConfigMap does not contain the Service CA bundle, however that bundle is utilized for the internal image registry and its tooling respects SSL_CERT_DIR
```
the operator doesnt handle the user workload, the transformation you propose, will only act on pipelines controllers (controller, watcher, chain, result ....)
there is no legacy applications here
The operator itself creates the ca bundle configmap, 
If your intent is to mount cabundleconfigmap into user workload, you should do the PR in pipelines repo

aThorp96:
@jkhelil 
> the operator doesnt handle the user workload, the transformation you propose, will only act on pipelines controllers (controller, watcher, chain, result ....)

Based on https://github.com/tektoncd/operator/pull/549 and [it's ticket SRVKP-1497](https://issues.redhat.com/browse/SRVKP-1497), the mutating admission controller (proxy-webhook) installed by the operator _can_ modify user workloads when [applying defaults](https://github.com/tektoncd/operator/blob/include-openshift-certs-in-system-certs/pkg/reconciler/proxy/proxy.go#L323) to TaskRuns or PipelineRuns, though possibly this happens indirectly somehow, I'm not 100% sure. That comment is to clarify why that logic is not included in `pkg.reconciler.common.AddCAConfigMapsToVolumes`. 

>  there is no legacy applications here

I will clarify in the PR description and commit message. Components may utilize "legacy applications" (e.g. the git-resolver uses `git`) and require this change

aThorp96:
Not sure what's going on with the e2e tests, the failure seems to be pulling the UI yaml due to a 403 from Github's CDN....

savitaashture:
@aThorp96  Thanks for the PR :pray: 
Few points

1. Can we add a doc for this
2. Is it possible to add an e2e for this change to validate the changes
3. Can you also verify it works on Both Openshift + Kubernetes based env
4. Please test upgrade scenario if not done yet
5. Based on the code changes i assume there is no changes required by end use but still can you confirm once thank you

aThorp96:
@savitaashture I will need some guidance on end to end tests and upgrade tests

aThorp96:
> Can you also verify it works on Both Openshift + Kubernetes based env
 
@savitaashture I've verified it works in openshift, though this does not apply to kubernetes based environments

aThorp96:
@savitaashture 

> Can we add a doc for this
Done 👍🏼 

> Is it possible to add an e2e for this change to validate the changes
Done 👍🏼 

> Can you also verify it works on Both Openshift + Kubernetes based env
Verified. This has no effect on a Kubernetes environment

> Please test upgrade scenario if not done yet
I believe I've tested this, but I'm new to contributing to the Tekton Operator so please check my process:
1. Checked out `main` and installed on an openshift cluster with `make apply TARGET=openshift` 
2. Validated the TektonInstallerSets were successfully reconciled and the pods in the `openshift-pipelines` namespace are running normally
3. Checked out this branch
4. Deleted the TektonInstallerSets from the testing cluster
5. Installed this branch to the testing cluster with `make apply TARGET=openshift`
6. Validated the TektonInstallerSets were successfully reconciled and the pods in the `openshift-pipelines` namespace are running normally
7. Validated the new pods had the correct volume per this change

> Based on the code changes i assume there is no changes required by end use but still can you confirm once thank you
No changes are required from end users 👍🏼

jkhelil:
@aThorp96 We aslo have ApplyCABundlesForStatefulSet
We should amend this too, with new system dir injection 
Please reword the description to hignlight that we do that injection for git resolver fix, even it is genralised here

aThorp96:
@jkhelil I've updated the description 👍🏼 

This change already applies to both `ApplyCABundlesToDeployment` and `ApplyCABundlesForStatefulSet` 👍🏼

jkhelil:
@savitaashture Please lgtm

PR #2831, opened 2025-08-12
Optimize proxy webhook performance by eliminating API calls during pod admission

jkhelil:
# Changes

During load tests, the proxy webhook was experiencing timeout issues that caused TaskRun failures with the error:
```
failed to create task run pod \"jhue05939bba6498044b9e123b65b3893c5-coverity-availability-check\": Internal error occurred: failed calling webhook \"proxy.operator.tekton.dev\": failed to call webhook: Post \"https://tekton-operator-proxy-webhook.openshift-pipelines.svc:443/defaulting?timeout=10s\": context deadline exceeded. Maybe missing or invalid Task jhutar-1-tenant/
```
The proxy webhook performed 2 synchronous API calls for every pod creation to check if CA bundle ConfigMaps exist:
checkConfigMapExist(client, ctx, namespace, "config-trusted-cabundle")
checkConfigMapExist(client, ctx, namespace, "config-service-cabundle")
Under high concurrency (load tests), these API calls became a bottleneck that exceeded the 10-second webhook timeout.
Solution
Replace API calls with optional ConfigMap volumes that gracefully handle missing ConfigMaps without requiring existence checks.

# Submitter Checklist

review them:

# Release Notes

```release-note
Eliminated proxy webhook timeout issues during high-concurrency load tests by removing synchronous API calls that checked ConfigMap existence during pod admission. The webhook now uses optional ConfigMap volumes that gracefully handle missing CA bundles without blocking pod creation, improving scalability while maintaining all existing SSL certificate functionality.
```

### DIFF
--- a/pkg/reconciler/common/certificates.go
+++ b/pkg/reconciler/common/certificates.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/ptr"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/reconciler/common/certificates.go
```diff
@@ -53,6 +54,26 @@ func NewVolumeWithConfigMap(volumeName, configMapName, configMapKey, configMapPa
 	}
 }
 
+// NewVolumeWithConfigMapOptional creates a new volume with the given ConfigMap marked as optional
+// This allows the pod to start even if the ConfigMap doesn't exist
```
savitaashture, line 58: One Q: 
Does that mean proxy pod don't require these configmap right 
and also which pod has dependency on these cm

jkhelil: the operator proxy, is an admission webhook that injects that mounts these configmaps to taskrun pods
we mark the volume as optional, so they got mounted even if the configmaps doesnt exist whihc avoids api call to check if they exist and then create a standard volume.
these configmaps are created at namespace reconciliation before taskruns pod etc, if the user opt to have these configmaps, otherwise they are not created

jkhelil:
@vdemeester can you review/approve please

savitaashture (approved):

PR #2983, opened 2025-11-18
Auto-upgrade Results PostgreSQL 13 ->15 

enarha:
Added a wrapper script for the Postgres image which checks the version of the running DB directory structure and trigger a one time upgrade if required levereging the tools provided with the image. Easy to remove after two or three versions assuming all existing deployments have switched to version 15.

Assisted-by: Cursor

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Update Tekton Results PostgreSQL dependency to version 15. The update requires transformation of the directory structure used by PostgreSQL. This only applies in cases where Tekton Results is using the default deployment of PostgreSQL image. Such deployment is only recommended for preview and testing purposes and there is no support for such deployments or guarantees for the directory structure migration. For production environments it is recommended that Tekton Results is used external DB allowing proper scaling and management.
```

### DIFF
--- a/cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
+++ b/cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
@@ -10,6 +10,56 @@ metadata:
   namespace: tekton-pipelines
 ---
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/name: tekton-results-postgres
+    app.kubernetes.io/part-of: tekton-results
+  name: postgres-upgrade-scripts
+  namespace: tekton-pipelines
+data:
+  postgres-wrapper.sh: |
+    #!/bin/bash
+    set -eu
+
+    PGDATA="${PGDATA:-/var/lib/pgsql/data}"
+    CURRENT_VERSION="15"
+    UPGRADE_FROM_VERSION="13"
+    PG_VERSION_FILE="$PGDATA/userdata/PG_VERSION"
+
+    echo "=== PostgreSQL Container Starting ==="
+
+    # Check if data directory exists with previous version
+    if [ -f "$PG_VERSION_FILE" ]; then
+        DATA_VERSION=$(cat "$PG_VERSION_FILE")
+        echo "Data directory PostgreSQL version: $DATA_VERSION"
+
+        # Check if upgrade is needed from version 13 to 15
+        if [ "$DATA_VERSION" = "$UPGRADE_FROM_VERSION" ]; then
+            echo "========================================="
+            echo "PostgreSQL Upgrade Mode Activated"
+            echo "Upgrading from PostgreSQL $UPGRADE_FROM_VERSION to PostgreSQL $CURRENT_VERSION"
+            echo "========================================="
+
+            # Set the POSTGRESQL_UPGRADE environment variable
+            # The sclorg PostgreSQL container will detect this and run pg_upgrade
+            export POSTGRESQL_UPGRADE="copy"
+
+            echo "Starting PostgreSQL with POSTGRESQL_UPGRADE=copy"
+        elif [ "$DATA_VERSION" = "$CURRENT_VERSION" ]; then
+            echo "PostgreSQL data is already version $CURRENT_VERSION - starting normally"
+        else
+            echo "WARNING: Unexpected PostgreSQL version $DATA_VERSION"
+        fi
+    else
+        echo "No existing data found - PostgreSQL will initialize new database"
+    fi
+
+    # Execute the standard PostgreSQL container entrypoint
+    exec run-postgresql
+immutable: true
+---
+apiVersion: v1
 kind: Service
 metadata:
   labels:
@@ -63,7 +113,7 @@ spec:
                 secretKeyRef:
                   key: POSTGRES_USER
                   name: tekton-results-postgres
-          image: registry.redhat.io/rhel9/postgresql-16@sha256:7e6de7395c848a5b22ec8256ecc6a9e422995caed528cf67773453a7287d5cbb
+          image: registry.redhat.io/rhel9/postgresql-15@sha256:90ec347a35ab8a5d530c8d09f5347b13cc71df04f3b994bfa8b1a409b1171d59
           name: postgres
           ports:
             - containerPort: 5432
--- a/pkg/reconciler/kubernetes/tektonresult/transform.go
+++ b/pkg/reconciler/kubernetes/tektonresult/transform.go
@@ -116,6 +116,7 @@ func (r *Reconciler) transform(ctx context.Context, manifest *mf.Manifest, comp 
 		common.StatefulSetImages(resultImgs),
 		common.AddConfigMapValues(tektonResultleaderElectionConfig, instance.Spec.Performance.PerformanceLeaderElectionConfig),
 		common.UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(&instance.Spec.Performance, tektonResultleaderElectionConfig, resultWatcherDeployment, resultWatcherContainer),
+		// Note: PostgreSQL upgrade transformer is NOT needed for Kubernetes
 	}
 
 	if instance.Spec.Performance.StatefulsetOrdinals != nil && *instance.Spec.Performance.StatefulsetOrdinals {
--- a/pkg/reconciler/openshift/tektonresult/extension.go
+++ b/pkg/reconciler/openshift/tektonresult/extension.go
@@ -37,17 +37,20 @@ import (
 
 const (
 	// manifests console plugin yaml directory location
-	routeRBACYamlDirectory  = "static/tekton-results/route-rbac"
-	logsRBACYamlDirectory   = "static/tekton-results/logs-rbac"
-	deploymentAPI           = "tekton-results-api"
-	serviceAPI              = "tekton-results-api-service"
-	routeAPI                = "tekton-results-api"
-	secretAPITLS            = "tekton-results-tls"
-	apiContainerName        = "api"
-	boundSAVolume           = "bound-sa-token"
-	boundSAPath             = "/var/run/secrets/openshift/serviceaccount"
-	lokiStackTLSCAEnvVar    = "LOGGING_PLUGIN_CA_CERT"
-	tektonResultWatcherName = "tekton-results-watcher"
+	routeRBACYamlDirectory     = "static/tekton-results/route-rbac"
+	logsRBACYamlDirectory      = "static/tekton-results/logs-rbac"
+	deploymentAPI              = "tekton-results-api"
+	serviceAPI                 = "tekton-results-api-service"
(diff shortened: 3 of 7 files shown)


cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
```diff
@@ -63,7 +112,7 @@ spec:
                 secretKeyRef:
                   key: POSTGRES_USER
                   name: tekton-results-postgres
-          image: registry.redhat.io/rhel9/postgresql-16@sha256:7e6de7395c848a5b22ec8256ecc6a9e422995caed528cf67773453a7287d5cbb
+          image: registry.redhat.io/rhel9/postgresql-15@sha256:90ec347a35ab8a5d530c8d09f5347b13cc71df04f3b994bfa8b1a409b1171d59
```
savitaashture, line 116: @enarha any reason to use postgresql-15 ??

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -37,17 +37,20 @@ import (
...
+	routeAPI                   = "tekton-results-api"
+	secretAPITLS               = "tekton-results-tls"
+	apiContainerName           = "api"
+	boundSAVolume              = "bound-sa-token"
+	boundSAPath                = "/var/run/secrets/openshift/serviceaccount"
+	lokiStackTLSCAEnvVar       = "LOGGING_PLUGIN_CA_CERT"
+	tektonResultWatcherName    = "tekton-results-watcher"
+	postgresStatefulSetName    = "tekton-results-postgres"
+	postgresContainerName      = "postgres"
+	postgresUpgradeScriptsName = "postgres-upgrade-scripts"
```
savitaashture, line 53: I see these constants are in 2 places can we keep it in common as its common to Openshift. + Kubernetes

pkg/reconciler/kubernetes/tektonresult/transform.go
```diff
@@ -675,3 +679,90 @@ func updateEnvWithDBSecretName(props v1alpha1.ResultsAPIProperties) mf.Transform
 		return nil
 	}
 }
+
+// injectPostgresUpgradeSupport modifies the postgres container to support automatic
+// upgrade from PostgreSQL 13 to PostgreSQL 15. The wrapper script checks the PG_VERSION
+// file in the data directory and sets POSTGRESQL_UPGRADE=copy if an upgrade is needed,
+// triggering the sclorg container's built-in upgrade mechanism.
+func injectPostgresUpgradeSupport() mf.Transformer {
```
savitaashture: I see this 
```
// Note: PostgreSQL upgrade transformer is NOT needed for Kubernetes
```
But function is defined in Kubernetes do we still need this ??

savitaashture (commented):
Thank you @enarha 
1. Can you also add a screenshot for the upgrade success scenario
2. Can you describe why we are going with PostgreSQL 15 and not PostgreSQL 16

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
enarha: The reason for using v15 and not v16 is that only upgrade from subsequent versions. There was no version 14, so the tooling provided by the image only works from v13 to v15. You can read about it at the original image doc:  https://github.com/sclorg/postgresql-container/tree/master/15/root/usr/share/container-scripts/postgresql#upgrading-database . It says `This image supports automatic upgrade of data directories created by the PostgreSQL server version 13 (and only this version)` and I also tested that and it is indeed the case.
BTW v15 is supported until November 11, 2027, so we have enough time to plan the next upgrade to v16.

I did not get the note about the screenshot. Add screenshot of what?

pkg/reconciler/kubernetes/tektonresult/transform.go
enarha: No. This is a leftover and I removed it. Initially the update was implemented for both k8s and OCP, but then during testing I found in k8s we use images from bitnami which do not have those v13, v15 etc versions, so the issue does not affect them.

pkg/reconciler/openshift/tektonresult/extension.go
enarha: They are only required in OCP. I removed them from the k8s file.

jkhelil:
@enarha there is a mecanism in the operator specific to openshift  that does generate the bundle csv file and inject IMAGE_RESULTS_POSTGRES env variable to operator container, so the image in results manifest gets replaced by this env var (a way to let user configure a custom image for any component)
youneed to change registry.redhat.io/rhel9/postgresql-16 to registry.redhat.io/rhel9/postgresql-13, in operatorhub/openshift/config.yaml  in main repo ( kind of revert of postgresql-16 pin done before)

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
savitaashture: Thank you for the clarification

> I did not get the note about the screenshot. Add screenshot of what?

I mean you did Upgrade from 13 to 15 na that screenshot if you attach it gives confidence for reviewer to merge the PR

enarha: [v15.log](https://github.com/user-attachments/files/23677453/v15.log)
@savitaashture I do not have a screenshot, but fortunately I stored the log files from testing, so he attached is the log from the actual directory structure upgrade.

jkhelil:
@anithapriyanatarajan @mbpavan can you have a look and lgtm please ?

anithapriyanatarajan:
@enarha - Is it mandatory that the customer has to migrate to v15 for results to work. If the customer intends to remain in v13 is that possible?

enarha:
> @enarha - Is it mandatory that the customer has to migrate to v15 for results to work. If the customer intends to remain in v13 is that possible?

If you are asking about extrernal database, no, it's not required for the customer to update, they can continue using v13 (even though it's EOL since November 13, 2025).
All default deployments (using the Postgres image provided by the operator) will upgrade. There is no way to opt out if that's what you are asking.

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
```diff
@@ -10,6 +10,55 @@ metadata:
...
 apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/name: tekton-results-postgres
+    app.kubernetes.io/part-of: tekton-results
+  name: postgres-upgrade-scripts
+  namespace: tekton-pipelines
+data:
+  postgres-wrapper.sh: |
```
anithapriyanatarajan, line 21: @enarha - If I understood this right, this would be a one time script we are providing to improve user experience. I have a few concerns with this approach:
This violates Operator principles & might pave  way for security backdoors. 

If possible can we include this as a init container or consider a documentation only approach and this script in a blog and provide it in downstream documentation. 

If we are very particular to incoporate this scirpt, please check the possibility of marking this `immutable` https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable. Thankyou.

enarha: Sure, I made the configmap immutable.
Re your other concerns, the automation is used just once and it leverages the internal upgrade mechanism provided by the Postgres image. It is idempotent, so it if called multiple times, it'll have an effect just once. We can either remove that after two releases, or use that for the subsequent upgrades to v16 etc, we just need to bump the from and to versions. For the record, initially I implemented that as a init container, but since the tools which actually migrate the data are part of the Postgres image, the wrapper script was always to be used as a second step, so it make more sense to make it a single step.
Without that automation, we introduce a breaking change and I would prefer a simple automation, than a breaking change and reliance on manual change from the users which requires disabling the operator function which will otherwise override such manual change.

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
```diff
@@ -10,6 +10,56 @@ metadata:
...
+            echo "PostgreSQL data is already version $CURRENT_VERSION - starting normally"
+        else
+            echo "WARNING: Unexpected PostgreSQL version $DATA_VERSION"
+        fi
+    else
+        echo "No existing data found - PostgreSQL will initialize new database"
+    fi
+
+    # Execute the standard PostgreSQL container entrypoint
+    exec run-postgresql
```
mbpavan, line 59: @enarha I have a question. What happens when the upgrade fails? from here - https://github.com/sclorg/postgresql-container/tree/master/15/root/usr/share/container-scripts/postgresql#upgrading-database I see that "** Warning! Before deciding to upgrade the data directory, ensure that you have carefully backed up all your data and are prepared for potential manual rollback! *" 

If the upgrade fails, does it attempt to revert changes, or is the data directory left in an inconsistent state? Specifically, what does the 'manual rollback' process entail in that scenario? Does this also mean there is a prereq step that all the data is backedup ?

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
anithapriyanatarajan: NIT: Is there a possibility to delete this configmap automatically once the upgrade is completed successfully? otherwise recommend to add text in release notes or in downstream documentation to remove the Configmap

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
enarha: It's up to the users to backup their data and that was always the case. Furthermore we do not recommend using that type of deployment (using Postgres default deployment) for production environments and users should ideally use an external DB and manage backups and version upgrades themselves while the default deployment is intended as a test/preview only. This was also clarified in the Release notes and we further emphasize on it in the documentation. With that said, we use the safer (and slower option) POSTGRESQL_UPGRADE="copy" which do not remove the original data structure before the transformation is completed. If the transformation fails and the only practical reason I can think of this happen is not having enough free space, then extending the PVC and deleting the pod will be enough to re-trigger that process. In terms of prereq steps I would recommend backing up the data and verifying enough free space is available. I haven't dug deeper into the various edge cases as we only use what's provided by the Postgres maintainer, we just trigger it.

cmd/openshift/operator/kodata/static/tekton-results/internal-db/db.yaml
enarha: Technically yes. The problem is it has to be done from the same wrapper script and has to be done before we actually start Postgres which is the last thing we do there. So there is a chance we remove the configmap, then Postgres does not start and it makes it harder for the user to re-try the process. IMO this beats the benefit of the cleanup. We can also add more code to the operator to do the cleanup, but it's just adding more code which is not directly related to what the operator is intended to do. That's IMO.

PR #3014, opened 2025-11-26
fix: replace webhook namespaceselector value to target namespace

anithapriyanatarajan:
# Changes

This pull request introduces a new transformer to handle namespace replacement in `ValidatingWebhookConfiguration` resources and adds unit tests to verify its behavior.

## Background

During release testing of the event-based pruner, we found that webhook validations were not being applied. The root cause was that the namespaceSelector in the webhook configuration always pointed to tekton-pipelines, regardless of the target namespace. This transformer resolves that issue by ensuring the namespaceSelector is updated appropriately during resource transformation.

### Namespace selector transformation

* Added the `ReplaceNamespaceInWebhookNamespaceSelector` transformer to `pkg/reconciler/common/transformers.go`, which replaces occurrences of the default namespace in `namespaceSelector.matchExpressions.values` within `ValidatingWebhookConfiguration` resources with the target namespace. This ensures correct scoping of webhook configurations when deploying to custom namespaces.
* Integrated the new transformer into the main `transformers` function so it is applied during resource reconciliation.

### Testing improvements

* Added a dedicated unit test `TestReplaceNamespaceInWebhookNamespaceSelector` in `pkg/reconciler/common/transformers_test.go` to verify that the transformer correctly replaces the namespace in `In` selectors, leaves `NotIn` selectors unchanged, and does not modify unrelated fields like `clientConfig.service.namespace`.… target namespace

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```
/kind bug

Assisted-by: Co-Pilot(claude Sonnet 4.5)

### DIFF
--- a/pkg/reconciler/common/transformers.go
+++ b/pkg/reconciler/common/transformers.go
@@ -76,6 +76,7 @@ func transformers(ctx context.Context, obj v1alpha1.TektonComponent) []mf.Transf
 		injectNamespaceCRDWebhookClientConfig(obj.GetSpec().GetTargetNamespace()),
 		injectNamespaceCRClusterInterceptorClientConfig(obj.GetSpec().GetTargetNamespace()),
 		injectNamespaceClusterRole(obj.GetSpec().GetTargetNamespace()),
+		ReplaceNamespaceInWebhookNamespaceSelector(obj.GetSpec().GetTargetNamespace()),
 		AddDeploymentRestrictedPSA(),
 	}
 }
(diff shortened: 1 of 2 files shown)


jkhelil:
@mbpavan Please have a look

pkg/reconciler/common/transformers.go
```diff
@@ -1246,3 +1247,77 @@ func getSortedKeys(input map[string]interface{}) []string {
 	sort.Strings(keys)
 	return keys
 }
+
+// replaces the namespace in ValidatingWebhookConfiguration namespaceSelector
+func ReplaceNamespaceInWebhookNamespaceSelector(targetNamespace string) mf.Transformer {
```
pramodbindal, line 1252: The logic seems specific to  pruner. should we move it  pruner instead of keeping in common

anithapriyanatarajan: This is not specific to pruner. In general for any validationwebhook config, if there is a namespaceselector and the target namespace is different from tekton-pipelines this will be applicable based on what the target namespace is . For openshift it is openshift-pipelines and in similar way if the user selects any other target namespace, that will be used

PR #3151, opened 2026-01-12
WIP: Inject tls policy results

enarha:
# Changes
The change include two commits. The first one adds generic TLS profiles centrally managed on the cluster level as OpenShift APIServer resources. It allows each Tekton component to use that functionality and set the TLS policy for their services.
The second change enables the centralized TLS profile for Tekton Results.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/kubernetes/tektonresult/tektonresult.go
+++ b/pkg/reconciler/kubernetes/tektonresult/tektonresult.go
@@ -320,10 +320,26 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tr *v1alpha1.TektonResul
 	} else {
 		// If target namespace and version are not changed then check if spec
 		// of TektonResult is changed by checking hash stored as annotation on
-		// TektonInstallerSet with computing new hash of TektonResult Spec
+		// TektonInstallerSet with computing new hash of TektonResult Spec.
+		// Also include external TLS profile state to trigger updates when
+		// platform TLS configuration changes.
 		logger.Debug("Checking for spec changes in TektonResult")
-		// Hash of TektonResult Spec
-		expectedSpecHash, err := hash.Compute(tr.Spec)
+
+		// Get platform TLS profile fingerprint if extension supports it (empty on vanilla k8s)
+		var tlsFingerprint string
+		if fp, ok := r.extension.(common.TLSProfileFingerprinter); ok {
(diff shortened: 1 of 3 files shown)


jkhelil:
@enarha I've reviewed your implementation and wanted to share some feedback and an alternative approach I've been working on.
we will have to use GetTLSEnvVarsFromAPIServer() calls the APIServer directly on every reconcile loop for every component. This creates unnecessary load on the API server and can impact performance at scale.
Why not use the observer pattern as suggested in techincal documentation : This is the standard pattern for OpenShift operators that need to observe cluster configuration:

All use library-go/pkg/operator/configobserver for exactly this purpose.

However i think it is smart to use the env var approch rather than configmaps
I have a proposal here with 
- Use Observer Pattern to watch APIServer TLS profile changes via informer (library-go recommended pattern)
- Fetch TLS Profile from APIServer at operator startup using library-go/pkg/operator/configobserver/apiserver.ObserveTLSSecurityProfile()
- Transform Ciphers from OpenSSL names to IANA names, then to Go-compatible tls.Config with proper cipher suite IDs
- Store in Context as *tls.Config and hash for efficient caching and change detection across all reconcilers
- Inject via Transformer into all component containers (Pipeline, Triggers, Results, Chains, PAC) as environment variables

https://github.com/tektoncd/operator/pull/3166

jkhelil:
few concerns regarding the implementation 
- I believe the use of library-go observer pattern is much simpler and avoid us api server watching/lister 
- regarding the fact that we will have to use an old version, of library-go, i think it is fine, we can use an old version, to just get the tlsprofile from observr then use ciphers conversion function inside our code to convert to go ciphers (this is working even for modern 1.3 tls version), old version is actually relaeted to kubernetes 0.32., we can updatethe version once pipeline uses kubernetes 0.34
- i think it is right to let the component reconciler call the observer to get the tls profile, so we could watched without having to relay on operator
- One thing in your implementation that we should change, is modifying the component CR, this is break the operator architecture, as the only thing that we update is tektonconfig and from there, the operator is generating the CRs for component, so i would like to not break this architecture to ensure operator stabilitty
- Can you please provide the implementation for the rest of component, we need to inject environement variable for all the containers handled by the operator

enarha:
Closed in favor of https://github.com/tektoncd/operator/pull/3225

PR #3197, opened 2026-02-04
feat(result): disable watcher and retention-policy-agent on Hub cluster

ab-ghosh:
# Changes

When multicluster is enabled (`multi-cluster-disabled: false`) and the cluster role is set to Hub (`multi-cluster-role: Hub`), automatically set replicas to 0 for `tekton-results-watcher` and `tekton-results-retention-policy-agent` deployments.

On a Hub cluster:
- The watcher is not needed as there are no local pipeline runs to watch
- The retention-policy-agent is not needed as cleanup happens on Spoke clusters

This is achieved by injecting deployment options in `GetTektonResultCR()` when creating the TektonResult CR from TektonConfig, leveraging the existing options mechanism.

- Inject deployment options to set replicas=0 for watcher and retention-policy-agent on Hub clusters
- Add unit tests for the hub cluster result configuration
- Add e2e tests for the hub cluster result configuration

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/shared/tektonconfig/result/result.go
+++ b/pkg/reconciler/shared/tektonconfig/result/result.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 	"knative.dev/pkg/apis"
+	"knative.dev/pkg/logging"
 )
 
 // This Ensure TektonResult CR is exist or not
@@ -171,7 +172,7 @@ func GetTektonResultCR(config *v1alpha1.TektonConfig, operatorVersion string) *v
 	// for watcher and retention-policy-agent deployments
 	if !config.Spec.Scheduler.MultiClusterDisabled &&
 		config.Spec.Scheduler.MultiClusterRole == v1alpha1.MultiClusterRoleHub {
-		result = injectHubClusterOptions(result)
+		result = disableWatcherAndRetentionAgentOnHubCluster(result)
 	}
 
 	return &v1alpha1.TektonResult{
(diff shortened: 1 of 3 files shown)


ab-ghosh:
**Upgrade test results:**

**Before the upgrade**

image

**After the upgrade**

image

Updating the TektonConfig with the scheduler configuration (multi-cluster-disabled: false and multi-cluster-role: Hub) Result's watcher and Retention Policy Agent being scaled down to 0.

image

image

pkg/reconciler/shared/tektonconfig/result/result.go
```diff
@@ -174,7 +186,30 @@ func GetTektonResultCR(config *v1alpha1.TektonConfig, operatorVersion string) *v
...
 			},
-			Result: config.Spec.Result,
+			Result: result,
 		},
 	}
 }
+
+// injectHubClusterOptions modifies the Result options to set replicas to 0
+// for watcher and retention-policy-agent deployments on Hub clusters
+func injectHubClusterOptions(result v1alpha1.Result) v1alpha1.Result {
```
pratap0007: nit: Can we use a more descriptive function name that clearly explains what this function does?

pratap0007:
LGTM

pkg/reconciler/shared/tektonconfig/result/result.go
ab-ghosh: thanks for the suggestion! I've renamed it to **disableWatcherAndRetentionAgentOnHubCluster** which should better convey that the function disables the watcher and retention-policy-agent deployments (by setting replicas to 0) on Hub clusters. Let me know if this works.

pratap0007: @ab-ghosh Thanks for updating it and looks good to me.

pratap0007:
@ab-ghosh Will the results deployment scale down to 1 replica if multi-cluster support is disabled?

ab-ghosh:
@pratap0007 Yes,I tested it. When the Hub condition is no longer met, the replica override is not applied, so the deployments use their default replica count 1.

pkg/reconciler/shared/tektonconfig/result/result.go
```diff
@@ -174,7 +186,30 @@ func GetTektonResultCR(config *v1alpha1.TektonConfig, operatorVersion string) *v
...
-			Result: config.Spec.Result,
+			Result: result,
 		},
 	}
 }
+
+// disableWatcherAndRetentionAgentOnHubCluster modifies the Result options to set replicas to 0
+// for watcher and retention-policy-agent deployments on Hub clusters
+func disableWatcherAndRetentionAgentOnHubCluster(result v1alpha1.Result) v1alpha1.Result {
+	// Initialize the Deployments map if nil
```
jkhelil, line 208: Add debug log

log.Debug("Disabling watcher and retention-policy-agent for Hub cluster")

pkg/reconciler/shared/tektonconfig/result/result.go
```diff
@@ -174,7 +186,30 @@ func GetTektonResultCR(config *v1alpha1.TektonConfig, operatorVersion string) *v
...
 				TargetNamespace: config.Spec.TargetNamespace,
 			},
-			Result: config.Spec.Result,
+			Result: result,
 		},
 	}
 }
+
+// disableWatcherAndRetentionAgentOnHubCluster modifies the Result options to set replicas to 0
+// for watcher and retention-policy-agent deployments on Hub clusters
```
jkhelil: Add comment how the function uses override, it is not obvious
// This function injects deployment overrides into Options.Deployments. These overrides are applied
// during reconciliation and will force the specified deployments to use 0 replicas, regardless of
// the default values from the release manifests.
//
// Note: Even if the Deployments map is empty or nil, we create new entries with zero-value Deployment
// structs that only have Spec.Replicas set. This is intentional - accessing a non-existent map key
// returns a zero-value struct, which we then modify and insert as an override.

pkg/reconciler/shared/tektonconfig/result/result.go
ab-ghosh: Thank you, added.

pkg/reconciler/shared/tektonconfig/result/result.go
ab-ghosh: ACK, updated the comment.

ab-ghosh:
@jkhelil I have addressed your review comments. Kindly re-review when you get a chance.

pramodbindal (approved):

PR #3225, opened 2026-02-17
Centralized TLS config

enarha:
# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Add support for centralized TLS configuration on OpenShift. When
`enableCentralTLSConfig` is set to `true` in `TektonConfig.spec.platforms.openshift`,
Tekton components automatically inherit TLS settings (minimum version, cipher suites)
from the cluster's APIServer TLS security profile. Changes to the APIServer profile
are detected and propagated to components without manual intervention. TektonResult
is the first component to support this; other components will follow.
```


enarha:
This is still missing documentation, maybe add few more tests, but in general it's tested working and ready for review.

jkhelil:
@enarha Please add PR deescrtipion with changes, architecture diagram,

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,229 @@
+/*
+Copyright 2024 The Tekton Authors
```
jkhelil: s/2024/2026

pkg/reconciler/openshift/common/tlsprofile_test.go
```diff
@@ -0,0 +1,167 @@
+/*
+Copyright 2024 The Tekton Authors
```
jkhelil: s/2024/2026

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,229 @@
...
+	sharedListerMu.Lock()
+	defer sharedListerMu.Unlock()
+	sharedAPIServerLister = lister
+	sharedConfigClient = client
+}
+
+// GetTLSEnvVarsFromAPIServer fetches the TLS security profile from the OpenShift APIServer
+// resource and converts it to environment variable values that can be used by Tekton components.
+// Returns nil if the APIServer has no TLS profile configured or if the shared lister is not initialized.
+func GetTLSEnvVarsFromAPIServer(ctx context.Context, _ *rest.Config) (*TLSEnvVars, error) {
```
jkhelil: Can we return tls.config here from crypto/tls
We Could need other configration in the future
We could use then func TLSEnvVarsFromConfig(cfg *tls.Config) *TLSEnvVars  to turn it to envvars

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,229 @@
...
+	// Note: ObserveTLSSecurityProfile requires:
+	// - non-nil recorder: it calls recorder.Eventf() to log changes
+	// - non-nil existingConfig: it reads from it via unstructured.NestedString()
+	// TODO: Once library-go is updated to a newer version (with TLS 1.3 cipher support),
+	// the supplementTLS13Ciphers workaround below can be removed.
+	existingConfig := map[string]interface{}{}
+	recorder := events.NewLoggingEventRecorder("tekton-operator")
+	observedConfig, errs := apiserver.ObserveTLSSecurityProfile(listers, recorder, existingConfig)
+	if len(errs) > 0 {
+		logger.Warnf("Errors observing TLS security profile: %v", errs)
```
jkhelil: should we return error here

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,229 @@
...
+		for _, c := range ciphers {
+			if cs, ok := c.(string); ok {
+				cipherSuites = append(cipherSuites, cs)
+			}
+		}
+	}
+
+	// If no TLS configuration is present, return nil
+	if minVersion == "" && len(cipherSuites) == 0 {
+		return nil, nil
```
jkhelil, line 174: should we log and return error ?

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,229 @@
...
+		apiServer, err := lister.Get("cluster")
+		if err == nil && apiServer.Spec.TLSSecurityProfile != nil {
+			cipherSuites = supplementTLS13Ciphers(apiServer.Spec.TLSSecurityProfile, cipherSuites)
+		}
+	}
+
+	return &TLSEnvVars{
+		MinVersion:       convertTLSVersionToEnvFormat(minVersion),
+		CipherSuites:     strings.Join(cipherSuites, ","),
+		CurvePreferences: "", // Will be populated once openshift/api#2583 is merged
```
jkhelil: can we mention this in PR description https://github.com/openshift/api/pull/2583/
because empty curve preferences means go defaults

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,229 @@
...
+	}, nil
+}
+
+// convertTLSVersionToEnvFormat converts library-go TLS version format (VersionTLSxx) to
+// the format expected by Go's crypto/tls (1.x)
+func convertTLSVersionToEnvFormat(version string) string {
```
jkhelil: We should consider parse TLSVversion to go format, webhook for example are expecting this in go native constant , I suppose the same thing for go apis using tls
https://github.com/knative/pkg/blob/main/webhook/webhook.go#L53

A conversion in this format feels more suitable to error prone constant
func parseTLSVersion(version string) (uint16, error) {
	switch version {
	case "VersionTLS10", "TLSv1.0":
		return tls.VersionTLS10, nil
	case "VersionTLS11", "TLSv1.1":
		return tls.VersionTLS11, nil
	case "VersionTLS12", "TLSv1.2":
		return tls.VersionTLS12, nil
	case "VersionTLS13", "TLSv1.3":
		return tls.VersionTLS13, nil
	default:
		return 0, fmt.Errorf("unknown TLS version: %s", version)
	}
}

pkg/reconciler/openshift/common/tlsprofile.go
theakshaypant: Question: Should it not be 2026?

pkg/reconciler/openshift/common/tlsprofile_test.go
theakshaypant: Same here.

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -101,18 +115,65 @@ func (oe openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Tr
...
+	return tlsEnvVars
+}
+
+// GetHashData returns TLS config fingerprint for hash computation.
+// This ensures installer set is updated when TLS config changes.
+func (oe *openshiftExtension) GetHashData() string {
+	if oe.resolvedTLSConfig == nil {
+		return ""
+	}
+	return fmt.Sprintf("%s:%s:%s", oe.resolvedTLSConfig.MinVersion, oe.resolvedTLSConfig.CipherSuites, oe.resolvedTLSConfig.CurvePreferences)
```
jkhelil: can we use a real hash function for the fingerprint like sha256, the ":" thingy cna be present in ciphers ro curves

pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -470,3 +531,69 @@ func injectPostgresUpgradeSupport() mf.Transformer {
 		return nil
 	}
 }
+
+// injectTLSConfig injects the TLS configuration as environment variables into the Results API deployment
```
jkhelil: this should be common i think

pkg/reconciler/openshift/tektonconfig/controller.go
```diff
@@ -46,5 +55,137 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	}); err != nil {
 		logger.Panicf("Couldn't register OpenShiftPipelinesAsCode informer event handler: %w", err)
 	}
+
+	// Setup APIServer TLS profile watcher
+	// When the cluster's TLS security profile changes, enqueue TektonConfig for reconciliation
+	if err := setupAPIServerTLSWatch(ctx, ctrl); err != nil {
+		// On OpenShift clusters the APIServer resource should always exist.
+		// This env var is an escape hatch for edge cases and must be explicitly enabled.
+		if os.Getenv("SKIP_APISERVER_TLS_WATCH") == "true" {
```
jkhelil: may be move this to a constant

pkg/reconciler/kubernetes/tektonresult/installerset.go
```diff
@@ -33,11 +33,18 @@ func (r *Reconciler) createInstallerSet(ctx context.Context, tr *v1alpha1.Tekton
...
 
-	// compute the hash of tektonresult spec and store as an annotation
-	// in further reconciliation we compute hash of td spec and check with
-	// annotation, if they are same then we skip updating the object
+	// compute the hash of tektonresult spec (including platform-specific data)
+	// and store as an annotation. In further reconciliation we compute hash
+	// and check with annotation, if they are same then we skip updating the object
 	// otherwise we update the manifest
-	specHash, err := hash.Compute(tr.Spec)
+	hashInput := struct {
```
jkhelil, line 40: This seems redudant and duplciate for all component
Can we have a common function that computes Hash give spec and common extension

Would make thing clear when redaing

pkg/reconciler/openshift/tektonresult/extension.go
jkhelil: Ok  first look, thaught that this function is calculating a hash, but hte hash is calculted in isntallerset code
mm I think the name is misleading here, could we called GetPlatformData rather than GetHash Data

pkg/reconciler/openshift/common/tlsprofile.go
enarha: We can. But we'll need to do the conversation twice APIServer string -> tls.Config -> env var strings instead of APIServer string -> env var strings. If you think we can immediately use the tls.Config with a specific consumer in mind (webhooks, nginx or whatever), then I'll follow your suggestion. If we do not have specific case immediately waiting for that, then I'll suggest we add this when we need it.
Please let me know what do you think.

pkg/reconciler/openshift/common/tlsprofile.go
enarha: Good question. It is a matter of policy - what do we want to happen in the case of a failure. I encountered two types of errors. In the first case we are able to read the APIServer, but failed to populate the map existingConfig with values. It was a bug in the way library-go handles this case in the older version we use, which I fixed, thus we have that two level check. The second case is we fail to read the APIServer resource. If we return an error and bubble it up, we'll break the reconciliation of the operands (e.g. TektonResult). The current behavior is to log a warning, degrade the TLS funcitonality (centralized TLS configuration does not have an effect), but operands continue working. The question is if we prefer that harder or softer form of failure. I believe with the hard failure, the Tekton administrator can still disable the centralized TLS configuration by disabling it through the TektonConfig and if that's the case I'm inclined to agree with a hard failure. WDYT?

docs/OpenShiftCentralizedTLSManagement.md
```diff
@@ -0,0 +1,93 @@
...
+
+### 2. APIServer Watcher
+
+- Single centralized watcher in TektonConfig controller monitors the APIServer cluster resource
+- Uses a shared informer with 30-minute resync interval
+- When APIServer TLS profile changes, enqueues TektonConfig for reconciliation
+
+### 3. Extension Interface Enhancement
+
+- Added `GetHashData() string` method to the Extension interface
```
jkhelil: can you use the new function name GetPlatformData

pkg/reconciler/openshift/tektonconfig/controller.go
```diff
@@ -46,5 +55,138 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+	}); err != nil {
+		return err
+	}
+
+	// Start the informer factory
+	configInformerFactory.Start(ctx.Done())
+
+	// Wait for caches to sync
+	if !cache.WaitForCacheSync(ctx.Done(), apiServerInformer.Informer().HasSynced) {
+		logger.Warn("Failed to sync APIServer informer cache")
```
jkhelil: What happens if the cache is not synced (because of network issue or ai server not reachable), is the code safe on this ? We should probably retrigger a reconcile if cache is not synced, but that will be more complex as for now, let just verify code is safe nothing will panic just after, may be we need to add unit test

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,375 @@
...
+	Get(name string) (*v1alpha1.TektonConfig, error)
+}
+
+// ResolveCentralTLSToEnvVars checks whether central TLS config is enabled in TektonConfig,
+// fetches the raw profile from the shared APIServer lister, and converts it to env vars.
+// Returns (nil, nil) if central TLS is disabled or no TLS config is available.
+func ResolveCentralTLSToEnvVars(ctx context.Context, lister TektonConfigLister) (*TLSEnvVars, error) {
+	tc, err := lister.Get(v1alpha1.ConfigResourceName)
+	if err != nil {
+		return nil, nil
```
jkhelil: Here we should propagate the error: failed to get tektonconfig

jkhelil:
@enarha can we add in this PR transformation for other components
we will need to inject tls env vars to pipeline,, pac, triggers,

pkg/reconciler/openshift/common/tlsprofile.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	// Build a set of already observed ciphers for quick lookup
+	observedSet := make(map[string]bool)
+	for _, c := range observedCiphers {
+		observedSet[c] = true
+	}
+
+	// TLS 1.3 cipher suite names (IANA names)
+	tls13Ciphers := map[string]bool{
```
jkhelil, line 388: should we add this 
As per this https://go.dev/blog/tls-cipher-suites#:~:text=Go%20does%20allow%20configuring%20cipher,order%20by%20default%2C%20unless%20Config.

ciphers are not configrable. with tls 1.3, because they are considered as safe
Now the issue if we add them here, that may procude issues, right ?

jkhelil:
@pratap0007 @anithapriyanatarajan @mbpavan Can you review please ?

pkg/reconciler/openshift/common/tlsprofile.go
enarha: Switched to function which returns an error if unknown tls version is set through the apiserver record. Also split the retrieval of the tls configuration from the transformation. If we need to add more transformations in the future (e.g. to tsl.Config) we can do that without affecting existing code.

pkg/reconciler/kubernetes/tektonresult/installerset.go
enarha: I kept it per component for now. Each component has specific deployments to update. I considered a separate function which takes a resource type (e.g. deployment/statefulset) and resource object (e.g. tekton-results-api), but it requires any type params and extra import of the common package, so I think for now the advantages are not much bigger than the disadvantages.

pkg/reconciler/openshift/tektonconfig/controller.go
enarha: From what I see, nothing will panic if the `cluster` apiserver resource is not found in the cache. The reconciliation of the operand (e.g. tektonresult) will fail repeatedly though. There will be some very rapid reconcile retries with an exponential backoff. The same goes for the cache sync. The admin can still disable the central TLS configuration. If they opt-in into that feature, then I think louder failure without a panic makes sense. If you prefer we do not affect the operand reconciliation, then we can ignore that error and threat it like "no TLS config available" in the APIServer resource and just log some warning.

pkg/reconciler/openshift/common/tlsprofile.go
enarha: done.

pkg/reconciler/openshift/common/tlsprofile.go
enarha: The reason this was included in the first place is that the newer version of library-go does the same. They are commented out in the 2023 version, but in the newer ones they are not.
They are not configurable indeed and not required for golang based operands. But we have some non golang based dependencies like PostgreSQL and Nginx maybe more. I'm not sure how will these handle the 1.3 ciphers if we do not explicitly specify them. So that's one reason to include them. For Tekton Results we already have the code which handles the ciphers list and there the 1.3 ciphers are simply ignored. To me it makes more sense to keep these for now. WDYT?

jkhelil: Yes we can have the "ignore mecansim for our go apps" centralised, not just for results

enarha:
/hold
Waiting for more reviews and after getting the required approvals I'll do another cycle of e2e testing, thus the hold.

enarha:
@jkhelil I think this PR is good to be merged. I created and used a comprehensive testing plan including copy-paste set of commands to test the different scenarios. It can be found here: docs/testing/CentralTLSConfigTestPlan.md , I make it a separate commit as it requires both the common functionality + the Tekton Results functionality to check that TLS configuration is actually propagated to deployment(s).

jkhelil:
@anithapriyanatarajan @mbpavan any chance you could have a final look to this or lgtm ?

enarha:
> @anithapriyanatarajan @mbpavan any chance you could have a final look to this or lgtm ?

kind reminder.

anithapriyanatarajan:
@enarha - could you rebase this PR.

jkhelil:
@anithapriyanatarajan @mbpavan any chance to have lgtm here ?

anithapriyanatarajan:
@enarha marked to be on hold to check the possibility of refactoring the implementation to be inclusive of plain kubernetes users as well. Basically to check the option to have `tektonconfig`  as central  to manage what TLS version and cipher suites to use. For Openshift, this gets populated by the API server CR watch and for Kubernetes, admins can manually configure as preferred. Thank you.

enarha:
> @enarha marked to be on hold to check the possibility of refactoring the implementation to be inclusive of plain kubernetes users as well. Basically to check the option to have `tektonconfig` as central  to manage what TLS version and cipher suites to use. For Openshift, this gets populated by the API server CR watch and for Kubernetes, admins can manually configure as preferred. Thank you.

Personally I'm against such approach until we get a clean evidence such requirement exists. That will complicate implementation and maintenance with no clear benefit (YAGNI). To be honest, before implementing it that way, I researched the topic (Kubernetes way of setting application configuration cluster wide) and found nothing specific, so I decided to stick to the current requirements. That also goes beyond the scope of Tekton, so I do not see how we can influence Kubernetes even if add support for it. That's IMO.

anithapriyanatarajan:
> > @enarha marked to be on hold to check the possibility of refactoring the implementation to be inclusive of plain kubernetes users as well. Basically to check the option to have `tektonconfig` as central  to manage what TLS version and cipher suites to use. For Openshift, this gets populated by the API server CR watch and for Kubernetes, admins can manually configure as preferred. Thank you.
> 
> Personally I'm against such approach until we get a clean evidence such requirement exists. That will complicate implementation and maintenance with no clear benefit (YAGNI). To be honest, before implementing it that way, I researched the topic (Kubernetes way of setting application configuration cluster wide) and found nothing specific, so I decided to stick to the current requirements. That also goes beyond the scope of Tekton, so I do not see how we can influence Kubernetes even if add support for it. That's IMO.

In its current form the variables (TLS_MIN_VERSION, TLS_CIPHER_SUITES, TLS_CURVE_PREFERENCES) are directly read from API server and Injected to Envars. This is tightly coupled to openshift API. My request is to have this stored to the tektonconfig CR and mark it as populated manually / automated. This way an admin gets an opportunity to override if required and also for kuberenetes plain version, admins can set this manually as required.  This way any other tekton components consumption model from Operator remains intact even if the source for reading the policy changes.

anithapriyanatarajan:
@enarha - Thank you for the PR. Approving the changes limiting the scope of Cryptoagility to Openshift.

docs/OpenShiftCentralizedTLSManagement.md
```diff
@@ -0,0 +1,108 @@
+# Centralized TLS Configuration Support
+
+This change adds support for centralized TLS configuration from OpenShift's APIServer resource, enabling Tekton components to inherit TLS settings (minimum version, cipher suites, curve preferences) from the cluster-wide security policy.
```
zakisk, line 3: @enarha this is technical doc but the start of the sentence looks like release note or commit message "This change adds support..."

PR #3228, opened 2026-02-19
fix: proxy-webhook selector matches operator pods

### LINKED ISSUE #3227: tekton-operator-proxy-webhook Service selector matches operator pods, causing ~50% webhook admission failures
## Environment

- Tekton Operator version: v0.78.1
- Kubernetes version: v1.33
- Platform: Kubernetes (bare-metal)

## Description

The `tekton-operator-proxy-webhook` Service uses `name: tekton-operator` as
its pod selector. This label is also present on pods of the main
`tekton-operator` Deployment. As a result, the Service load-balances admission
webhook traffic across **both** Deployments, even though `tekton-operator`
pods do not listen on port 8443.

This causes approximately **50% of all admission webhook requests** to fail
with `connection refused`, which — because the MutatingWebhookConfiguration
has `failurePolicy: Fail` — immediately rejects the creation of any TaskRun
pod.

## Steps to Reproduce

1. Deploy Tekton Operator v0.78.1 on a Kubernetes cluster.
2. Inspect the endpoints of the `tekton-operator-proxy-webhook` Service:
   ```
   kubectl get endpoints tekton-operator-proxy-webhook -n <namespace>
   ```
3. Observe that the Endpoints list includes pods from **both**
   `tekton-operator` and `tekton-operator-proxy-webhook` Deployments.
4. Trigger any Pipeline/TaskRun. Observe that roughly half of new TaskRun pod
   creation attempts fail immediately.

## Expected Behavior

The `tekton-operator-proxy-webhook` Service should only route traffic to
`tekton-operator-proxy-webhook` pods (port 8443). The `tekton-operator` pods
should never appear in this Service's Endpoints.

## Actual Behavior

The Service Endpoints include pods from both Deployments:

```yaml
# kubectl get endpoints tekton-operator-proxy-webhook -n tekton -o yaml
subsets:
- addresses:
  - ip: 172.26.0.66   # tekton-operator-proxy-webhook pod  ✅ serves on 8443
  - ip: 172.26.1.157  # tekton-operator pod                ❌ does not serve on 8443
  ports:
  - port: 8443
```

Stress-testing the Service directly (20 requests via ClusterIP) showed
roughly 50% failing with `connection refused`:

```
req-1: PASS (HTTP 415)  req-2: FAIL (000)  req-3: FAIL (000)
req-4: FAIL (000)       req-5: FAIL (000)  req-6: PASS (HTTP 415)
req-7: PASS (HTTP 415)  req-8: FAIL (000)  req-9: PASS (HTTP 415)
req-10: FAIL (000)
```

The failure manifests as the following error when creating TaskRun pods:

```
failed to create task run pod "<pod-name>":
Internal error occurred: failed calling webhook "proxy.operator.tekton.dev":
failed to call webhook:
Post "https://tekton-operator-proxy-webhook.<ns>.svc:443/defaulting?timeout=10s":
dial tcp <ClusterIP>:443: connect: connection refused
```

Note: the error appends a misleading hint ("Maybe missing or invalid Task
…") that does not reflect the real cause.

## Root Cause

Both Deployments use the same pod template label `name: tekton-operator`:

**`tekton-operator` Deployment** (`config/kubernetes/base/operator.yaml`):
```yaml
selector:
  matchLabels:
    name: tekton-operator   # ← same label
template:
  metadata:
    labels:
      name: tekton-operator # ← same label
```

**`tekton-operator-proxy-webhook` Deployment**
(`cmd/kubernetes/operator/kodata/webhook/webhook.yaml`):
```yaml
selector:
  matchLabels:
    name: tekton-operator   # ← collision!
template:
  metadata:
    labels:
      name: tekton-operator # ← collision!
```

**`tekton-operator-proxy-webhook` Service**:
```yaml
selector:
  name: tekton-operator     # ← matches both Deployments!
```

The same issue exists in the OpenShift manifest
(`cmd/openshift/operator/kodata/webhook/webhook.yaml`).

## Proposed Fix

Change the proxy-webhook Deployment's `matchLabels` selector and pod template
label from `name: tekton-operator` to `name: tekton-operator-proxy-webhook`,
and update the Service selector to match. The existing `app: tekton-operator`
label is left unchanged.

I have a patch ready and will submit a PR.

bowling233:
# Changes

Fixes #3227

Both the `tekton-operator` and `tekton-operator-proxy-webhook` Deployments
label their Pods with `name: tekton-operator`. The
`tekton-operator-proxy-webhook` Service uses this same label as its only
selector, so it inadvertently load-balances traffic across both Deployments.
Because `tekton-operator` pods do not serve on port 8443, ~50% of admission
webhook requests fail with `connection refused`. Since the
MutatingWebhookConfiguration has `failurePolicy: Fail`, each failure
immediately rejects TaskRun Pod creation.

**Changes:**

- `cmd/kubernetes/operator/kodata/webhook/webhook.yaml`: rename the
  proxy-webhook Deployment's `matchLabels` selector and pod template label
  from `name: tekton-operator` to `name: tekton-operator-proxy-webhook`;
  update the Service `selector` to match.
- `cmd/openshift/operator/kodata/webhook/webhook.yaml`: same change for the
  OpenShift manifest.

The existing `app: tekton-operator` label is preserved on both Deployments.
No other resources are affected.

**Alternative considered:** adding a set-based (`NotIn`) expression to the
Service selector to exclude `tekton-operator` pods. This was not viable
because Kubernetes Services only support equality-based (`matchLabels`)
selectors.

# Submitter Checklist

review them:

> **Note on tests:** This bug only manifests at the Service routing layer
> (i.e., ~50% of requests land on a pod with no server). There is no
> in-tree unit or integration test that exercises which pods a Service
> selects. A targeted e2e test verifying that the proxy-webhook Service
> endpoints do not include `tekton-operator` pods would be a good addition,
> but is left for a follow-up.

# Release Notes

```release-note
Fix: the tekton-operator-proxy-webhook Service selector incorrectly matched
tekton-operator pods in addition to proxy-webhook pods, causing ~50% of
admission webhook requests to fail with "connection refused" and TaskRun Pod
creation to be rejected. Users on v0.78.1 can work around this until upgrading
by adding `pod-template-hash: <webhook-pod-hash>` to the Service selector.
```

### DIFF
--- a/cmd/kubernetes/operator/kodata/webhook/webhook.yaml
+++ b/cmd/kubernetes/operator/kodata/webhook/webhook.yaml
@@ -92,11 +92,11 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      name: tekton-operator
+      name: tekton-operator-proxy-webhook
   template:
     metadata:
       labels:
-        name: tekton-operator
+        name: tekton-operator-proxy-webhook
         app: tekton-operator
     spec:
       serviceAccountName: tekton-operators-proxy-webhook
@@ -152,7 +152,7 @@ spec:
       port: 443
       targetPort: 8443
   selector:
-    name: tekton-operator
+    name: tekton-operator-proxy-webhook
 
 ---
 
--- a/cmd/openshift/operator/kodata/webhook/webhook.yaml
+++ b/cmd/openshift/operator/kodata/webhook/webhook.yaml
@@ -99,11 +99,11 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      name: tekton-operator
+      name: tekton-operator-proxy-webhook
   template:
     metadata:
       labels:
-        name: tekton-operator
+        name: tekton-operator-proxy-webhook
         app: tekton-operator
     spec:
       securityContext:
@@ -160,7 +160,7 @@ spec:
       port: 443
       targetPort: 8443
   selector:
-    name: tekton-operator
+    name: tekton-operator-proxy-webhook
 
 ---
 
--- a/pkg/reconciler/kubernetes/tektoninstallerset/install.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/install.go
@@ -31,6 +31,7 @@ import (
 	autoscalingv2 "k8s.io/api/autoscaling/v2"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -528,6 +529,18 @@ func (i *installer) ensureResource(ctx context.Context, expected *unstructured.U
 			"expectedHash", expectedHashValue,
 		)
 
+		if selectorChanged, err := deploymentSelectorChanged(expected, existing); err != nil {
+			loggerWithContext.Errorw("failed to compare deployment selectors", "error", err)
+			return err
+		} else if selectorChanged {
+			loggerWithContext.Infow("deployment selector changed, deleting resource before recreate")
+			if err := i.mfClient.Delete(existing); err != nil {
+				loggerWithContext.Errorw("failed to delete deployment with changed selector", "error", err)
+				return v1alpha1.RECONCILE_AGAIN_ERR
+			}
+			return v1alpha1.RECONCILE_AGAIN_ERR
+		}
+
 		err = i.copyResourceFields(expected, existing, reconcileFields...)
 		if err != nil {
 			loggerWithContext.Errorw("failed to copy resource fields", "error", err)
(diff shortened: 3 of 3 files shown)


jkhelil:
@bowling233 , thank for your PR.
- can you check what happens to existing clusters during upgrade? ( Install 0.78.1 and then apply your change)
Please describe and post a proof that upgrade is working and not broken

anithapriyanatarajan:
@bowling233 - Request you to address the review comment, if you are still pursuing this PR. Thank you. :bowing_woman:

bowling233:
Hi @anithapriyanatarajan,

So sorry for the late response! I won't have the bandwidth to properly validate these changes until next month.

I can confirm this approach has side effects—specifically, the **HorizontalPodAutoscaler** is hitting `FailedGetResourceMetric` errors because it's incorrectly picking up the main operator pods, which lack the expected CPU requests in the `tekton-operator-lifecycle` container.

This PR definitely needs more refinement to handle the selector immutability and the HPA configuration. Should I move this to a **Draft** for now, or would you prefer I **close this** and resubmit once I've validated a full fix?

anithapriyanatarajan:
@bowling233 - would it be possible to progress this PR?

bowling233:
@jkhelil Thanks for the review. I checked the upgrade path and also clarified why this issue happens in our environment.

The root cause is that our deployment uses a single namespace for both the operator and its operands. We install the release manifest through Kustomize with `namespace: tekton`, and our `TektonConfig` uses `spec.targetNamespace: tekton`. In this layout, both the `tekton-operator` pod and the `tekton-operator-proxy-webhook` pod can exist in the same namespace.

Before this fix, the proxy webhook Service selected only `name: tekton-operator`, so in this single-namespace layout it could select both pods. The operator pod does not serve the webhook on 8443, which explains the intermittent admission failures. This is probably less visible in the default two-namespace layout because the operator pod and proxy webhook Service are separated by namespace.

While testing upgrade from `v0.78.1`, I found one more issue: changing `Deployment.spec.selector` directly is not upgrade-safe because that field is immutable. The update fails with:

```text
Deployment "tekton-operator-proxy-webhook" is invalid:
spec.selector: Invalid value: {"matchLabels":{"name":"tekton-operator-proxy-webhook"}}:
field is immutable
```

So I added upgrade handling in the TektonInstallerSet reconciler: when an existing Deployment selector differs from the desired selector, the reconciler deletes the old Deployment and lets the next reconcile recreate it with the new selector.

<details>
<summary>Validation details</summary>

I verified the failure mode by simulating the old selector behavior in minikube:

- proxy webhook Service selector: `name=tekton-operator`
- endpoints included both the real proxy webhook pod and a same-namespace non-webhook pod
- creating a pod that matched the proxy webhook `objectSelector` intermittently failed with:

```text
failed calling webhook "proxy.operator.tekton.dev":
Post "https://tekton-operator-proxy-webhook.tekton-pipelines.svc:443/defaulting?timeout=10s":
dial tcp <ClusterIP>:443: connect: connection refused
```

After applying the fix:

- the Deployment selector, pod label, and Service selector all use `name: tekton-operator-proxy-webhook`
- the Service endpoints only include the proxy webhook pod
- matching pod creation succeeds and is mutated by the proxy webhook
- upgrade from `v0.78.1` to this branch completes successfully, with `TektonConfig`, `TektonPipeline`, and `TektonTrigger` all Ready

</details>

PR #3244, opened 2026-02-24
proxy-aae: override WORKERS_SECRET_NAMESPACE for OpenShift

khrm:
On OpenShift, the kueue operator runs in the openshift-kueue-operator namespace rather than the upstream default kueue-system. Add an OpenShift transform in the proxy-aae extension to replace the WORKERS_SECRET_NAMESPACE env var accordingly. Export DeploymentEnvVars in the common package to make the transformer reusable. Add tests covering both the default transform and the ability for spec.options to override the value.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/openshift/tektonmulticlusterproxyaae/extension.go
+++ b/pkg/reconciler/openshift/tektonmulticlusterproxyaae/extension.go
@@ -20,6 +20,8 @@ import (
 	"context"
 
 	mf "github.com/manifestival/manifestival"
+	corev1 "k8s.io/api/core/v1"
+
 	"github.com/tektoncd/operator/pkg/apis/operator/v1alpha1"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	occommon "github.com/tektoncd/operator/pkg/reconciler/openshift/common"
@@ -35,6 +37,12 @@ func (oe openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Tr
 	return []mf.Transformer{
 		occommon.RemoveRunAsUser(),
 		occommon.RemoveRunAsGroup(),
+		common.DeploymentEnvVars([]corev1.EnvVar{
+			{
+				Name:  "WORKERS_SECRET_NAMESPACE",
+				Value: "openshift-kueue-operator",
+			},
+		}),
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/reconciler/openshift/tektonmulticlusterproxyaae/extension.go
```diff
@@ -35,6 +37,12 @@ func (oe openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.Tr
 	return []mf.Transformer{
 		occommon.RemoveRunAsUser(),
 		occommon.RemoveRunAsGroup(),
+		common.DeploymentEnvVars([]corev1.EnvVar{
+			{
+				Name:  "WORKERS_SECRET_NAMESPACE",
+				Value: "openshift-kueue-operator",
```
pramodbindal, line 43: instead of hardcoding the namespace. can we extract this  from some kueue  resource ?

khrm: Isn't the worker secret needed in Kueue's namespace? That's why we need to configure it.

khrm: If it's installed somewhere, we can use options.

pramodbindal: There is always an option to install the kueue operator in non-default namespace.
SO if customer installs there. then this will fail

image

khrm: Customer can set options field on deployment to change this environment variable.

pramodbindal: editing operator owned deployment is not something they should be doing. Maybe we can set this value in tektonconfig if namespace is different from default 

So in scheduler config. can we have a field called  `kueue-namespace` which defaults to  "`openshift-kueue-operator`" in openshift and `kueue-system` in kubernetes

khrm: No, they aren't editing deployment. They are using the options field. Check the document.

khrm: https://github.com/tektoncd/operator/pull/3244/changes#diff-e74377553d2b5814f786cbf51128675a389c7ef3f3555b6c6967deeabc2d89dbR454

PR #3341, opened 2026-04-13
fix: prevent Chain and Result installation on lite profile

### LINKED ISSUE #2781: Bug Report: Profile ‘lite’  is creating extra components
# Description

When deploying Tekton Operator with spec.profile: lite, the expected components are not being correctly filtered. Unnecessary components like Tekton Results and Chains are being installed, even though they are not part of the lite profile.

# Expected Behavior
With spec.profile: lite, the Operator should install only:

   tekton pipelines controllers

# Actual Behavior
NAME                                                     READY   STATUS    RESTARTS   AGE
tekton-chains-controller-5ccfb454f9-cf2v4                1/1     Running   0          4m42s
tekton-events-controller-686ff59c66-29nzx                1/1     Running   0          6m30s
tekton-operator-proxy-webhook-94f785b44-rc7qw            1/1     Running   0          6m29s
tekton-pipelines-controller-d768b6f47-q2ck9              1/1     Running   0          6m30s
tekton-pipelines-remote-resolvers-6fdd86bfb7-8vh7m       1/1     Running   0          6m30s
tekton-pipelines-webhook-6f4784bb55-dmnx2                1/1     Running   0          6m30s
tekton-results-api-6f6f858d79-jrwtc                      1/1     Running   0          4m19s
tekton-results-postgres-0                                1/1     Running   0          4m19s
tekton-results-retention-policy-agent-5865bf5fd4-9qcxr   1/1     Running   0          4m19s
tekton-results-watcher-779bbf74f7-tmq47                  1/1     Running   0          4m19s

1. Install Tekton Operator on a clean Kubernetes/Openshift cluster.

1. Create a TektonConfig with:

kind: TektonConfig
metadata:
  name: config
  namespace: tekton-pipelines
spec:
  profile: lite

3. Observe pods in tekton-pipelines:

oc get pods -n tekton-pipelines

4. Note that Results and Chains pods are still running.

# Additional Info

- Kubernetes version:

Client Version: v1.17.0-4-g38212b5
Server Version: v1.32.0

 

- Tekton Pipeline version:
version: v1.1.0

### LINKED ISSUE #3377: TektonResult stuck on lite → basic profile transition
Surfaced by #3341, where fixing the `lite` profile to correctly *not* install Chain/Result exposed a pre-existing reconciler issue.

### What happens

The e2e test `TestTektonConfigTestSuite/Test02_ChangeProfile` walks the sequence `default → lite → basic → all`. With the lite-profile fix in #3341, the `lite` step deletes the `TektonResult` CR (correct behaviour). The next step (`basic`) then has to recreate it, and the test times out with the controller stuck reporting:

```
config   devel   False   Components not in ready state: TektonResult reconcile again and proceed
```

CI run with the failing transition: https://github.com/tektoncd/operator/actions/runs/24321929589

### Why it wasn't seen before

Before #3341 the `lite` profile silently installed Chain + Result alongside Pipelines (the bug #3341 fixes). The existing Result CR was therefore always present, so the `lite → basic` transition was effectively a no-op for Result and never exercised the (re)create path under those conditions. Once `lite` started actually deleting Result, the re-create on `lite → basic` started failing/looping.

### Where to look

`pkg/reconciler/kubernetes/tektoninstallerset` and the TektonResult reconciler, around the path that goes from `Result deleted` → `Result recreated and waited on for ready`. The "reconcile again and proceed" message comes from the install-set readiness check.

### Acceptance criteria

`TestTektonConfigTestSuite/Test02_ChangeProfile` runs the full `default → lite → basic → all` sequence to completion against #3341 (or its successor) without timing out.

cc @anithapriyanatarajan @khrm @savitaashture

alliasgher:
## Changes

Fix the `lite` profile installing TektonChain and TektonResult components when it should only install TektonPipelines.

Also fixes #3377: TektonResult stuck on `lite -> basic` profile transition.

### Root Cause

The TektonTrigger section correctly gates creation on both the disabled flag AND the profile:
```go
if !tc.Spec.Trigger.Disabled && (tc.Spec.Profile == v1alpha1.ProfileAll || tc.Spec.Profile == v1alpha1.ProfileBasic) {
```

But the TektonChain and TektonResult sections only checked the disabled flag:
```go
if !tc.Spec.Chain.Disabled {   // missing profile check
if !tc.Spec.Result.Disabled {  // missing profile check
```

Since `Chain.Disabled` and `Result.Disabled` default to `false`, the lite profile would create these components, then the FinalizeKind path would try to delete them, causing a reconciliation loop.

### Root Cause (follow-up changes included for  #3377)

With the lite profile correctly deleting TektonResult, switching from `lite → basic` requires recreating it. The TektonResult reconciler was finding the old TektonInstallerSet still present in etcd with a `DeletionTimestamp` (async deletion from FinalizeKind), treating it as the active set, and looping forever since a set under deletion can never become Ready.

### Fix

1. Add the same `ProfileAll || ProfileBasic` check to both Chain and Result conditions, matching
   the existing Trigger pattern.
2. In the TektonResult reconciler, check `DeletionTimestamp` on the fetched InstallerSet and
   requeue if non-nil, so reconciliation only proceeds after the dying set is fully gone.

Fixes #2781
Fixes #3377

## Verification

- `go build ./...` passes
- Lite profile will now only create TektonPipelines (as documented)
- Basic profile continues to create Pipelines + Triggers + Chains + Results
- All profile creates everything including Dashboard

# Submitter Checklist

- [x] Commit messages follow [commit message best practices (https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

```release-note
Fix `lite` profile incorrectly installing TektonChain and TektonResult components; only TektonPipelines should be installed on the lite profile. Also fixes a reconciler loop when transitioning from `lite` to `basic` profile due to a stale TektonInstallerSet with a DeletionTimestamp.
```

/kind bug

Co-Authored-by: Co-pilot(Claude Sonnet 4.6)

### DIFF
--- a/pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
+++ b/pkg/reconciler/kubernetes/tektoninstallerset/tektoninstallerset.go
@@ -176,7 +176,12 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, installerSet *v1alpha1.T
 	if err != nil {
 		logger.Errorw("StatefulSet resources installation failed", "error", err)
 		installerSet.Status.MarkStatefulSetNotReady(err.Error())
-		return r.handleError(err, installerSet)
+		// StatefulSet "not ready" is inherently transient (pods starting up, first-time
+		// DB initialization, etc.). Return a fixed 10-second requeue instead of letting
+		// Knative's exponential backoff push retry intervals to 128s+, which causes
+		// profile transitions (e.g. lite → basic) to time out waiting for postgres.
+		// Real API errors on the ensureResource path are also safe to retry at 10s.
+		return v1alpha1.REQUEUE_EVENT_AFTER
 	}
 
 	// Update Status for StatefulSet Resources
--- a/pkg/reconciler/kubernetes/tektonresult/tektonresult.go
+++ b/pkg/reconciler/kubernetes/tektonresult/tektonresult.go
@@ -278,6 +278,16 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tr *v1alpha1.TektonResul
 		return err
 	}
 
+	// If the installer set is being deleted (e.g. leftover from a previous profile
+	// that included Result, such as lite → basic transition), wait for it to be
+	// fully removed before proceeding. Using a dying set causes the reconciler to
+	// loop forever because the set can never become Ready.
+	if installedTIS.DeletionTimestamp != nil {
+		logger.Infow("Existing installer set is being deleted, waiting for cleanup", "name", existingInstallerSet)
+		tr.Status.MarkInstallerSetNotReady("Waiting for previous installer set to be deleted")
+		return v1alpha1.REQUEUE_EVENT_AFTER
+	}
+
 	installerSetTargetNamespace := installedTIS.Annotations[v1alpha1.TargetNamespaceKey]
 	installerSetReleaseVersion := installedTIS.Labels[v1alpha1.ReleaseVersionKey]
 
--- a/pkg/reconciler/shared/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/shared/tektonconfig/tektonconfig.go
@@ -280,7 +280,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 	}
 
 	// Ensure Chain CR
-	if !tc.Spec.Chain.Disabled {
+	if !tc.Spec.Chain.Disabled && (tc.Spec.Profile == v1alpha1.ProfileAll || tc.Spec.Profile == v1alpha1.ProfileBasic) {
 		tektonchain := chain.GetTektonChainCR(tc, r.operatorVersion)
 		logger.Debug("Ensuring TektonChain CR exists")
 		if _, err := chain.EnsureTektonChainExists(ctx, r.operatorClientSet.OperatorV1alpha1().TektonChains(), tektonchain); err != nil {
@@ -302,7 +302,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 	}
 
 	// Ensure Result CR
-	if !tc.Spec.Result.Disabled {
+	if !tc.Spec.Result.Disabled && (tc.Spec.Profile == v1alpha1.ProfileAll || tc.Spec.Profile == v1alpha1.ProfileBasic) {
 		tektonresult := result.GetTektonResultCR(tc, r.operatorVersion)
 		if platformData := r.extension.GetPlatformData(); platformData != "" {
 			if tektonresult.Annotations == nil {
--- a/pkg/reconciler/shared/tektonconfig/tektonconfig.go
+++ b/pkg/reconciler/shared/tektonconfig/tektonconfig.go
@@ -299,7 +299,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 	}
 
 	// Ensure Chain CR
-	if !tc.Spec.Chain.Disabled {
+	if !tc.Spec.Chain.Disabled && (tc.Spec.Profile == v1alpha1.ProfileAll || tc.Spec.Profile == v1alpha1.ProfileBasic) {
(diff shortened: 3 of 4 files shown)


khrm (commented):
/lgtm

alliasgher:
The e2e test `TestTektonConfigTestSuite/Test02_ChangeProfile` is failing in CI ([run](https://github.com/tektoncd/operator/actions/runs/24321929589)) on this PR. The test sequence is `default → lite → basic → all`, and it's timing out on the `lite → basic` transition with:

```
config   devel   False   Components not in ready state: TektonResult reconcile again and proceed
```

Some analysis on why my fix exposes this:

- **Before this PR**, the `lite` profile silently installed Chain + Result in addition to Pipelines (the bug). So the existing Result CR was always present, and `lite → basic` was effectively a no-op for Result.
- **With this PR**, `lite` correctly deletes Result. The `lite → basic` transition has to re-create it, and that re-creation appears to be hitting a race condition or stuck state (see the controller status above).

This is a real underlying issue in the Result reconciliation/transition flow, but my fix is what surfaced it. Two ways forward:

1. **Land this fix as-is** and treat the transition issue as a separate bug to fix in the Result reconciler.
2. **Leave the buggy behavior in place** until the transition issue is fixed.

I'd appreciate guidance from @khrm / @savitaashture on the preferred path. Happy to investigate the Result reconciler stuck-transition issue as a follow-up if option (1) is preferred.

anithapriyanatarajan:
@alliasgher Thank you for this PR. Would recommend to have the CI passing before the PR is merged. Hence, would it be possible to proceed with your proposed first option in last comment:

**Land this fix as-is and treat the transition issue as a separate bug to fix in the Result reconciler.**.

If you have the bandwidth, please raise a new issue and submit a PR for that. Once merged, we can rerun the tests on this PR. If bandwidth is tight, even a GitHub issue to capture the problem would be helpful.

Very keen to get this PR merged, but want to do it the right way. Feel free to flag any concerns. Thanks again!

alliasgher:
Filed #3377 to capture the TektonResult `lite → basic` transition issue. I'll start looking at the reconciler this week and open a separate PR when I have a fix; happy for this PR to wait on that per your suggestion. @anithapriyanatarajan

anithapriyanatarajan:
> Filed #3377 to capture the TektonResult `lite → basic` transition issue. I'll start looking at the reconciler this week and open a separate PR when I have a fix; happy for this PR to wait on that per your suggestion. @anithapriyanatarajan

@alliasgher Rebased onto current `main` and added a follow-up [commit](https://github.com/tektoncd/operator/pull/3341/changes/ef10bfce9c71e2a6e10a80b2dbff6420b0961185) to address the blocking issue from #3377.

The `lite -> basic` stuck transition was caused by the TektonResult reconciler finding the old `TektonInstallerSet` still present in etcd with a `DeletionTimestamp` after `FinalizeKind` issued the async `DeleteCollection`. The reconciler was treating that dying set as the active one and looping forever since it can never become `Ready`.

Fix: check `DeletionTimestamp != nil` on the fetched InstallerSet and return `REQUEUE_EVENT_AFTER` — reconciliation resumes only after the set is fully gone.

alliasgher:
Thanks for picking up the reconciler investigation @anithapriyanatarajan, and for the clear write-up. I reviewed both commits against current main and they look correct to me.

- The `DeletionTimestamp != nil` guard in `tektonresult.go` is exactly the dying-InstallerSet race from #3377. Returning `REQUEUE_EVENT_AFTER` so we only proceed once the old set is gone is the right call.
- The `tektoninstallerset.go` change makes sense: `handleError` only maps the `RECONCILE_AGAIN_ERR` sentinel to a fixed requeue and lets every other error fall through to Knative's exponential backoff, so postgres first-init would never converge inside the 3-minute CI window. Fixed 10s polling fixes that.

Two small FYIs, neither blocking:
1. That change also routes genuine (non-sentinel) errors on the StatefulSet path to the fixed 10s requeue instead of backing off. Your inline comment already calls this out and I think it is fine for a readiness wait, just noting it.
2. The commit message describes returning `RECONCILE_AGAIN_ERR` from `isStatefulSetAvailable`, but the code returns `REQUEUE_EVENT_AFTER` directly from `ReconcileKind`. Same net effect, might be worth a tweak to the message if you re-push.

LGTM from my side. Happy for this to carry the #3377 fix. My original commit is unchanged by the rebase.

PR #3370, opened 2026-04-23
fix(release): correct devel version rewriting in publish-operator-release-task

anithapriyanatarajan:
# Changes

The publish-operator-release task had  issues with rewriting devel version labels in the release artifacts.

Problems fixed:

1. Removed `pipeline.tekton.dev/release` sed pattern -  this label never exists in this repo (copied from tektoncd/pipeline), making it a no-op.

2. Fixed unquoted devel not being matched - switched from BRE to ERE (-E) and made quotes optional ("?devel"?) to handle both version: devel and version: "devel" which coexist in source files.

3. Added **rewrite-devel-in-source step** (runs before run-kustomize-ko) to handle two cases that are never reachable via release.yaml sed:

- > - name: VERSION / value: "devel" env var in config/${KUBE_DISTRO}/base/operator.yaml , rendered into the container binary, not a k8s label, so not matched by any label-targeted pattern
- > cmd/openshift/operator/kodata/cabundles/*.yaml ,  bundled inside the container image via kodata, never appear in release.yaml

4. Uses awk (not sed) for the **multi-line VERSION** rewrite since the ko-gcloud image does not ship GNU sed. This is unique to Operator and does not apply to rest of the components.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/tekton/build-publish-images-manifests.yaml
+++ b/tekton/build-publish-images-manifests.yaml
@@ -95,6 +95,27 @@ spec:
       done
       cp ${DOCKER_CONFIG} /workspace/${KUBE_DISTRO}-docker-config.json
 
+  - name: rewrite-devel-in-source
+    image: ghcr.io/tektoncd/plumbing/ko-gcloud:v20240920-6c2a999d36@sha256:1756ca55a09b360028695792e638a7cc366292d7aef44c926a8cb765085664c8
+    script: |
+      #!/usr/bin/env sh
+      set -ex
+
+      # Rewrite VERSION env var value in operator.yaml (not in ko/kustomize output, sed on release.yaml won't catch it)
+      # Uses awk for multi-line: match '- name: VERSION', then rewrite the following 'value:' line
+      # (avoids reliance on GNU sed which is not available in this image)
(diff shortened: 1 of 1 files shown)


tekton/build-publish-images-manifests.yaml
```diff
@@ -95,6 +95,27 @@ spec:
...
 
+  - name: rewrite-devel-in-source
+    image: ghcr.io/tektoncd/plumbing/ko-gcloud:v20240920-6c2a999d36@sha256:1756ca55a09b360028695792e638a7cc366292d7aef44c926a8cb765085664c8
+    script: |
+      #!/usr/bin/env sh
+      set -ex
+
+      # Rewrite VERSION env var value in operator.yaml (not in ko/kustomize output, sed on release.yaml won't catch it)
+      # Uses awk for multi-line: match '- name: VERSION', then rewrite the following 'value:' line
+      # (avoids reliance on GNU sed which is not available in this image)
```
jkhelil, line 106: may be this comment is not right, sed is there and you are using it just after

anithapriyanatarajan: the sed version that is there in the image doe snot handle multiple line replacement. Hence used awk only for that file

PR #3440, opened 2026-05-19
docs(install): document that default namespace is unsupported for pipelines on OpenShift (#3427)

### LINKED ISSUE #3427: Document that the `default` namespace is not supported for running pipelines on OpenShift
On OpenShift, the `default` namespace is classified as a "highly privileged" system namespace. Pod Security Admission (PSA) label synchronization is **permanently disabled** on it by the platform.

This means that even though the operator correctly creates the `pipeline` ServiceAccount and RBAC bindings in the `default` namespace, PipelineRuns will fail with `permissionDenied` errors because PSA enforces the `restricted` profile and the SCC-to-PSA label sync never runs there.

Custom/user-created namespaces work fine because the Cluster Policy Controller automatically syncs SCC privileges into PSA labels.

OpenShift docs explicitly state: *"Do not run workloads in or share access to default projects."*

**Request**: Add documentation (e.g. in the OpenShift platform docs or a troubleshooting guide) noting that the `default` namespace is not supported for running pipelines, and users should use dedicated namespaces.

Related downstream Jira: https://redhat.atlassian.net/browse/SRVKP-12017

mvanhorn:
Closes #3427.

Adds a "Platform notes" section to `docs/install.md` explaining that the `default` namespace on OpenShift is unsupported for running pipelines, and pointing installers to dedicated namespaces. The operator silently creates the `pipeline` ServiceAccount and RBAC in `default`, but PipelineRuns then fail with `permissionDenied` because OpenShift permanently disables PSA label synchronization on `default` and PSA enforces the `restricted` profile.

This matches the OpenShift documentation guidance ("Do not run workloads in or share access to default projects") and the report in the linked downstream Jira (SRVKP-12017).

### Change Type

- [x] Documentation

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -30,3 +30,13 @@ To configure images from a custom registry, follow the [Air Gap Configuration](.
     ```
     $ kubectl apply -f https://raw.githubusercontent.com/tektoncd/operator/main/config/crs/kubernetes/config/all/operator_v1alpha1_config_cr.yaml
     ```
+
+## Platform notes
+
+### OpenShift: do not run pipelines in the `default` namespace
+
+On OpenShift, the `default` namespace is classified as a "highly privileged" system namespace. Pod Security Admission (PSA) label synchronization is permanently disabled there by the platform, so even though the operator correctly creates the `pipeline` ServiceAccount and RBAC bindings in `default`, PipelineRuns submitted to that namespace fail with `permissionDenied` errors: PSA enforces the `restricted` profile and the SCC-to-PSA label sync never runs.
+
+User-created namespaces are not affected because the Cluster Policy Controller automatically syncs SCC privileges into PSA labels. The OpenShift documentation has the same guidance ([Do not run workloads in or share access to default projects](https://docs.openshift.com/container-platform/latest/welcome/index.html#about-namespaces)).
+
+Run pipelines in a dedicated namespace instead of `default` on OpenShift. See [tektoncd/operator#3427](https://github.com/tektoncd/operator/issues/3427) for the original report.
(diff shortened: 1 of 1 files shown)


mathur07 (commented):
Minor concern with placement - the new section appears at the very end of the install doc, after custom registry configuration. Users might install, try to run pipelines by default, and only discover this limitation after reading the whole doc. Worth considering moving it higher, maybe right after the basic installation steps, so it's visible before users start experimenting.

Not a blocker, just means some folks might miss it until they hit the error.

mathur07:
Nevermind the doc seems to be not that big either. :)

docs/install.md
```diff
@@ -30,3 +30,13 @@ To configure images from a custom registry, follow the [Air Gap Configuration](.
...
     $ kubectl apply -f https://raw.githubusercontent.com/tektoncd/operator/main/config/crs/kubernetes/config/all/operator_v1alpha1_config_cr.yaml
     ```
+
+## Platform notes
+
+### OpenShift: do not run pipelines in the `default` namespace
+
+On OpenShift, the `default` namespace is classified as a "highly privileged" system namespace. Pod Security Admission (PSA) label synchronization is permanently disabled there by the platform, so even though the operator correctly creates the `pipeline` ServiceAccount and RBAC bindings in `default`, PipelineRuns submitted to that namespace fail with `permissionDenied` errors: PSA enforces the `restricted` profile and the SCC-to-PSA label sync never runs.
+
+User-created namespaces are not affected because the Cluster Policy Controller automatically syncs SCC privileges into PSA labels. The OpenShift documentation has the same guidance ([Do not run workloads in or share access to default projects](https://docs.openshift.com/container-platform/latest/welcome/index.html#about-namespaces)).
```
anithapriyanatarajan: The hyperlink to docs is not redirecting the namespace page. Please check or add link to the blog - https://www.redhat.com/en/blog/openshift-runtime-security-best-practices

anithapriyanatarajan (changes requested):

anithapriyanatarajan:
@mvanhorn Thank you for the PR. Minor NIT to verify the doc link.

mvanhorn:
Fixed in c4471df9 -- swapped to https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/building_applications/projects, which contains the exact "Do not run workloads in or share access to default projects" text matching the link label and explains the highly-privileged classification of the default project. Stayed on the canonical product docs (same source as the original link) instead of the blog so the citation matches the surrounding doc style. Thanks for the catch.

mvanhorn:
Thanks to the maintainers for merging the install note about the default namespace being unsupported for pipelines on OpenShift.

PR #3473, opened 2026-06-04
feat(openshift): add mTLS for Prometheus metrics

jkhelil:
Secure component metrics endpoints with mutual TLS on OpenShift. Each reconciler syncs the Prometheus client CA bundle from kube-system/extension-apiserver-authentication into the component namespace as a metrics-client-ca ConfigMap.

OpenShift's serving-cert controller is triggered via annotation on each metrics Service to provision a per-component TLS Secret. Two new manifest transformers (InjectMetricsServingCert, ApplyMetricsTLS) wire the Secret and ConfigMap as volumes and inject METRICS_PROMETHEUS_TLS_* env vars so that the knative/pkg prometheus.Server enables mTLS with require client auth.

ServiceMonitor resources are updated with scheme: https, scrapeClass tls-client-certificate-auth, and tlsConfig.serverName. The existing UpdateServiceMonitorTargetNamespace transformer is extended to also patch the namespace segment inside serverName at runtime.

Relates-To: SRVKP-8172

Assisted-by: Claude Sonnet 4.6 (via Cursor)

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Tekton Operator now enables mTLS between Prometheus and all Tekton
component metrics endpoints on OpenShift. Each component namespace
receives a synced client-CA ConfigMap and a serving-cert Secret so
that the knative/pkg prometheus.Server can enforce mutual TLS on its
/metrics endpoint.
```

### DIFF
--- a/cmd/openshift/operator/kodata/openshift-monitoring/00-monitoring.yaml
+++ b/cmd/openshift/operator/kodata/openshift-monitoring/00-monitoring.yaml
@@ -58,8 +58,14 @@ metadata:
 spec:
   endpoints:
     - interval: 10s
-      port: http-metrics
+      port: https-metrics
       honorLabels: true
+      scheme: https
+      tlsConfig:
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+        certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+        keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
+        serverName: tekton-pipelines-controller.openshift-pipelines.svc
   jobLabel: app
   namespaceSelector:
     matchNames:
--- a/pkg/reconciler/openshift/common/metricstls.go
+++ b/pkg/reconciler/openshift/common/metricstls.go
@@ -1,0 +1,305 @@
+/*
+Copyright 2026 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"fmt"
+
+	mf "github.com/manifestival/manifestival"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
+
+	rcommon "github.com/tektoncd/operator/pkg/reconciler/common"
+)
+
+const (
+	// MetricsHTTPPort is the upstream port name for plain-HTTP metrics.
+	MetricsHTTPPort = "http-metrics"
+	// MetricsHTTPSPort is the renamed port name after mTLS is enabled.
+	MetricsHTTPSPort = "https-metrics"
+
+	// metricsServingCertAnnotation is the OpenShift annotation that triggers
+	// automatic provisioning of a service serving certificate.
+	metricsServingCertAnnotation = "service.beta.openshift.io/serving-cert-secret-name"
+
+	// metricsServingCertVolume and metricsClientCAVolume are the Pod volume names.
+	metricsServingCertVolume = "metrics-serving-cert"
+	metricsClientCAVolume    = "metrics-client-ca"
+
+	// MetricsTLSMountPath is where the serving-cert Secret is mounted.
+	MetricsTLSMountPath = "/etc/metrics-tls"
+	// MetricsClientCAMountPath is where the client-CA ConfigMap is mounted.
+	MetricsClientCAMountPath = "/etc/metrics-client-ca"
+
+	// Env vars read by knative/pkg prometheus.Server to configure mTLS.
+	envMetricsTLSCert       = "METRICS_PROMETHEUS_TLS_CERT"
+	envMetricsTLSKey        = "METRICS_PROMETHEUS_TLS_KEY"
+	envMetricsTLSClientCA   = "METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE"
+	envMetricsTLSClientAuth = "METRICS_PROMETHEUS_TLS_CLIENT_AUTH"
+
+	// Paths inside the Prometheus pod where CMO mounts the scraping credentials.
+	// These match the tls-client-certificate-auth scrapeClass definition.
+	promCertFile = "/etc/prometheus/secrets/metrics-client-certs/tls.crt"
+	promKeyFile  = "/etc/prometheus/secrets/metrics-client-certs/tls.key"
+	promCAFile   = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
+)
+
+// MetricsServingCertSecretName returns the conventional TLS Secret name for
+// the given Service (i.e. "<serviceName>-metrics-tls").
+func MetricsServingCertSecretName(serviceName string) string {
+	return serviceName + "-metrics-tls"
+}
+
+// AnnotateMetricsServingCert adds the OpenShift serving-cert annotation to the
+// named Service so that OpenShift automatically provisions a TLS Secret named
+// "<serviceName>-metrics-tls". Call this once per Service that needs its own
+// metrics serving certificate. If another transformer already sets this
+// annotation on the Service, do not call this — use the existing secret instead.
+func AnnotateMetricsServingCert(serviceName string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
+		if u.GetKind() != "Service" || u.GetName() != serviceName {
+			return nil
+		}
+		annotations := u.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		annotations[metricsServingCertAnnotation] = MetricsServingCertSecretName(serviceName)
+		u.SetAnnotations(annotations)
+		return nil
+	}
+}
+
+// RenameServicePort renames fromPort to toPort on the named Service.
+// Use this to align the Service port name with what the ServiceMonitor expects
+// (e.g. "http-metrics" → "https-metrics", "prometheus" → "https-prometheus").
+func RenameServicePort(serviceName, fromPort, toPort string) mf.Transformer {
+	return func(u *unstructured.Unstructured) error {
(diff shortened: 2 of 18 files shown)


pkg/reconciler/openshift/tektonresult/extension.go
```diff
@@ -108,6 +112,15 @@ func (oe *openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.T
 		injectLokiStackTLSCACert(instance.Spec.LokiStackProperties),
 		injectResultsAPIServiceCACert(instance.Spec.ResultsAPIProperties),
 		injectPostgresUpgradeSupport(),
+		// mTLS for Prometheus scraping.
+		// tekton-results-watcher is a StatefulSet that exposes a "metrics" port.
+		// tekton-results-api-service is a Deployment that exposes a "prometheus" port.
+		occommon.InjectMetricsServingCertWithPort(tektonResultWatcherName, "metrics"),
+		occommon.ApplyMetricsTLS("StatefulSet", tektonResultWatcherName,
+			occommon.MetricsServingCertSecretName(tektonResultWatcherName)),
+		occommon.InjectMetricsServingCertWithPort(serviceAPI, "prometheus"),
```
infernus01: This might create annotation conflict with [injectResultsAPISeerviceCACert](https://github.com/tektoncd/operator/pull/3473/changes#diff-a08d70d4f7f1a1932aafcdf80de931bc4218ee3838bc09fb6db75139805ba3c5R113) as this already sets [service.beta.openshift.io/serving-cert-secret-name](https://github.com/tektoncd/operator/pull/3473/changes#diff-a08d70d4f7f1a1932aafcdf80de931bc4218ee3838bc09fb6db75139805ba3c5R243) to "tekton-results-tls" on tekton-results-api-service. This transformer overwrites that same annotation with "tekton-results-api-service-metrics-tls" (via [MetricsServingCertSecretName](https://github.com/tektoncd/operator/pull/3473/changes#diff-0efc7d6f2deab46ddecca3fb520c396bb586308ff5c3609005ff85a406898ff4R66)).

cmd/openshift/operator/kodata/openshift-monitoring/00-monitoring.yaml
```diff
@@ -58,8 +58,14 @@ metadata:
 spec:
   endpoints:
     - interval: 10s
-      port: http-metrics
+      port: https-metrics
       honorLabels: true
+      scheme: https
+      tlsConfig:
```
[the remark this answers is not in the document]
jkhelil: scrapeClass is not working on ocp 4.19, we inject directly the client cert

infernus01 (commented):
/lgtm

pkg/reconciler/openshift/common/metricstls.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// Add the client-CA ConfigMap volume.
+	clientCAVol := corev1.Volume{
+		Name: metricsClientCAVolume,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: MetricsClientCAConfigMap,
+				},
+			},
```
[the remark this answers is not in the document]
jkhelil: The ordering guarantee is the fix. EnsureMetricsClientCA runs in TektonConfig.PreReconcile, which executes before any child component (TektonPipeline, TektonResult, etc.) is reconciled and before their pods are created. The ConfigMap is always present by the time pods are scheduled.

If the ConfigMap is deleted after install, TektonConfig's periodic reconciliation recreates it. In the small window between deletion and recreation, a pod restart will block — which is intentional. A pod without the client-CA bundle cannot perform mTLS client verification at all, so failing loudly (pod unschedulable) is preferable to starting silently with broken mTLS. This is consistent with how other mandatory ConfigMaps (e.g. CA bundles) are handled throughout the operator.

pkg/reconciler/openshift/common/transformer.go
```diff
@@ -176,15 +178,67 @@ func UpdateServiceMonitorTargetNamespace(targetNamespace string) mf.Transformer
...
 		return nil
 	}
 }
 
+// replaceServerNameNamespace replaces the namespace segment in a Kubernetes
+// in-cluster DNS name of the form "<service>.<namespace>.svc[…]" with
+// targetNamespace. Returns the original string unchanged if it does not match.
+func replaceServerNameNamespace(serverName, targetNamespace string) string {
+	// Preserve any suffix after ".svc" (e.g. ".cluster.local").
+	const svsSuffix = ".svc"
```
enarha: typo? `svsSuffix` -> `svcSuffix`

pkg/reconciler/openshift/common/transformer.go
```diff
@@ -176,15 +178,67 @@ func UpdateServiceMonitorTargetNamespace(targetNamespace string) mf.Transformer
...
+		}
+
 		return nil
 	}
 }
 
+// replaceServerNameNamespace replaces the namespace segment in a Kubernetes
+// in-cluster DNS name of the form "<service>.<namespace>.svc[…]" with
+// targetNamespace. Returns the original string unchanged if it does not match.
+func replaceServerNameNamespace(serverName, targetNamespace string) string {
```
enarha, line 226: replaceServerNameNamespace is easy to test and I consider it worth adding test.

pkg/reconciler/openshift/tektonchain/extension.go
```diff
@@ -25,21 +25,28 @@ import (
 	operatorclient "github.com/tektoncd/operator/pkg/client/injection/client"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	occommon "github.com/tektoncd/operator/pkg/reconciler/openshift/common"
+	"k8s.io/client-go/kubernetes"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
 )
 
 const (
 	tektonChainsControllerName = "tekton-chains-controller"
+	// tektonChainsMetricsService is the Service that exposes the http-metrics
```
enarha: `exposes the http-metrics` is a bit misleading because after the transformation this will be https. I suggest you entirely drop the http/https reference e.g. "tektonChainsMetricsService is the metrics Service for chains."

pkg/reconciler/openshift/openshiftpipelinesascode/extension.go
```diff
@@ -106,16 +123,14 @@ func (oe *openshiftExtension) Transformers(comp v1alpha1.TektonComponent) []mf.T
...
-			// injecting with no prefix is harmless and keeps them consistent for future use.
 			occommon.InjectTLSEnvVars(oe.resolvedTLSConfig, "Deployment", pacControllerDeployment, []string{pacControllerContainerName}, ""),
 		)
 	}
 
 	return trns
 }
 
-func (oe *openshiftExtension) PreReconcile(ctx context.Context, _ v1alpha1.TektonComponent) error {
+func (oe *openshiftExtension) PreReconcile(ctx context.Context, comp v1alpha1.TektonComponent) error {
```
enarha: `comp` is not really used in that function. Why change `_`?

enarha (commented):
metricsca.go, metricstls.go were added, but the PR does not include tests for these. Here are few things we can tests (asked Claude to generate):

## Test recommendations for `metricsca.go`, `metricstls.go`, and `UpdateServiceMonitorTargetNamespace`

### `metricsca.go` — `EnsureMetricsClientCA`

Unit tests using `fake.NewSimpleClientset()`:

| Case | What it verifies |
|------|-----------------|
| Source ConfigMap missing | Returns error, no target ConfigMap created |
| Source ConfigMap exists but key `client-ca-file` absent | Returns error with descriptive message |
| Target ConfigMap doesn't exist | Creates it with correct name, namespace, labels, data |
| Target ConfigMap exists, same CA bundle | No update call (verify via fake client action list) |
| Target ConfigMap exists, different CA bundle | Updates with new bundle, preserves other fields |
| Target ConfigMap exists, `Data` is nil | Updates without panic |

These are straightforward fake-clientset tests. The function has clear branch points that map 1:1 to test cas
es. Verify side effects through the fake client's `Actions()` — e.g. confirm no `Update` action fires in the 
no-op case.

### `metricstls.go` — transformer unit tests

Each transformer takes an `*unstructured.Unstructured` and mutates it. The existing `cabundle_test.go` and `t
ransformer_test.go` already establish the pattern: build an unstructured object, run the transformer, assert 
on the result.

**`AnnotateMetricsServingCert`**:
- Matching Service with no annotations → annotation added
- Matching Service with existing annotations → annotation added, others preserved
- Non-matching Service name → unchanged
- Non-Service resource (e.g. Deployment) → unchanged

**`RenameServicePort`**:
- Port matching `fromPort` → renamed to `toPort`
- No port matches `fromPort` → unchanged (no error)
- Multiple ports, only one matches → only that one renamed
- Non-matching Service → unchanged

**`UpdateServiceMonitorForMetricsMTLS`**:
- Matching ServiceMonitor → port renamed, scheme set to `https`, tlsConfig injected with correct serverName
- Non-matching name → unchanged
- No endpoints in spec → unchanged (no error)
- Multiple endpoints → all get scheme + tlsConfig

**`ApplyMetricsTLS`**:
- Deployment: volumes, volume mounts, and env vars injected into all containers
- StatefulSet: same
- Non-matching kind or name → unchanged
- **Idempotency**: run twice, result is identical to running once (this is important — `AddOrReplaceInList` should handle it, but verify)
- Multi-container pod → all containers get mounts and env vars

**`replaceServerNameNamespace`** — pure function, table-driven:

| Input | Expected output |
|-------|-----------------|
| `"svc.openshift-pipelines.svc"` | `"svc.new-ns.svc"` |
| `"svc.openshift-pipelines.svc.cluster.local"` | `"svc.new-ns.svc.cluster.local"` |
| `"no-dot-svc"` | `"no-dot-svc"` (unchanged) |
| `"svc.ns.nomatch"` | `"svc.ns.nomatch"` (unchanged) |
| `""` | `""` (unchanged) |

### `UpdateServiceMonitorTargetNamespace` (refactored in `transformer.go`)

The refactored function now does two things: replaces `matchNames[0]` and patches `tlsConfig.serverName`. Test them together since they're one transformer:

| Case | What it verifies |
|------|-----------------|
| ServiceMonitor with matchNames + tlsConfig.serverName | Both namespace segments replaced |
| ServiceMonitor with matchNames, no tlsConfig | matchNames updated, no error on missing tlsConfig |
| ServiceMonitor with no matchNames | No panic (old code would panic on this) |
| ServiceMonitor with multiple endpoints, only some have tlsConfig | Only endpoints with tlsConfig get serverName patched |
| Non-ServiceMonitor resource | Unchanged |

The regression value here is high — the old code had bare type assertions that would panic on unexpected structure. The new tests should confirm that malformed inputs produce errors, not panics.

### Priority

Highest-value tests to add first:

1. `replaceServerNameNamespace` — pure function, quick to write, catches real edge cases
2. `ApplyMetricsTLS` idempotency — the most complex transformer, most likely to regress
3. `EnsureMetricsClientCA` create-vs-update — branching logic with real operational consequences

enarha (commented):
/approve

infernus01 (commented):
/lgtm

PR #3507, opened 2026-06-17
Add denylist for system namespaces in targetNamespace validation

### LINKED ISSUE #3248: Add denylist for system namespaces in `targetNamespace` validation
## Description

The `targetNamespace` field on TektonConfig (and other Tekton component CRDs via `CommonSpec`) currently lacks validation against well-known system namespaces. While existing controls (read-only shipped RBAC, singleton enforcement, immutability after creation, auto-creation defaults) make misuse unlikely, adding a denylist is a sensible defense-in-depth measure.

## Current Behavior

In `pkg/apis/operator/v1alpha1/common_validation.go`, the only validations on `targetNamespace` are:
- Must be non-empty
- On OpenShift only: `openshift-operators` is blocked

On Kubernetes, any non-empty namespace is accepted.

## Proposed Change

Add a denylist in `CommonSpec.validate()` that rejects the following namespaces on **both** platforms:
- `kube-system`
- `kube-public`
- `kube-node-lease`
- `default`

On OpenShift, additionally keep the existing `openshift-operators` check and consider also blocking `openshift-*` prefixed namespaces (aligning with the existing `NamespaceIgnorePattern` regex used in the RBAC reconciler).

This should apply to all components using `CommonSpec` (TektonConfig, TektonPipeline, TektonTrigger, TektonChain, TektonResult, TektonHub, etc.).

mvanhorn:
# Changes

`CommonSpec.validate()` previously accepted any non-empty `targetNamespace` on Kubernetes, so a component could be told to install into a well-known system namespace such as `kube-system`, `kube-public`, `kube-node-lease`, or `default`. On OpenShift only `openshift-operators` was blocked.

This adds a defense-in-depth denylist that rejects those four reserved system namespaces on both Kubernetes and OpenShift, while preserving the existing OpenShift-only `openshift-operators` rejection. Because `validate()` is the shared method called by every component validator, the denylist applies across components with no caller changes.

While wiring this up I found that `TektonResult` bypassed the shared validation entirely: `TektonResult.Validate` calls `TektonResultSpec.validate`, which never invoked the embedded `CommonSpec.validate`. This change also calls the common validation from `TektonResultSpec.validate` so the denylist (and the pre-existing empty-field check) is enforced consistently for `TektonResult` too.

Tests cover the happy path, each denied namespace on both platforms, the preserved `openshift-operators` rejection, the empty-namespace case, and the `TektonResult` path.

Fixes #3248

# Submitter Checklist

review them:

# Release Notes

```release-note
Reject well-known system namespaces (kube-system, kube-public, kube-node-lease, default) in `targetNamespace` validation across Tekton components.
```


pkg/apis/operator/v1alpha1/common_validation.go
```diff
@@ -22,16 +22,38 @@ import (
...
+var reservedSystemNamespaces = []string{
+	"kube-system",
+	"kube-public",
+	"kube-node-lease",
+	"default",
+}
+
 func (ta *CommonSpec) validate(path string) *apis.FieldError {
 	var errs *apis.FieldError
+	targetNamespace := ta.GetTargetNamespace()
```
jkhelil, line 34: this is cleaner than calling GetTargetNamespace() multile time

pkg/apis/operator/v1alpha1/tektonresult_validation_test.go
```diff
@@ -37,6 +37,23 @@ func TestTektonResult_Validate(t *testing.T) {
...
 
+func TestTektonResult_ValidateTargetNamespaceDenylist(t *testing.T) {
+	tr := &TektonResult{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "result",
+			Namespace: "namespace",
+		},
+		Spec: TektonResultSpec{
+			CommonSpec: CommonSpec{
+				TargetNamespace: "kube-system",
```
jkhelil, line 48: this skips validation  kube-public, kube-node-lease, default, and the OpenShift openshift-operators case, without any failing test at the TektonResult level.

jkhelil:
@mvanhorn thank you for you contribution, would you add the validation to pkg/apis/operator/v1alpha1/tektonscheduler_validation.go:27–41
TektonScheduler embeds CommonSpec but its Validate() only calls ts.Spec.MultiClusterConfig.validate() so it never calls CommonSpec.validate().

pkg/apis/operator/v1alpha1/common_validation.go
```diff
@@ -22,16 +22,38 @@ import (
...
+			if targetNamespace == "openshift-operators" {
+				errs = errs.Also(apis.ErrInvalidValue(targetNamespace, targetNamespacePath, "'openshift-operators' namespace is not allowed"))
+			}
 		}
 	}
 	return errs
 }
+
+func isReservedSystemNamespace(namespace string) bool {
+	for _, reservedNamespace := range reservedSystemNamespaces {
```
l-qing, line 53: Minor: this loop can be simplified with slices.Contains (Go 1.21+):
```
return slices.Contains(reservedSystemNamespaces, namespace)
```

mvanhorn:
Done - `TektonScheduler.Validate()` now calls `ts.Spec.CommonSpec.validate("spec")` the same way the other components do, so the common spec checks (targetNamespace presence + reserved-namespace rejection) actually run. Updated the existing scheduler validation fixtures to set a valid targetNamespace and added a test asserting a reserved system namespace is now rejected. go build/vet/test on the package are green.

anithapriyanatarajan:
@mvanhorn. Please incorporate the comment - https://github.com/tektoncd/operator/pull/3507#discussion_r3575621561 and if possible request to include the validation for `SyncerService` and `TektonMulticlusterProxyAAE`. Else submit an issue so that it could be picked up later. Thank you.

mvanhorn:
Done, and I went ahead and covered both.

- `isReservedSystemNamespace` now uses `slices.Contains` per the inline comment.
- `SyncerService` and `TektonMulticlusterProxyAAE` both validate targetNamespace now.

One thing worth flagging on those two: I pulled the denylist out of `CommonSpec.validate` into `validateTargetNamespaceDenylist` rather than calling `validate` directly. `validate` also errors on an empty targetNamespace, and neither of those two requires it today, so wiring in the whole thing would have started rejecting resources that omit it. The split keeps the denylist without that behavior change.

Denylist tests are table-driven now and cover all four reserved namespaces plus the openshift-operators case. `go test ./pkg/apis/operator/v1alpha1/...` passes. I couldn't run `make lint` locally (golangci-lint download failed a checksum here), so CI is the check on that.

anithapriyanatarajan:
/approve
@mvanhorn - could you please squash the commits

mvanhorn:
@anithapriyanatarajan squashed to a single commit, and the tree is byte-identical to what you approved.

The combined message keeps the reasoning for all three parts: the denylist itself, plus the two components that never reached `CommonSpec.validate` and so bypassed it (TektonResult via `TektonResultSpec.validate`, and TektonScheduler which only called `MultiClusterConfig.validate`).

`go test ./pkg/apis/operator/v1alpha1/...` passes.

PR #3508, opened 2026-06-18
feat(helm): publish chart to OCI registry on release

### LINKED ISSUE #3493: [chart] invalid Chart version - Flux CD validation error
Greetings!
I think this was mentioned in a previous issue [#3483 ](https://github.com/tektoncd/operator/issues/3483)
Trying to install the latest version of the tekton operator (v0.80.0) via FluxCD but getting a validation error because of the chart Version. It isn't valid semver

And can confirm that the version is still wrong
[v0.80.0](https://github.com/tektoncd/operator/blob/v0.80.0/charts/tekton-operator/Chart.yaml)

# Expected Behavior

Helm Release installed successfully

# Actual Behavior

`HelmChart 'flux-system/tekton-operator-tekton-operator' is not ready: invalid chart reference: validation: chart.metadata.version "devel" is invalid`

Currently sticking to 0.79.1 but would like this to be fixed in the near future if possible

Thanks in advance

jkhelil:
## Summary

- Add OCI publishing step to the `helm-release` workflow so the Tekton Operator Helm chart is available at `oci://ghcr.io/tektoncd/operator/charts/tekton-operator` on every release, in addition to the existing GitHub Pages Helm repository.
- Adds `packages: write` permission to the job.
- Uses `docker/login-action` (SHA-pinned, consistent with repo security posture) for `ghcr.io` authentication.

## Motivation

The chart is currently only distributed via `git+https://github.com/tektoncd/operator@charts?ref=main`, which Flux CD's `HelmRepository` cannot consume (it requires `https://` index or `oci://`). Users forced to use a `GitRepository` source pointing to the git tag get `version: "devel"` from the raw `Chart.yaml`, which Flux rejects as invalid semver.

Publishing to OCI allows Flux users to use the standard pattern:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: HelmRepository
metadata:
  name: tekton-operator
  namespace: flux-system
spec:
  type: oci
  url: oci://ghcr.io/tektoncd/operator/charts
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
spec:
  chart:
    spec:
      chart: tekton-operator
      version: "0.80.0"
```

Closes #3493

## Test plan

- [ ] Merge and cherry-pick onto `release-v0.80.x`
- [ ] Delete and recreate `v0.80.0` tag to trigger the workflow
- [ ] Verify package appears at `https://github.com/orgs/tektoncd/packages/container/package/operator%2Fcharts%2Ftekton-operator`
- [ ] Verify `helm pull oci://ghcr.io/tektoncd/operator/charts/tekton-operator --version 0.80.0` succeeds
- [ ] Verify Flux CD `HelmRepository` with `type: oci` installs successfully

### DIFF
--- a/.github/workflows/helm-release.yaml
+++ b/.github/workflows/helm-release.yaml
@@ -8,6 +8,7 @@ jobs:
   release:
     permissions:
       contents: write
+      packages: write
     runs-on: ubuntu-latest
     if: github.repository_owner == 'tektoncd'  # do not run this elsewhere
     steps:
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
Make sense ! We'll want to make sure the oci repository is public after the first build, but I think it makes a lot of sense !

PR #3530, opened 2026-06-23
ci(dependabot): remove unsupported cooldown fields from github-actions

jkhelil:
## Summary

The \`semver-major-days\`, \`semver-minor-days\` and \`semver-patch-days\` cooldown fields are not supported for the \`github-actions\` ecosystem, causing Dependabot validation errors on every entry:

\`\`\`
The property '#/updates/0/cooldown/semver-major-days' is not supported for the package ecosystem 'github-actions'.
\`\`\`

These fields are semver-specific and only valid for package managers like \`gomod\`. GitHub Actions uses git tags/SHAs, not semver.

Fix by keeping only \`default-days\` in the \`github-actions\` ecosystem cooldown config and regenerating \`dependabot.yml\`.

## Test plan
- [ ] Verify no Dependabot configuration errors after merge

/kind misc
/release-note-none

### DIFF
--- a/.github/dependabot.config.yml
+++ b/.github/dependabot.config.yml
@@ -39,9 +39,6 @@ ecosystems:
       - release-note-none
     cooldown:
       default-days: 7
-      semver-major-days: 30
-      semver-minor-days: 7
-      semver-patch-days: 3
   - package-ecosystem: gomod
     directory: /
     schedule:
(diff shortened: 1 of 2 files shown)


vdemeester (approved):
/lgtm

@waveywaves we need to check on pipeline

PR #3581, opened 2026-06-26
feat(namespacesync): add NamespaceSyncController for per-namespace resource management

jkhelil:
# Changes

Introduces a new `NamespaceSyncController` that replaces the
scan-based batch processing in `rbac.go` with an event-driven,
per-namespace reconciler. Addresses RFE-7814 (flexible service
account binding for Quay Robot credentials).

## What this does

- Adds `NamespaceSyncController` — a watch-based controller that
  reacts to Namespace, ServiceAccount, Secret, and TektonConfig
  events to ensure per-namespace Tekton resources are present and
  up to date in every user namespace
- Creates `pipeline` SA, SCC RoleBinding, edit RoleBinding, and
  CA bundle ConfigMaps in each namespace
- Manages `openshift-pipelines-clusterinterceptors` CRB
  incrementally (add/remove only the affected namespace's subject)
- Supports Quay Robot secret binding via `spec.platforms.openshift
  .namespaceSync.secretBindings` (the core RFE-7814 use case)
- Adds typed `spec.platforms.openshift.namespaceSync` API field to
  TektonConfig, replacing legacy `spec.params` entries
- Removes ~1300 lines of dead per-namespace batch code from `rbac.go`
- Registers the controller in the lifecycle container and adds
  `LeaderAwareFuncs` embedding (required by Knative sharedmain)
- Adds `namespaceSync` schema to the TektonConfig CRD

## Status

Work in progress — more e2e tests needed before merge:
- [ ] E2E test: Quay Robot secret binding end-to-end
- [ ] E2E test: namespace deletion removes CRB subject
- [ ] E2E test: self-healing when pipeline SA is deleted manually
- [ ] `make test lint` full pass

## Verified on cluster

All per-namespace resources (pipeline SA, SCC RoleBinding, edit
RoleBinding, CA bundles, CRB subject) created correctly in new
namespaces within seconds. Established namespaces unaffected.

# Submitter Checklist

# Release Notes

```release-note
Add NamespaceSyncController: event-driven per-namespace provisioning
of pipeline SA, SCC/edit RoleBindings, CA bundles, and Quay Robot
secret binding on OpenShift. Replaces legacy batch scan in rbac.go.
New typed API: spec.platforms.openshift.namespaceSync in TektonConfig.

The spec.params entries createRbacResource, createCABundleConfigMaps,
and legacyPipelineRbac are now deprecated in favor of the typed
namespaceSync fields (createPipelineSA, createCABundles,
createEditRoleBinding). They are still honored for backward
compatibility and are automatically migrated to the typed fields
(and removed from spec.params) during the first reconcile after
upgrade -- no manual action is required, but new configurations
should use namespaceSync directly.
```


anithapriyanatarajan:
@jkhelil - If you are still planning to pursue this PR, please rebase.

jkhelil:
@waveywaves @pratap0007 Would you mind review this please ?

pkg/reconciler/openshift/namespacesync/reconciler.go
```diff
@@ -0,0 +1,851 @@
...
+	tc, err := r.tektonConfigLister.Get(v1alpha1.ConfigResourceName)
+	if errors.IsNotFound(err) {
+		logger.Debug("TektonConfig not found, skipping namespace sync")
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	cfg := tc.Spec.Platforms.OpenShift.NamespaceSync
```
waveywaves, line 110: Older TektonConfig objects will not have this field, so namespace setup may stop after upgrade. Could we handle existing configurations and add an upgrade test?

pkg/reconciler/openshift/namespacesync/reconciler.go
```diff
@@ -0,0 +1,851 @@
...
+// namespaceMatchesSelector returns true when the namespace should be synced
+// according to cfg.NamespaceSelector. When no selector is configured every
+// non-ignored namespace matches (opt-in all by default). Setting the selector
+// to an empty matchLabels ({}) matches nothing, effectively disabling sync for
+// all namespaces without touching the individual feature flags.
+func namespaceMatchesSelector(ns *corev1.Namespace, cfg *v1alpha1.NamespaceSyncConfig) bool {
+	if cfg.NamespaceSelector == nil {
+		return true
+	}
+	sel, err := metav1.LabelSelectorAsSelector(cfg.NamespaceSelector)
```
waveywaves, line 981: `{}` matches every namespace, not none, and invalid selectors also match every namespace. Could we fix and test both cases?

pkg/reconciler/openshift/namespacesync/controller.go
```diff
@@ -0,0 +1,261 @@
...
+		tektonConfigLister: tcInf.Lister(),
+	}
+
+	impl := controller.NewContext(ctx, rec, controller.ControllerOptions{
+		WorkQueueName: "NamespaceSyncController",
+		Logger:        logger,
+	})
+
+	// Namespace Add/Update → reconcile that namespace.
+	if _, err := nsInf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
waveywaves, line 79: Deleting a namespace leaves its `pipeline` entry in the cluster-wide permission binding. Could we remove it on deletion?

pkg/reconciler/openshift/namespacesync/controller.go
```diff
@@ -0,0 +1,261 @@
...
+	//   - A deleted named secret is unbound from the pipeline SA.
+	//
+	// We use the cluster-wide kube factory here (NOT the namespacedkube one)
+	// because we need to observe secrets in all user namespaces, not just the
+	// operator's own namespace.
+	//
+	// To avoid a thundering-herd on clusters that don't use secret bindings,
+	// we only enqueue when NamespaceSync has SecretBindings configured AND the
+	// secret name/labels match at least one binding rule.
+	if _, err := secretInf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
waveywaves, line 178: Secret label changes are not watched. Could we handle updates so matching Secrets are added and removed?

pkg/reconciler/openshift/namespacesync/reconciler.go
```diff
@@ -0,0 +1,851 @@
...
+	}
+
+	if cfg.CreateCABundles != nil && *cfg.CreateCABundles {
+		if err := r.ensureCABundles(ctx, ns); err != nil {
+			logger.Errorf("failed to ensure CA bundles in %s: %v", ns.Name, err)
+			return err
+		}
+	}
+
+	if len(cfg.SecretBindings) > 0 {
```
waveywaves: Changing or removing `secretBindings` leaves old Secret references. Could we remove references added by this controller?

pkg/reconciler/openshift/namespacesync/reconciler.go
```diff
@@ -0,0 +1,851 @@
...
+			logger.Infof("Binding secret %s/%s to pipeline SA", ns.Name, name)
+			changed = true
+		}
+	}
+
+	if !changed {
+		return nil
+	}
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
waveywaves: Each retry uses the same outdated ServiceAccount, so it cannot resolve a conflict. Could we fetch it again on each retry?

pkg/reconciler/openshift/namespacesync/controller.go
```diff
@@ -0,0 +1,261 @@
...
+)
+
+// NewController initialises the NamespaceSyncController and registers event
+// handlers for Namespace, ServiceAccount (pipeline only), Secret, and TektonConfig.
+func NewController(ctx context.Context, _ configmap.Watcher) *controller.Impl {
+	logger := logging.FromContext(ctx)
+
+	nsInf := nsinformer.Get(ctx)
+	saInf := sainformer.Get(ctx)
+	secretInf := secretinformer.Get(ctx)
```
waveywaves, line 51: This loads every Secret’s contents into operator memory. Could we limit the watch to the information needed?

pkg/reconciler/openshift/namespacesync/reconciler.go
```diff
@@ -0,0 +1,851 @@
...
+	}
+
+	// Verify the edit ClusterRole exists before creating the binding.
+	if _, err := r.kubeClient.RbacV1().ClusterRoles().Get(ctx, editClusterRole, metav1.GetOptions{}); err != nil {
+		return err
+	}
+
+	logger.Infof("Creating edit RoleBinding in namespace %s", ns.Name)
+	_, err = rbClient.Create(ctx, &rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
```
waveywaves: New RoleBindings have no owner, and incorrect existing bindings are not fixed. Could we preserve ownership and correct existing bindings?

pkg/reconciler/openshift/namespacesync/controller.go
jkhelil: uncessary data are striped now

pkg/reconciler/openshift/namespacesync/reconciler.go
jkhelil: fixed

pkg/reconciler/openshift/namespacesync/controller.go
jkhelil: fixed

jkhelil:
# NamespaceSyncController E2E Test Proof

**Result: 15/15 specs PASSED, 0 failed, 0 errors.**

- Suite: `NamespaceSyncController` (Ginkgo v2), `openshift-pipelines/release-tests-ginkgo`
- PR: https://github.com/openshift-pipelines/release-tests-ginkgo/pull/77 (closes [issue #76](https://github.com/openshift-pipelines/release-tests-ginkgo/issues/76))
- Branch: `agent/76-namespacesync-e2e` @ commit `7f5fc8c`
- Feature under test: `spec.platforms.openshift.namespaceSync` on `TektonConfig` — [tektoncd/operator#3581](https://github.com/tektoncd/operator/pull/3581) (RFE-7814)
- Run date: 2026-09-03, ~08:51–09:00 UTC+2 (06:51–07:00 UTC)
- Total run time: 535.5s (~8m56s) for the 15 focused specs

## Cluster under test

See `cluster-evidence.txt` for the raw command output. Summary:

| Field | Value |
|---|---|
| Cluster | `api.sq3p3-8mzcc-ss2.i5ut.p3.openshiftapps.com` |
| Access | `cluster-admin` |
| Operator image | `quay.io/jkhelil/operator-1d69a75f22dd094880847eac907fb2c1@sha256:bdc535f2...` |
| TektonConfig `status.version` | `devel` |
| TektonConfig `Ready` condition | `True` |

## Per-spec results

| # | Spec | Result | Duration |
|---|---|---|---|
| 1 | Baseline sync on namespace creation > creates the pipeline SA, CA bundles, edit RoleBinding and SCC RoleBinding | PASS | 39.6s |
| 2 | Per-feature toggles > createPipelineSA | PASS | 35.6s |
| 3 | Per-feature toggles > createCABundles | PASS | 31.4s |
| 4 | Per-feature toggles > createEditRoleBinding | PASS | 31.9s |
| 5 | Per-feature toggles > createSCCRoleBinding | PASS | 29.2s |
| 6 | Secret binding — named secret > binds on create, unbinds on delete, and stays unbound once the binding rule is removed | PASS | 50.1s |
| 7 | Secret binding — label selector > binds all matching secrets and unbinds only the one that stops matching | PASS | 49.5s |
| 8 | Self-healing > recreates manually deleted resources and re-adds a manually removed secret ref | PASS | 60.0s |
| 9 | Namespace deletion > removes the pipeline SA subject from the cluster-scoped ClusterInterceptors CRB | PASS | 31.5s |
| 10 | namespaceSelector scoping > syncs only namespaces matching matchLabels | PASS | 34.4s |
| 11 | namespaceSelector scoping > an explicit empty namespaceSelector ({}) opts out every namespace | PASS | 16.7s |
| 12 | System namespace exclusion > never touches namespaces matching the system-namespace ignore pattern | PASS | 32.1s |
| 13 | Legacy spec.params migration > maps createRbacResource/createCABundleConfigMaps/legacyPipelineRbac onto typed defaults | PASS | 34.2s |
| 14 | Legacy spec.params migration > persists the migration onto the stored CR via the pre-upgrade job, not just in-memory defaulting | PASS | 28.0s |
| 15 | Upgrade compatibility > keeps syncing with documented defaults when namespaceSync is entirely absent | PASS | 31.3s |

## Artifacts in this folder

- `nssync-full-run.log` — full raw console transcript of the run (ginkgo `-v`, every step, every `oc` call and its output).
- `nssync-junit.xml` — JUnit XML report (`tests=63 failures=0 errors=0 skipped=46`; the 63 includes ginkgo's `SynchronizedBeforeSuite`/`AfterSuite` container nodes alongside the 15 focused specs and 46 skipped specs from the rest of the suite that weren't in scope for this focus).
- `nssync-report.json` — Ginkgo's structured JSON report (machine-readable, one entry per spec with state/timing).
- `cluster-evidence.txt` — raw `oc` output proving the cluster identity, operator pod status, image digest, and `TektonConfig` readiness at test time.

## How this was run

```bash
PATH="/tmp/faux-bin:$PATH" go run github.com/onsi/ginkgo/v2/ginkgo -v \
  --focus "NamespaceSyncController" \
  --junit-report=nssync-junit.xml \
  --json-report=nssync-report.json \
  ./tests/operator/...
```

(`/tmp/faux-bin` contains a stub `opc` binary so the suite's `ValidateOperatorInstallStatus` prerequisite check passes; the OpenShift Pipelines CLI itself isn't required for anything these specs actually assert against.)

Note: an earlier run against this same cluster caught a real test-code bug (two legacy-params specs issued the `namespaceSync` reset and the params patch as separate API calls, racing with TektonConfig's mutating webhook). That was root-caused and fixed in commit `7f5fc8c` before this proof run — see the PR for details.

docs/TektonConfig.md
```diff
@@ -606,6 +606,132 @@ In the deployment the environment name will be converted as follows,
...
+| `config-service-cabundle` | `ConfigMap` | OpenShift service CA bundle |
+| `openshift-pipelines-clusterinterceptors` | `ClusterRoleBinding` subject | Lets the pipeline SA call ClusterInterceptors |
+
+#### Configuration fields
+
+```yaml
+spec:
+  platforms:
+    openshift:
+      namespaceSync:
```
anithapriyanatarajan, line 656: @jkhelil Any specific reason why we allow user's to make these choices for all these reconciliations? In legacy batch RBAC mode users/admins did not have the option to manually intervene this behavior. If we still consider manual intervention is required, could you consider addressing the following concerns:
1. namespaceSelector misconfig fails silently, 

A typo or malformed selector turns provisioning OFF everywhere. When the selector can't be parsed, namespaceMatchesSelector returns false (match nothing). So a small YAML mistake means every new namespace silently gets no pipeline SA, no SCC binding, no CA bundles — and pipelines there fail with permission errors. **TektonConfig still shows Ready, so there's no hint anything is wrong.**

2.  {} means the opposite of standard Kubernetes.
namespaceSelector: {} means "match nothing" here, but in every other Kubernetes API an empty selector means "match everything." Someone will almost certainly write {} expecting "all namespaces" and get the exact opposite.

The current design hardens only one direction, "a typo can't accidentally widen sync to the whole cluster"  but leaves the more likely mistake (accidentally turning everything off) both easy to hit and invisible.

anithapriyanatarajan (changes requested):

docs/TektonConfig.md
jkhelil: thanks @anithapriyanatarajan 
1.and 2 fixed
the reason for adding this is to avoid watching selected namespaces, eventually all, if the user dont want to opt in rbac provisioning through operator

pkg/reconciler/openshift/namespacesync/reconciler.go
```diff
@@ -0,0 +1,1036 @@
...
+	}
+
+	if !changed {
+		return nil
+	}
+
+	imagePullSecrets := sa.ImagePullSecrets
+	secretRefs := sa.Secrets
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
psrvere, line 645: The retry re-fetches the SA but then writes back the `imagePullSecrets` and `secretRefs` lists that were computed before the closure. So when the first update hits a 409, the retry overwrites whatever the other writer changed.

This happens on OpenShift as soon as the pipeline SA is created. The registry controller adds `pipeline-dockercfg-*` to the SA at the same time we update it. If we lose that race, the retry removes the pull secret. A user running `oc secrets link` at the same moment loses their secret the same way.

I reproduced this with a test. A reactor appends `pipeline-dockercfg-abcde` to the SA on the first update and returns a conflict. After Reconcile, the SA only has `quay-robot`.

I also tested a fix. Move the Get and the stale/bind merge inside the RetryOnConflict func, and remove the two captured lists. `resolveSecretBindingTargets` can stay outside. With this change the repro passes and the existing package tests still pass. Please add the conflict case to the tests so this does not regress.

pkg/apis/operator/v1alpha1/tektonconfig_validation.go
```diff
@@ -206,7 +210,27 @@ func isValueInArray(arr []string, key string) bool {
 }
 
 func isOpenShiftPlatformsSectionSet(o OpenShift) bool {
-	return o.PipelinesAsCode != nil || o.SCC != nil
+	return o.PipelinesAsCode != nil || o.SCC != nil || o.NamespaceSync != nil
+}
+
+func (ns *NamespaceSyncConfig) validate(path string) *apis.FieldError {
```
anithapriyanatarajan, line 217: A malformed namespaceSelector (invalid operator, bad label key, or In with empty values) passes CRD schema validation and gets stored, because NamespaceSyncConfig.validate() only checks secretBindings — it never validates the selector. At reconcile time LabelSelectorAsSelector returns an error and we treat that as "match nothing," so every namespace is silently skipped (no pipeline SA, SCC, or CA bundles) while TektonConfig still reports Ready.

Could we validate namespaceSelector (and secretBindings[].labelSelector) in validate() and reject on parse error, so oc apply fails fast instead of quietly disabling sync everywhere?

anithapriyanatarajan (changes requested):

pkg/apis/operator/v1alpha1/tektonconfig_validation.go
jkhelil: fixed now

anithapriyanatarajan (approved):

PR #3611, opened 2026-06-29
fix: resolve manifest list digest instead of platform-specific digest

ab-ghosh:
# Changes

Fix `get_image_sha()` in `hack/openshift/update-image-sha.sh` to resolve the manifest list digest instead of a platform-specific digest.
The function was using `.manifests[0].digest` which picks the first platform entry (amd64), causing the operator CSV to pin task images (buildah, s2i, skopeo,ubi-minimal) to amd64-only digests. This breaks ARM64/multi-arch clusters with `exec format error`.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/hack/openshift/update-image-sha.sh
+++ b/hack/openshift/update-image-sha.sh
@@ -26,10 +26,10 @@ find_latest_version() {
   echo "$version"
 }
 
-# Get SHA digest for an image:tag
-get_image_sha() {
+# Get manifest list digest for an image:tag (multi-arch)
+get_manifest_list_digest() {
   local image_url=$1
-  skopeo inspect --raw docker://${image_url} | jq -r '.manifests[0].digest // .digest'
+  skopeo inspect --no-tags docker://${image_url} | jq -r '.Digest'
 }
 
 # Update image SHA in YAML files
@@ -59,7 +59,7 @@ main() {
     echo "  Latest version: $latest_version"
 
     image_url="${image_registry}:${latest_version}"
-    image_sha=$(get_image_sha "$image_url")
+    image_sha=$(get_manifest_list_digest "$image_url")
     echo "  SHA: $image_sha"
 
     update_yaml_files "$image_registry" "$image_sha"


hack/openshift/update-image-sha.sh
```diff
@@ -26,10 +26,10 @@ find_latest_version() {
   echo "$version"
 }
 
-# Get SHA digest for an image:tag
+# Get manifest list SHA digest for an image:tag (multi-arch)
```
enarha: I'll suggest to also change the function name. It's no longer obtaining an image SHA. Instead it's generating SHA from some text input.
Another thing is that the function can return different SHA even if the image hasn't changed (extra white spaces, order of json map etc.). So instead, you can pin on the top level manifest digest, e.g.: `skopeo inspect docker://quay.io/enatan/jenkins2tekton:latest | jq -r '.Digest'
`. Here is also what AI tells me:
To support multi-arch clusters, your Operator CSV must pin to the top-level Manifest List Digest (sometimes called an index digest).
When you point Kubernetes to a multi-arch digest, the container runtime on each node (cri-o or containerd) automatically inspects the list, matches its own native CPU architecture (amd64 or arm64), and pulls the correct underlying layer.

hack/openshift/update-image-sha.sh
```diff
@@ -26,10 +26,10 @@ find_latest_version() {
...
 get_image_sha() {
   local image_url=$1
-  skopeo inspect --raw docker://${image_url} | jq -r '.manifests[0].digest // .digest'
+  skopeo inspect --raw docker://${image_url} | sha256sum | awk '{print "sha256:"$1}'
```
pramodbindal: This method of calculating manifest digest is not guaranteed to match with registered digest.

Can you try something like this

```
docker buildx imagetools inspect \
  registry.redhat.io/rhel9/postgresql-15:latest \
  --format '{{json .Manifest}}'
```

This command  gives output like this

```
sha256:0b2579cba0fd9f1754c2c9515d5f873f45de009cf044ce950284dbd1a4e6f10c
```

pramodbindal: `docker buildx imagetools inspect   registry.redhat.io/rhel9/postgresql-15:latest   --format '{{json .Manifest}}' | jq -r .digest`

hack/openshift/update-image-sha.sh
pramodbindal: skopeo inspect does not return .Digest 
I have  comment a different solution for the same

ab-ghosh: Thanks for the review. Updated and renamed to get_manifest_list_digest and using skopeo inspect | jq -r '.Digest'.

hack/openshift/update-image-sha.sh
```diff
@@ -26,10 +26,10 @@ find_latest_version() {
   echo "$version"
 }
 
-# Get SHA digest for an image:tag
-get_image_sha() {
+# Get manifest list digest for an image:tag (multi-arch)
+get_manifest_list_digest() {
   local image_url=$1
-  skopeo inspect --raw docker://${image_url} | jq -r '.manifests[0].digest // .digest'
+  skopeo inspect --no-tags docker://${image_url} | jq -r '.Digest'
```
pramodbindal, line 32: This will again return the platform specific digest only

hack/openshift/update-image-sha.sh
ab-ghosh: @pramodbindal i tested it locally and it gives the manifest list digest not sure if i am missing anything

hack/openshift/update-image-sha.sh
enarha: @pramodbindal I do not think that's correct. It returns the manifest digest, not specific image (arch one).

enarha:
@ab-ghosh if you still prefer to use the raw data as stored in the registry, you can also use something like `skopeo manifest-digest <(skopeo inspect --raw docker://quay.io/enatan/jenkins2tekton:latest)`. I'm OK with either option.

hack/openshift/update-image-sha.sh
pramodbindal: skopeo inspect (without --raw) returns the digest matching with host system platform. thats the reason spoke inspect does not work on my machine as it expects darwin/arm64

hack/openshift/update-image-sha.sh
pramodbindal: }bash-5.3$ docker buildx imagetools inspect   registry.redhat.io/rhel9/postgresql-15:latest   --format '{{json .Manifest.Digest}}' 

This command also worked for me

PR #3657, opened 2026-07-02
fix(manualapprovalgate): propagate TLS profile hash via TektonConfig

### LINKED ISSUE #3656: feat: integrate ManualApprovalGate as a TektonConfig child component
## Background

`ManualApprovalGate` was introduced in commit `9c1d1b3b5` (April 2024) as a standalone CRD with types and generated artifacts, but the integration work to make it a proper `TektonConfig`-managed component was never completed.

As a result, `ManualApprovalGate` is the only active component that is:
- **not referenced in `TektonConfig.spec`** (no user-facing field)
- **not present in `shared/tektonconfig/`** (no `EnsureMAGExists` / `GetMAGCR` helpers)
- **not wired into `shared/tektonconfig/tektonconfig.go`** (TektonConfig never creates or touches the MAG CR)

Every other component (`TektonPipeline`, `TektonTrigger`, `TektonChain`, `TektonResult`, `TektonPruner`, PAC) follows this pattern:
1. User configures the component via a field in `TektonConfig.spec`
2. TektonConfig creates the child CR with an `ownerRef` pointing to itself
3. TektonConfig writes `platform-data-hash` onto the child CR annotation whenever the cluster TLS profile changes
4. The child controller sees the annotation change and re-reconciles, picking up the new TLS env vars

## Impact

Because MAG is not wired through TektonConfig, the `platform-data-hash` propagation chain is broken for the MAG webhook deployment. When the cluster TLS profile changes (e.g. Intermediate → Modern), all other components pick up the new TLS version and cipher suites automatically — but the `manual-approval-gate-webhook` deployment stays stale.

This was observed when switching to the Modern TLS profile: the MAG webhook still showed `WEBHOOK_TLS_MIN_VERSION: 1.2` with 9 ciphers instead of the expected TLS 1.3 with 3 ciphers.

### 1. `shared/tektonconfig/manualapprovalgate/` package (~170 lines, mirrors `chain/chain.go`)

```
EnsureManualApprovalGateExists()       ← create or update the MAG CR
EnsureManualApprovalGateCRNotExists()  ← cleanup when disabled
GetManualApprovalGate()                ← fetch existing CR
CreateManualApprovalGate()             ← initial creation
UpdateManualApprovalGate()             ← update spec + annotations
GetManualApprovalGateCR()              ← build CR object with ownerRef + platform-data-hash
```

### 2. `ManualApproval` field in `TektonConfig.spec` (`pkg/apis/operator/v1alpha1/tektonconfig_types.go`)

```go
// ManualApproval holds the customizable options for the manual-approval-gate component
// +optional
ManualApproval ManualApproval `json:"manualApproval,omitempty"`
```

### 3. Wire into `shared/tektonconfig/tektonconfig.go` (same ~10-line pattern as chain)

```go
magCR := manualapprovalgate.GetManualApprovalGateCR(tc, r.operatorVersion)
if platformData := r.extension.GetPlatformData(); platformData != "" {
    magCR.Annotations[v1alpha1.PlatformDataHashKey] = platformData
}
if _, err := manualapprovalgate.EnsureManualApprovalGateExists(ctx, r.operatorClientSet..., magCR); err != nil {
    ...
}
```

### 4. No controller changes needed

Once TektonConfig writes `platform-data-hash` onto the MAG CR, the existing MAG controller event handlers (`FilterController` + `EnqueueControllerOf` on TektonInstallerSet) pick up the change automatically — exactly like TektonChain does today.

## Migration concern

Currently users apply the `ManualApprovalGate` CR directly. Once TektonConfig manages it, the CR would be created by TektonConfig with an `ownerRef`. Existing standalone MAG CRs should be updated to add the `ownerRef` during migration (similar to how other components handle upgrades).

## Related

- Original MAG introduction: `9c1d1b3b5` — _"Adds Manual Approval Gate types and generated artifacts"_
- TLS propagation fix: SRVKP-9613

jkhelil:
## Summary

`ManualApprovalGate` is a standalone CR (not created or owned by TektonConfig). As a result it was never part of the `platform-data-hash` propagation chain, so when the cluster TLS profile changed (e.g. Intermediate → Modern), the MAG webhook deployment stayed stale — still showing `WEBHOOK_TLS_MIN_VERSION: 1.2` with 9 ciphers instead of TLS 1.3 with 3 ciphers.

## Root cause

The `InstallerSetClient` detects TLS changes by including `platform-data-hash` in the InstallerSet hash computation. TektonConfig writes this annotation onto every CR it owns (`TektonPipeline`, `TektonChain`, `TektonResult`, etc.) during `PostReconcile`. MAG was never wired into this flow because it was introduced as a standalone CRD without the TektonConfig integration (see #3656).

## Fix

Added `propagateMAGPlatformData()` called from `OpenShiftExtension.PostReconcile()`. It:

1. Computes the current TLS profile hash via `GetPlatformData()`
2. Lists all existing `ManualApprovalGate` CRs
3. Writes the hash into `operator.tekton.dev/platform-data-hash` on each CR (skipping if already up to date)

The existing MAG controller informer fires on the annotation change, triggers a reconcile, and the webhook deployment is re-applied with the correct TLS env vars. **No changes to the MAG controller or reconciler are needed.**

## Safety

| Scenario | Behaviour |
|---|---|
| No MAG CR installed | `List` returns empty — loop is a no-op |
| MAG CRD not present | `List` returns error — logged as warning, `PostReconcile` continues |
| First-time hash write | InstallerSet hash changes → one-time reconcile → correct TLS applied |
| Hash already current | Annotation unchanged → no API call → no reconcile triggered |

## Long-term fix

Proper integration of MAG as a full TektonConfig child (ownerRef, spec field, shared package) is tracked in #3656. This PR is the minimal correct stopgap that follows the same architectural pattern as PAC propagation in `PostReconcile`.

Relates-To: SRVKP-9613
Closes #3656 (partially — the immediate TLS symptom; full child integration deferred)

### DIFF
--- a/pkg/reconciler/openshift/tektonconfig/extension.go
+++ b/pkg/reconciler/openshift/tektonconfig/extension.go
@@ -217,9 +217,47 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
 		}
 	}
 
+	// Propagate platform-data-hash to any existing ManualApprovalGate CR.
+	// ManualApprovalGate is a standalone CR (not created by TektonConfig — see
+	// https://github.com/tektoncd/operator/issues/3656), so it never receives
+	// platform-data-hash through the normal child-CR path. We update it here
+	// (same PostReconcile layer as PAC) so that the MAG controller re-applies
+	// the webhook deployment with updated TLS env vars when the cluster TLS
+	// profile changes.
+	oe.propagateMAGPlatformData(ctx)
+
 	// execute console plugin reconciler
 	// TLS config was already resolved and cached in PreReconcile via SetTLSConfig.
 	return oe.consolePluginReconciler.reconcile(ctx, configInstance)
+}
+
+// propagateMAGPlatformData writes the current TLS profile hash into the
+// platform-data-hash annotation of every existing ManualApprovalGate CR.
+// It is a best-effort operation — failures are logged but do not block the
+// TektonConfig reconciliation.
+func (oe openshiftExtension) propagateMAGPlatformData(ctx context.Context) {
+	platformData := oe.GetPlatformData()
+	if platformData == "" {
+		return
+	}
+	logger := logging.FromContext(ctx)
+	magList, err := oe.operatorClientSet.OperatorV1alpha1().ManualApprovalGates().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		logger.Warnf("failed to list ManualApprovalGate CRs for platform-data-hash propagation: %v", err)
+		return
+	}
+	for i := range magList.Items {
+		mag := &magList.Items[i]
+		if mag.Annotations[v1alpha1.PlatformDataHashKey] == platformData {
+			continue
+		}
+		patch := fmt.Sprintf(`{"metadata":{"annotations":{%q:%q}}}`, v1alpha1.PlatformDataHashKey, platformData)
+		if _, patchErr := oe.operatorClientSet.OperatorV1alpha1().ManualApprovalGates().Patch(
+			ctx, mag.Name, types.MergePatchType, []byte(patch), metav1.PatchOptions{},
+		); patchErr != nil {
+			logger.Warnf("failed to patch platform-data-hash on ManualApprovalGate %s: %v", mag.Name, patchErr)
+		}
+	}
 }
 
 func (oe openshiftExtension) GetPlatformData() string {


pkg/reconciler/openshift/tektonconfig/extension.go
```diff
@@ -217,11 +217,49 @@ func (oe openshiftExtension) PostReconcile(ctx context.Context, comp v1alpha1.Te
...
+// platform-data-hash annotation of every existing ManualApprovalGate CR.
+// It is a best-effort operation — failures are logged but do not block the
+// TektonConfig reconciliation.
+func (oe openshiftExtension) propagateMAGPlatformData(ctx context.Context) {
+	platformData := oe.GetPlatformData()
+	if platformData == "" {
+		return
+	}
+	logger := logging.FromContext(ctx)
+	magList, err := oe.operatorClientSet.OperatorV1alpha1().ManualApprovalGates().List(ctx, metav1.ListOptions{})
```
pramodbindal, line 244: Can there be more than one MAG crs ?

PR #3688, opened 2026-07-09
fix(pruner): ensure enforcedConfigLevel is always defaulted and DeepCopy is correct

infernus01:
# Changes
  - Fix `enforcedConfigLevel` silently vanishing from the pruner global config
    when a user provides a partial `global-config` (e.g. only `ttlSecondsAfterFinished`)
  - Fix `DeepCopyInto` shallow copy bug that prevented the defaulting webhook
    from detecting changes made by `SetDefaults`

  ## Problem

  When a user provides any `global-config` field without explicitly setting
  `enforcedConfigLevel`, the v0.79.1 `SetDefaults` skips defaulting entirely
  because `GlobalConfig` is non-nil. Combined with `omitempty` struct tags in
  the vendored pruner, `enforcedConfigLevel` is omitted from the ConfigMap.
  The pruner controller then falls back to the internal default (`resource`
  level), which never consults namespace-level ConfigMaps — all selector-based
  pruning rules silently stop working.

  On top of this, `DeepCopyInto` for `TektonPrunerConfig` does `*out = *in`,
  which copies the `*GlobalConfig` pointer instead of the struct. The Knative
  defaulting webhook deep-copies the object before calling `SetDefaults`, then
  compares it with the original to generate a patch. Because both copies share
  the same pointer, the webhook always sees no difference (`PatchBytes: null`)
  and never applies the defaults.

  ## Root Cause

  Commit 774867c changed `GlobalConfig` from a value type to a pointer and
  guarded `SetDefaults()` with `if p.GlobalConfig == nil` to prevent
  `historyLimit` from resetting to 100 when set to null. The side effect was
  that `enforcedConfigLevel` was never filled in for partial configs. The
  existing `DeepCopyInto` was not updated for the pointer type change.

  ## Fix

  **`tektonpruner_defaults.go`**: Always fill in `enforcedConfigLevel` when
  nil, without touching other fields:

  ```go
  if p.GlobalConfig == nil {
      p.GlobalConfig = &config.GlobalConfig{}
      p.GlobalConfig.SetDefaults()
  } else if p.GlobalConfig.EnforcedConfigLevel == nil {
      v := config.EnforcedConfigLevelGlobal
      p.GlobalConfig.EnforcedConfigLevel = &v
  }

  tektonpruner_types.go: Deep-copy the GlobalConfig pointer:

  func (in *TektonPrunerConfig) DeepCopyInto(out *TektonPrunerConfig) {
      *out = *in
      if in.GlobalConfig != nil {
          out.GlobalConfig = new(config.GlobalConfig)
          *out.GlobalConfig = *in.GlobalConfig
      }
  }
```

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonpruner_defaults.go
+++ b/pkg/apis/operator/v1alpha1/tektonpruner_defaults.go
@@ -35,5 +35,8 @@ func (p *Pruner) SetDefaults() {
 	if p.GlobalConfig == nil {
 		p.GlobalConfig = &config.GlobalConfig{}
 		p.GlobalConfig.SetDefaults()
+	} else if p.GlobalConfig.EnforcedConfigLevel == nil {
+		v := config.EnforcedConfigLevelGlobal
+		p.GlobalConfig.EnforcedConfigLevel = &v
 	}
 }
--- a/pkg/apis/operator/v1alpha1/tektonpruner_types.go
+++ b/pkg/apis/operator/v1alpha1/tektonpruner_types.go
@@ -111,5 +111,8 @@ func (p *Pruner) IsDisabled() bool {
 
 func (in *TektonPrunerConfig) DeepCopyInto(out *TektonPrunerConfig) {
 	*out = *in
-	return
+	if in.GlobalConfig != nil {
+		out.GlobalConfig = new(config.GlobalConfig)
+		*out.GlobalConfig = *in.GlobalConfig
+	}
 }


anithapriyanatarajan:
@infernus01 did we verify the  impact on upgrade due to this change?

infernus01:
@anithapriyanatarajan There can be 4 cases for the upgrade scenario:
Case 1: User explicitly set enforcedConfigLevel: namespace
The else if p.GlobalConfig.EnforcedConfigLevel == nil is false — their value is non-nil. Nothing changes. No impact.

Case 2: User explicitly set enforcedConfigLevel: global
Same — non-nil, skipped. No impact.

Case 3: User never set enforcedConfigLevel, but old operator (<=v0.78.1) filled it in
On versions <=v0.78.1, SetDefaults always ran and set global. The field already exists. This fix skips it. No impact.

Case 4: User never set enforcedConfigLevel, running v0.79.1+ where it was silently dropped
This is the only case where behavior changes. Let's say, enforcedConfigLevel is missing from the ConfigMap(which can, without this fix). A user sets per-namespace overrides in their global config:
```
  tektonpruner:
    global-config:
      ttlSecondsAfterFinished: 60
      namespaces:
        dev:
          ttlSecondsAfterFinished: 30
```

Without explicitly setting enforcedConfigLevel. The broken SetDefaults never filled it in, so it's missing from the ConfigMap. The pruner's fallback is resource level, which happens to check globalSpec.Namespaces["dev"] and finds TTL=30. And this I haven't yet verified will work or not. But after this fix, enforcedConfigLevel: global is filled in and at global level, the pruner only uses root-level settings. The `namespaces:` section is skipped. The dev namespace override will not work.
So this case is technically possible but unlikely in practice.

pramodbindal (approved):

PR #3689, opened 2026-07-09
fix(release): bump pinned ko/koparse image for go 1.26.4

jkhelil:
# Changes

`tekton/build-publish-images-manifests.yaml` pinned the ko/kustomize
build steps to `ghcr.io/tektoncd/plumbing/ko-gcloud:v20240920-6c2a999d36`,
which ships Go 1.22.1. `go.mod` now requires `go 1.26.4` (CVE-2026-27145
stdlib fix), and no `ko-gcloud` digest published yet ships Go >= 1.26.4.

- Switch the `rewrite-devel-in-source` and `run-kustomize-ko` steps to
  the plain `ko` image instead, which already ships Go 1.26.4 and
  includes every tool they actually use (`ko`, `kustomize`, `git`)
  without needing the gcloud/kubectl/yq extras that `ko-gcloud` bundles.
- Bump the `koparse` step to its current digest.
- Add a `docker` package-ecosystem entry for `/tekton` to
  `.github/dependabot.config.yml` (regenerated `.github/dependabot.yml`
  via `./hack/generate-dependabot.sh`) so these pinned image digests are
  kept in sync with `go.mod` going forward instead of silently drifting
  behind the required Go toolchain version.

Verified by pulling both the old and new images and confirming
`go version` inside each, and that the new `ko` image has `tar`, `sed`,
`awk`, `git`, `kustomize`, and `ko` available.

## Follow-up fix: kodata symlink rejected by ko >= v0.19

The `ko` image now pinned here (aligned with `tektoncd/pipeline`'s digest)
ships `ko` v0.19.1, which enforces that `kodata/` symlinks must not resolve
outside the kodata root (ko-build/ko#1619). The `run-kustomize-ko` step
symlinks `cmd/${KUBE_DISTRO}/*/kodata/source.tar.gz` to a tarball in an
external `mktemp -d` directory, which this newer `ko` now rejects:

```
Error: error processing import paths ...: tarring kodata: kodata symlink
"cmd/<distro>/operator/kodata/source.tar.gz" resolves to "/tmp/tmp.XXXXXX/source.tar.gz"
which is outside the kodata root "cmd/<distro>/operator/kodata"
```

We hit this exact failure in `tektoncd/chains`' release pipeline after a
similar Go 1.26.4-driven `ko` digest bump
([tektoncd/chains#1791](https://github.com/tektoncd/chains/pull/1791)),
and the same unfixed pattern also exists upstream in
`tektoncd/pipeline`'s `tekton/publish.yaml`
([tektoncd/pipeline#10418](https://github.com/tektoncd/pipeline/pull/10418)).
Since this PR moves operator onto the same `ko` v0.19.1 digest, it would hit
the identical error on the next real release, so the fix is included here.

Fix: copy the tarball into each `cmd/${KUBE_DISTRO}/*/kodata/` directory
instead of symlinking to it from a shared tmpdir. `ko` already dereferences
symlinks and embeds the resolved file's bytes into the image layer when
tarring kodata, so this has no effect on final image size — only the
symlink itself was rejected, not the underlying content.

# Submitter Checklist

- [ ] Includes tests (if functionality changed/added)
- [ ] Includes docs (if user facing)
- [x] Commit messages follow commit message best practices

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.config.yml
+++ b/.github/dependabot.config.yml
@@ -66,3 +66,22 @@ ecosystems:
       semver-major-days: 30
       semver-minor-days: 7
       semver-patch-days: 3
+  # Keeps the pinned ko/koparse image digests in tekton/build-publish-images-manifests.yaml
+  # in sync so they don't silently drift behind the Go toolchain version required by go.mod
+  # (see https://github.com/tektoncd/chains/pull/1784 for the class of failure this prevents:
+  # "go.mod requires go >= X (running go Y; GOTOOLCHAIN=local)" during release builds).
+  - package-ecosystem: docker
+    directory: /tekton
+    schedule:
+      interval: weekly
+    labels:
+      - ok-to-test
+      - dependencies
+      - kind/misc
+      - release-note-none
+    groups:
+      all:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -43,6 +43,21 @@ updates:
         semver-major-days: 30
         semver-minor-days: 7
         semver-patch-days: 3
+    - package-ecosystem: docker
+      directory: /tekton
+      schedule:
+        interval: weekly
+      labels:
+        - ok-to-test
+        - dependencies
+        - kind/misc
+        - release-note-none
+      groups:
+        all:
+            patterns:
+                - '*'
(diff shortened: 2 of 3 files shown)


vdemeester (approved):

infernus01 (commented):
/lgtm

PR #3709, opened 2026-07-13
feat(results): expose Tekton Results Watcher config via TektonConfig

adchauha:
Add spec.result.watcher so watcher flags can be configured through TektonConfig without manually patching tekton-results-watcher Deployment args.

# Changes
- Add `ResultsWatcherProperties` and `spec.result.watcher` to TektonConfig and TektonResult CRDs
- Wire watcher configuration into `tekton-results-watcher` Deployment container args via `UpdateWatcherFlagsInDeployment`
- Add validation for watcher fields (label selector syntax, non-negative durations)
- Regenerate CRDs and deepcopy codegen
- Add unit tests for watcher validation and deployment arg transformation
- Document watcher configuration in `docs/TektonConfig.md` and `docs/TektonResult.md`

Watcher flags exposed include:
- `completed_run_grace_period`
- `check_owner`
- `store_deadline`
- `disable_storing_incomplete_runs`
- `logs_api`, `logs_timestamps`, `store_event`
- `summary_labels`, `summary_annotations`, `label_selector`
- `requeue_interval`, `forward_buffer`, `update_log_timeout`, `dynamic_reconcile_timeout`, `disable_crd_update`

# Submitter Checklist

review them:

# Release Notes

```release-note
TektonConfig and TektonResult now support `spec.result.watcher` to configure Tekton Results Watcher behavior (for example `completed_run_grace_period`, `check_owner`, `store_deadline`, and `disable_storing_incomplete_runs`) without manually editing the `tekton-results-watcher` Deployment.
```


enarha (commented):
The idea of storing the watcher configuration under .result.watcher is good.

The watcher has 21 configuration options (https://github.com/tektoncd/results/blob/main/cmd/watcher/main.go#L64-L85) , but here I see exposed only 15. Was that done on purpose and what's the criteria to decide which flags are exposed through the TektonConfig and which not?

I believe that as implemented, the new flags will be set on new deployment, but if user changes the TektonConfig later, the configuration on the cluster won't be updated. Can you please test that scenario and ensure it works? Updating the TektonConfig should always trigger reconciliation and new configuration should take an effect immediately.

pkg/reconciler/common/transformers.go
```diff
@@ -1240,6 +1240,76 @@ func UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(performanceSpec *v1alp
...
+		}
+
+		podLabels := dep.Spec.Template.Labels
+		if podLabels == nil {
+			podLabels = map[string]string{}
+		}
+		labelKeys := getSortedKeys(flags)
+		for _, key := range labelKeys {
+			labelKey := fmt.Sprintf("%s.data.%s", deploymentName, key)
+			podLabels[labelKey] = fmt.Sprintf("%v", flags[key])
```
enarha: This writes every watcher setting into a Kubernetes label, but a few of these settings (summary_labels, summary_annotations, label_selector) are free-text strings that can contain characters Kubernetes doesn't allow in label values — like /, ,, or =.

For example, your own docs example sets summary_labels: tekton.dev/pipeline. If that value gets written as a label, Kubernetes will reject it with a validation error (label values can only contain letters, numbers, -, _, .), and the whole deployment will fail to apply.

Suggestion: don't put the raw value into the label. Instead, use something like a hash of the value (e.g. fmt.Sprintf("%x", md5.Sum([]byte(value)))) as the label value. The label here is only used to force the pod to restart when a setting changes — it doesn't need to be human-readable, so a hash works and is always a valid label value.

pkg/reconciler/kubernetes/tektonresult/transform_test.go
```diff
@@ -391,3 +396,68 @@ func Test_AddConfiguration(t *testing.T) {
 	// Verify PriorityClassName was applied
 	assert.Equal(t, deployment.Spec.Template.Spec.PriorityClassName, "system-cluster-critical")
 }
+
+func TestUpdateWatcherFlagsInDeployment(t *testing.T) {
```
enarha, line 396: This only tests the happy path, but more importantly tests should be testing the edge cases, invalid values, repeated values, errors handling, etc.

pkg/reconciler/common/transformers.go
```diff
@@ -1240,6 +1240,76 @@ func UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(performanceSpec *v1alp
...
+			if container.Name != containerName {
+				continue
+			}
+			for _, flagKey := range flagKeys {
+				expectedArg := fmt.Sprintf("-%s", flagKey)
+				argStringValue := fmt.Sprintf("%v", flags[flagKey])
+
+				argUpdated := false
+				for argIndex, existingArg := range container.Args {
+					if strings.HasPrefix(existingArg, expectedArg) {
```
enarha: This code assumes each existing flag in the container's args list looks like a single string, e.g. "-check_owner=true". But if you look at the base watcher deployment manifest, some flags are written as two separate list entries instead, e.g.:
args:
  - -api_addr
  - $(TEKTON_RESULTS_API_SERVICE)
If one of the flags you're now configuring is ever defined that way, this code would overwrite just the -api_addr entry with -api_addr=value, but leave the old value ($(TEKTON_RESULTS_API_SERVICE)) sitting in the list as a stray, unrelated argument. That stray value would break how the watcher parses its command-line flags (it would think that's the next positional argument and stop reading further flags).

It happens not to bite you today because none of the 15 flags you're adding currently exist in the base manifest's args list. But it's worth handling now so it doesn't silently break later (e.g. if this helperis reused for api_addr/auth_mode, or if the watcher's manifest changes format upstream).

Suggestion: when you find a matching flag, check if the next array element is a plain value (not startingwith -) and remove it too before inserting the combined -flag=value form. Or simpler: always remove the old flag (and its value, if separate) first, then append the new -flag=value — that avoids the two different “shapes” colliding.

pkg/apis/operator/v1alpha1/tektonresult_validation.go
```diff
@@ -57,5 +57,8 @@ func (trs *TektonResultSpec) validate(path string) (errs *apis.FieldError) {
 	// validate performance properties
 	errs = errs.Also(trs.Performance.Validate(fmt.Sprintf("%s.performance", path)))
 
+	// validate watcher properties
+	errs = errs.Also(trs.Watcher.Validate(fmt.Sprintf("%s.watcher", path)))
```
enarha, line 61: This validation only runs when someone creates/edits a TektonResult object directly. But the PR description says the main way users will set this is through TektonConfig (spec.result.watcher) — and TektonConfig's own validation (pkg/apis/operator/v1alpha1/tektonconfig_validation.go, around line 143) never calls this Watcher.Validate(). It only validates tc.Spec.Result.Options.

That means if a user sets an invalid label_selector or a negative duration in TektonConfig, Kubernetes will accept it right away (no error shown to the user). The mistake would only be caught later, indirectly, when the operator tries to create the underlying TektonResult object — and even then it might not be obvious to the user why it failed.

Suggestion: add a line in tektonconfig_validation.go next to the existing tc.Spec.Result.Options.validate(...) call, e.g.:
errs = errs.Also(tc.Spec.Result.Watcher.Validate("spec.result.watcher")) so users get immediate, clear feedback when they make a mistake in TektonConfig — which is the entry point most people will actually use.

jkhelil:
@adchauha Can you please adress @enarha comments. Thank you

adchauha:
> The idea of storing the watcher configuration under .result.watcher is good.
> 
> The watcher has 21 configuration options (https://github.com/tektoncd/results/blob/main/cmd/watcher/main.go#L64-L85) , but here I see exposed only 15. Was that done on purpose and what's the criteria to decide which flags are exposed through the TektonConfig and which not?
> 
> I believe that as implemented, the new flags will be set on new deployment, but if user changes the TektonConfig later, the configuration on the cluster won't be updated. Can you please test that scenario and ensure it works? Updating the TektonConfig should always trigger reconciliation and new configuration should take an effect immediately.

It is intentional. Under spec.result.watcher we expose user-facing behavior flags (grace period, store deadline, selectors, timeouts, etc.). Performance knobs (threadiness/qps/burst/disable-ha) already live under spec.result.performance. Connection/install wiring (api_addr, auth_mode) comes from the Results manifest. Token must not live in the CR, use Secrets. Documented in docs/TektonConfig.md / ResultsWatcherProperties.

UpdateResult synced Performance/Options but not Watcher, so later TektonConfig edits could be ignored. Fixed by syncing Spec.Watcher the same way, plus TestUpdateResult_WatcherPropagation. Updating TektonConfig watcher fields now reconciles through to the live Deployment.

pkg/reconciler/common/transformers.go
adchauha: You’re right, Free-text fields like summary_labels / label_selector can include /, =, ,, which are invalid as Kubernetes label values.
In UpdateWatcherFlagsInDeployment, pod-template label values are now a truncated SHA-256 hex digest of the setting (hashLabelValue), instead of the raw string. Container args still receive the real value so watcher behavior is unchanged. I used SHA-256 (truncated to 32 chars to stay under the 63-char label limit) rather than MD5, consistent with other hashing in this repo.
Added/updated unit tests also.

pkg/reconciler/kubernetes/tektonresult/transform_test.go
adchauha: The original test only covered the happy path. I expanded coverage in transform_test.go:
- Overwrite / two arg shapes: existing -check_owner=true and two-element -summary_labels / old-value are replaced cleanly, stray old-value is removed.
- Free-text + labels: summary_labels / label_selector appear as real values in args, but pod-template labels store hashes (not raw strings). Changing a value changes the hash
- No-op edge cases: new TestUpdateWatcherFlagsInDeployment_NoOp for nil watcher, empty watcher, and non-matching Deployment name
Invalid user input (bad selectors, negative durations) is covered in watcher / TektonConfig validation tests, since that should fail at admission rather than inside the transformer.

pkg/reconciler/common/transformers.go
adchauha: Introduced replaceOrAppendContainerArg, which:
1. Matches either -flag or -flag=...
2. If the match is bare -flag and the next entry is a value (doesn’t start with -), removes that too
3. Inserts a single combined -flag=value (or appends if the flag wasn’t present)
UpdateWatcherFlagsInDeployment now uses this helper for every watcher flag. Unit tests cover both shapes, including asserting that a leftover old-value from the two-element form is removed.

pkg/apis/operator/v1alpha1/tektonresult_validation.go
adchauha: Changes made in tektonconfig_validation.go:
errs = errs.Also(tc.Spec.Result.Watcher.Validate("spec.result.watcher"))
This reuses the same ResultsWatcherProperties.Validate logic already used by TektonResult, so users get immediate admission feedback on TektonConfig. Added Test_ValidateTektonConfig_ResultWatcher to cover that path.

pkg/apis/operator/v1alpha1/watcher_types.go
```diff
@@ -0,0 +1,96 @@
...
+	// +optional
+	LogsTimestamps *bool `json:"logs_timestamps,omitempty"`
+
+	// Store Kubernetes events related to TaskRuns and PipelineRuns.
+	// +optional
+	StoreEvent *bool `json:"store_event,omitempty"`
+
+	// Comma-separated label keys copied into the Result summary.
+	// +optional
+	SummaryLabels string `json:"summary_labels,omitempty"`
```
khrm: If we don't use *string , it means we don't allow clearing any default.

pkg/apis/operator/v1alpha1/watcher_types.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Comma-separated annotation keys copied into the Result summary.
+	// +optional
+	SummaryAnnotations string `json:"summary_annotations,omitempty"`
```
khrm: Same here. It's fine tbh. Just document that "" won't clear the default value. Or you can use a pointer to for this

pkg/reconciler/common/transformers.go
```diff
@@ -1240,6 +1241,107 @@ func UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(performanceSpec *v1alp
...
+	expectedArg := fmt.Sprintf("-%s", flagKey)
+	newArg := fmt.Sprintf("%s=%s", expectedArg, value)
+
+	out := make([]string, 0, len(args)+1)
+	replaced := false
+	for i := 0; i < len(args); i++ {
+		arg := args[i]
+		if arg == expectedArg || strings.HasPrefix(arg, expectedArg+"=") {
+			// Skip a separate value entry from the two-element form.
+			if arg == expectedArg && i+1 < len(args) && !strings.HasPrefix(args[i+1], "-") {
```
khrm: What would happen in case of `-completed_run_grace_period -1h`  which means delete immediately. https://github.com/tektoncd/results/blob/da051e620e19bc0491963ef2a86ae996e82a92b9/cmd/watcher/main.go#L68

pkg/apis/operator/v1alpha1/watcher_types.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Label selector for Runs eligible for deletion after the grace period.
+	// +optional
+	LabelSelector string `json:"label_selector,omitempty"`
```
khrm: Same *string vs string issue here.

pkg/reconciler/common/transformers.go
```diff
@@ -1240,6 +1241,107 @@ func UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(performanceSpec *v1alp
...
+		}
+
+		podLabels := dep.Spec.Template.Labels
+		if podLabels == nil {
+			podLabels = map[string]string{}
+		}
+		// Hash label values so free-text settings (summary_labels, label_selector, etc.)
+		// always produce valid Kubernetes label values. Labels only force a pod restart
+		// when config changes; they do not need to be human-readable.
+		labelKeys := getSortedKeys(flags)
```
khrm: Why are we adding these labels? Or is it something that was already done? I don't recall this.

pkg/reconciler/common/transformers.go
```diff
@@ -1240,6 +1241,107 @@ func UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(performanceSpec *v1alp
...
+		for _, key := range labelKeys {
+			labelKey := fmt.Sprintf("%s.data.%s", deploymentName, key)
+			podLabels[labelKey] = hashLabelValue(fmt.Sprintf("%v", flags[key]))
```
khrm: Another way can be to use the prefix form.

pkg/apis/operator/v1alpha1/watcher_types.go
adchauha: switched these fields to *string so nil=unset and empty=clear.

pkg/apis/operator/v1alpha1/watcher_types.go
adchauha: switched these fields to *string so nil=unset and empty=clear.

pkg/apis/operator/v1alpha1/watcher_types.go
adchauha: switched these fields to *string so nil=unset and empty=clear.

pkg/reconciler/common/transformers.go
adchauha: Updated replaceOrAppendContainerArg to treat negative numeric/duration values as the flag value, and added a unit test.

pkg/reconciler/common/transformers.go
adchauha: Same rollout pattern as UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap: fingerprint config on the pod template so changes force a restart. Values are hashed so free-text settings remain valid labels.

pkg/reconciler/common/transformers.go
adchauha: Kept the existing *.data.* style for consistency with performance flags.

config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
```diff
@@ -531,6 +531,8 @@ spec:
                           type: string
                       type: object
                     type: array
+                required:
+                - options
```
enarha: This seems unrelated. I also mentioned it on your other PR.

enarha: https://github.com/tektoncd/operator/pull/3842 seems related. We might need to merge it first, then you can rebase and regenerate the manifests.

enarha (commented):
Just one comment, the rest is LGTM.

config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
adchauha: Done

pkg/reconciler/common/transformers.go
```diff
@@ -1286,6 +1287,127 @@ func UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap(performanceSpec *v1alp
 	}
 }
 
+// UpdateWatcherFlagsInDeployment injects Tekton Results Watcher configuration
+// as command-line flags on the watcher deployment container.
+func UpdateWatcherFlagsInDeployment(watcher *v1alpha1.ResultsWatcherProperties, deploymentName, containerName string) mf.Transformer {
```
pratap0007: This seems to be Results-specific. Can we move it to the Results transformer instead of the common package?

adchauha: Thanks for the suggestion. I kept it in common intentionally to follow the same pattern as "UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap", which already lives there and is shared across components (including Results). "UpdateWatcherFlagsInDeployment" is the same kind of deployment-arg transformer (plus pod-label fingerprinting for rollout), so I think common is the consistent place for now. Happy to move it later if we want Results only.

khrm: But it's quite hard coded with results specific deployment and can't be reused by other components.

adchauha: Moved UpdateWatcherFlagsInDeployment (and its helpers) into the Results transformer package.

pratap0007:
@adchauha  Thank you for updating the PR and could you please squash all the commits into a single commit?

PR #3710, opened 2026-07-13
fix(common): apply proxy settings to StatefulSets too

jkhelil:
# Changes

`ApplyProxySettings` only handled resources of kind `Deployment`, so when
`spec.pipeline.performance.statefulset-ordinals` is enabled in `TektonConfig`
and the `tekton-pipelines-controller` / `tekton-pipelines-remote-resolvers`
Deployments are converted into StatefulSets, the resulting StatefulSet pods
never inherited the cluster-wide proxy environment variables (`HTTP_PROXY`,
`HTTPS_PROXY`, `NO_PROXY`) that Deployment-based pods correctly pick up.

This PR:
- Extends `ApplyProxySettings` to also process `StatefulSet` resources
  (it's already invoked for both kinds from `tektoninstallerset` `ensureResource`,
  but silently no-op'd for StatefulSets).
- Hardens the containers lookup so a present-but-`null` `containers` field
  (e.g. an empty/placeholder StatefulSet) is treated as "no containers"
  instead of returning a type-mismatch error.
- Adds unit tests covering StatefulSet proxy injection/removal and confirms
  other resource kinds remain unaffected.

Fixes: https://redhat.atlassian.net/browse/SRVKP-12831

# Submitter Checklist

# Release Notes

```release-note
Fix StatefulSet pods (used when `statefulset-ordinals` is enabled) not
inheriting cluster-wide proxy environment variables.
```

### DIFF
--- a/pkg/reconciler/common/proxy.go
+++ b/pkg/reconciler/common/proxy.go
@@ -24,10 +24,10 @@ import (
 )
 
 // ApplyProxySettings is a transformer that propagate any proxy environment variables
-// set on the operator deployment to the underlying deployment.
+// set on the operator deployment to the underlying deployment or statefulset.
 func ApplyProxySettings(u *unstructured.Unstructured) error {
-	if u.GetKind() != "Deployment" {
-		// Don't do anything on something else than Deployment
+	if u.GetKind() != "Deployment" && u.GetKind() != "StatefulSet" {
+		// Don't do anything on something else than Deployment or StatefulSet
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


l-qing:
Thanks for this fix @jkhelil. All checks are green and it already has a `/lgtm`, but tide is blocking the merge with `Not mergeable. PR can't be rebased` — the branch has drifted from `main` and Prow can't auto-rebase it.

Could you rebase `fix/statefulset-proxy-settings-SRVKP-12831` onto the latest `main` and force-push? That should clear the tide block and let it merge. Thanks!

PR #3762, opened 2026-07-20
feat(manualapprovalgate): integrate MAG as TektonConfig child component

### LINKED ISSUE #3656: feat: integrate ManualApprovalGate as a TektonConfig child component

pratap0007:
## Changes

Add ManualApproval field to TektonConfig spec so that TektonConfig manages the ManualApprovalGate CR lifecycle including ownerRef and platform-data-hash propagation for TLS profile changes.

- Disabled by default on fresh installs
- On upgrade, standalone MAG CRs (no ownerRef) are adopted under TektonConfig instead of being deleted
- Add validation for spec.manualApproval.options
- Add unit tests for create, delete, ownerRef migration, and platform-data-hash annotation sync

Fixes #3656

Assisted-by: Claude Opus 4.6 (via Claude Code)

# Testing

- Verified fresh installation with MAG enabled and disabled; both scenarios work as expected.
- Verified the upgrade scenario with an existing MAG installation that is adopted by the TektonConfig; the upgrade works as expected.

# Submitter Checklist

review them:

# Release Notes

```release-note
Add ManualApproval component to the TektonConfig to manage the ManualApprovalGate CR lifecycle.
```


pkg/apis/operator/v1alpha1/manualapprovalgate_types.go
```diff
@@ -51,11 +51,25 @@ type ManualApprovalGateSpec struct {
...
 type ManualApproval struct {
+	// enable or disable manual approval gate feature
+	// +optional
+	Disabled *bool `json:"disabled,omitempty"`
 	// options holds additions fields and these fields will be updated on the manifests
 	// +optional
 	Options AdditionalOptions `json:"options"`
 }
 
+func (m *ManualApproval) setDefaults() {
```
anithapriyanatarajan: Minor to follow pattern: setDefaults() and IsDisabled() belong in a dedicated manualapprovalgate_defaults.go, not _types.go. All other components follow this split: Chain.setDefaults() -> tektonchain_defaults.go, Pruner.SetDefaults() -> tektonpruner_defaults.go. The _types.go file should only contain struct definitions. Please move these two methods to a new manualapprovalgate_defaults.go and also add a top-level func (mag *ManualApprovalGate) SetDefaults(_ context.Context) that calls mag.Spec.ManualApproval.setDefaults()  that's the entry point the webhook calls. This violates the LAYOUT defined in [AGENTS.md](https://github.com/tektoncd/operator/blob/main/AGENTS.md?plain=1#L81-L84)

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -26,6 +26,7 @@ import (
 	tektonConfigreconciler "github.com/tektoncd/operator/pkg/client/injection/reconciler/operator/v1alpha1/tektonconfig"
 	"github.com/tektoncd/operator/pkg/reconciler/common"
 	"github.com/tektoncd/operator/pkg/reconciler/shared/tektonconfig/chain"
+	"github.com/tektoncd/operator/pkg/reconciler/shared/tektonconfig/manualapprovalgate"
```
anithapriyanatarajan, line 29: ManualApprovalGate is not registered in [webhook.go'](https://github.com/tektoncd/operator/blob/main/pkg/webhook/webhook.go)s types map. This means SetDefaults and Validate are never called when users create or edit MAG CRs directly. Every component that has a standalone reconciler and is in AGENTS.md's "pattern references" table is registered there (TektonPipeline, TektonChain, TektonResult, TektonPruner, etc.). Please add MAG to the types map.

anithapriyanatarajan (changes requested):

anithapriyanatarajan:
@pratap0007 please update docs https://github.com/tektoncd/operator/blob/main/docs/TektonConfig.md to indicate that with this PR MAG configs can now be managed with tektonconfig. Thank you.

pratap0007:
@anithapriyanatarajan Thank you for the review, I've updated the PR and could you please take another look

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -320,6 +324,45 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
 	}
 
+	// Ensure ManualApprovalGate CR
+	// On upgrade: if a standalone MAG CR exists (no ownerRef, from a previous version),
+	// adopt it under TektonConfig regardless of the disabled setting.
+	// On fresh install: MAG is disabled by default and not created.
+	magEnabled := !tc.Spec.ManualApproval.IsDisabled()
+	if !magEnabled {
+		existingMAG, err := manualapprovalgate.GetManualApprovalGate(ctx, r.operatorClientSet.OperatorV1alpha1().ManualApprovalGates(), v1alpha1.ManualApprovalGates)
+		if err == nil && (tc.Spec.ManualApproval.Disabled == nil || len(existingMAG.OwnerReferences) == 0) {
```
jkhelil: Any error other than "not found" (e.g. a transient API error, RBAC issue) is silently treated the same as "no standalone CR found" with no log line. Consider logging non-NotFound errors so real problems (e.g. missing RBAC for manualapprovalgates) aren't invisible.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -320,6 +324,45 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
...
+			logger.Debug("Found standalone ManualApprovalGate CR from previous version, adopting under TektonConfig")
+			magEnabled = true
+		}
+	}
+	if magEnabled {
```
jkhelil: No test was added for the new tc.Spec.ManualApproval.Options.validate("spec.manualApproval.options") line in tektonconfig_validation.go, even though sibling fields (trigger.options, result.options, multiclusterProxyAAE.options) presumably have similar coverage patterns nearby. Consider adding a case.

pkg/reconciler/shared/tektonconfig/tektonconfig.go
jkhelil: Step 1 — Fresh upgrade (day 0) A cluster already has a standalone ManualApprovalGate CR (no owner) from before this feature existed. The TektonConfig CR is old too, so its stored spec has no manualApproval field at all → tc.Spec.ManualApproval.Disabled is nil (Go zero value, since the field never existed in that stored object).

Code says:

if tc.Spec.ManualApproval.Disabled == nil || len(existingMAG.OwnerReferences) == 0 {
    magEnabled = true   // adopt it
}
Disabled == nil is true → it adopts the MAG CR (sets ownerRef on it). Good, this works as intended.

Step 2 — Someone edits TektonConfig later (day 30) Say an admin changes an unrelated field (e.g. targetNamespace, or a GitOps tool just re-applies the same YAML). That triggers a normal Kubernetes Update call, which goes through the mutating webhook. The webhook runs SetDefaults(), and SetDefaults always fills in Disabled: true if it's nil. So now the stored TektonConfig spec permanently has manualApproval.disabled: true written into it — even though nobody touched that field on purpose.

Step 3 — Next reconcile (day 30, same moment) Now check the same condition again:

tc.Spec.ManualApproval.Disabled == nil → false (it's true now, not nil anymore)
len(existingMAG.OwnerReferences) == 0 → false (we already set the ownerRef back in Step 1)
Both are false, so magEnabled stays false. The reconciler thinks "user wants MAG disabled" and calls EnsureManualApprovalGateCRNotExists, which deletes the MAG CR — even though it was already adopted and working fine, and nobody explicitly asked to disable it.



The fix, simply: stop using Disabled == nil as the signal. Once a MAG CR already has the TektonConfig's ownerRef on it, that's proof adoption already happened — the code should just trust the current explicit Disabled value from then on (and never re-trigger deletion just because defaulting ran). Only use "no ownerRef" as the one-time trigger for adopting a legacy CR.
May be a fix would be 
magEnabled := !tc.Spec.ManualApproval.IsDisabled()
if !magEnabled {
    // If a standalone MAG CR exists with no ownerRef (pre-upgrade legacy CR),
    // adopt it now. Once adopted (ownerRef set), the user's explicit Disabled
    // value takes over — we never force magEnabled=true again after that.
    existingMAG, err := manualapprovalgate.GetManualApprovalGate(ctx, ...)
    if err == nil && len(existingMAG.OwnerReferences) == 0 {
        logger.Debug("Found standalone ManualApprovalGate CR from previous version, adopting under TektonConfig")
        magEnabled = true
    }
}

pkg/reconciler/shared/tektonconfig/tektonconfig.go
```diff
@@ -326,6 +331,49 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tc *v1alpha1.TektonConfi
 		logger.Debug("TektonChain CR removal reconciled successfully")
 	}
 
+	// Ensure ManualApprovalGate CR
+	// If a standalone MAG CR exists (no ownerRef, from a previous version),
+	// adopt its config into TektonConfig and enable it.
+	if tc.Spec.ManualApproval.IsDisabled() {
```
jkhelil, line 338: Transient GetManualApprovalGate error silently falls through to the MAG deletion branch, potentially destroying a standalone pre-upgrade MAG CR before it has been adopted.
The upgrade scenario to trace:

- Cluster has a standalone MAG CR (no ownerRef) from a pre-integration version.
- Operator is upgraded; preUpgradeManualApprovalGate runs but the MAG API temporarily returns a 503 (or the RBAC permission for get on manualapprovalgates is slow to propagate).
GetManualApprovalGate returns a non-NotFound error → err == nil is false → adoption block does NOT return early.
- Code falls through to if !tc.Spec.ManualApproval.IsDisabled() → false → else → - EnsureManualApprovalGateCRNotExists → deletes the standalone MAG CR.
The user's pre-existing MAG installation is destroyed without ever being adopted

Fix 
if tc.Spec.ManualApproval.IsDisabled() {
    existingMAG, err := manualapprovalgate.GetManualApprovalGate(ctx, ...)
    if err != nil {
        if !apierrs.IsNotFound(err) {
            // Transient error — can't determine whether a standalone MAG
            // exists, so don't proceed to the delete branch.
            logger.Warnw("Transient error checking for standalone ManualApprovalGate CR, will retry", "error", err)
            return v1alpha1.REQUEUE_EVENT_AFTER
        }
        // NotFound — no standalone MAG, fall through to delete branch (correct)
    } else if len(existingMAG.OwnerReferences) == 0 {
        // Adopt...
        return v1alpha1.REQUEUE_EVENT_AFTER
    }
}

jkhelil:
@pratap0007 There is no E2E test covering the adoption scenario at all. The ideal deliverable from the author is a new sub-test in test/e2e/common/06_manualapprovalgatedeployment_test.go (or a new 06_mag_via_tektonconfig_test.go) that:

- Creates a standalone MAG CR (no ownerRef, simulating pre-upgrade state).
- Triggers a TektonConfig reconcile.
- Asserts TektonConfig.Spec.ManualApproval.Disabled == false (adoption happened).
- Asserts ManualApprovalGate.OwnerReferences contains TektonConfig.
- Asserts MAG reaches Ready.
As part of upgrade tests

pkg/reconciler/shared/tektonconfig/tektonconfig.go
pratap0007: Thank you for reviewing and identifying these edge cases. I will update with the above fix.

pratap0007:
I have tested both the upgrade scenario and fresh installation, and everything is working as expected.
Thank you, @jkhelil, for your review and suggestions!
@jkhelil could you please take another look?

jkhelil:
@anithapriyanatarajan it is lgtm for me, can you have a quick look

PR #3808, opened 2026-07-24
feat(tektonresult): add NetworkPolicy support for Results

adchauha:
Reconcile default NetworkPolicies for Results API, watcher, retention-policy-agent, and postgres via InstallerSet, propagated from TektonConfig.spec.networkPolicy (SRVKP-12055).

# Changes

Adds NetworkPolicy support for Tekton Results, following the existing Pipeline/Triggers pattern.
- Add `spec.networkPolicy` on `TektonResult` (CRD + validation) and propagate `TektonConfig.spec.networkPolicy` → `TektonResult`
- Reconcile default policies via Custom InstallerSet `results-network-policies`:
  - `results-default-deny`
  - `results-api` (ingress 8080 / Prometheus 9090; egress DNS, Postgres 5432, API server)
  - `results-watcher` (Prometheus ingress; allow-all egress for API server)
  - `results-retention-policy-agent` (egress DNS + Postgres; no ingress)
  - `results-postgres` (ingress 5432 from API + retention; egress DNS)
- Disable via `TektonConfig.spec.networkPolicy.disabled: true` removes the managed policies
- Document Results policies in `docs/NetworkPolicy.md`

# Submitter Checklist

review them:

# Release Notes

```release-note
Add NetworkPolicy support for Tekton Results. Default policies for Results API, watcher, retention-policy-agent, and postgres are reconciled from TektonConfig.spec.networkPolicy (enabled by default; set disabled: true to remove them).
```

### DIFF
--- a/config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
+++ b/config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
@@ -531,8 +531,6 @@ spec:
                           type: string
                       type: object
                     type: array
-                required:
-                - options
                 type: object
               multiclusterProxyAAE:
                 description: MulticlusterProxyAAE holds the customizable options for
--- a/config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
+++ b/config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
@@ -531,6 +531,8 @@ spec:
                           type: string
                       type: object
                     type: array
+                required:
+                - options
                 type: object
               multiclusterProxyAAE:
                 description: MulticlusterProxyAAE holds the customizable options for
@@ -574,8 +576,8 @@ spec:
                 description: |-
                   NetworkPolicy configures NetworkPolicy resources for the operand namespace.
                   This field is propagated to TektonTrigger, TektonPipeline, TektonChain,
-                  and TektonPruner, which are the components with NetworkPolicy reconciliation
-                  implemented. Other components (Results, Dashboard) do not yet act on this field.
+                  TektonPruner, and TektonResult, which implement NetworkPolicy reconciliation.
+                  Other components (Dashboard) do not yet act on this field.
                 properties:
                   disabled:
                     description: |-
--- a/docs/NetworkPolicy.md
+++ b/docs/NetworkPolicy.md
@@ -188,7 +188,7 @@ NetworkPolicy is enabled.
 **Recommended:** add a dedicated temporary NetworkPolicy that allows a dedicated
 admin label. Do **not** reuse `app: tekton-results-api` on a debug pod — that label
 is also used by the Results API Service selector and can route API traffic to the
-wrong pod.
+wrong pod. A dedicated label (below) avoids that.
 
 Apply directly:
 
--- a/docs/NetworkPolicy.md
+++ b/docs/NetworkPolicy.md
@@ -8,8 +8,8 @@ weight: 15
 
 The operator can manage [NetworkPolicy][np] resources for Tekton component workloads.
 Currently TektonPipeline (core controllers, resolvers, and proxy-webhook),
-TektonTrigger, TektonChain, Pipelines-as-Code, ManualApprovalGate, and TektonPruner
-are supported; other components will be added later.
+TektonTrigger, TektonChain, Pipelines-as-Code, ManualApprovalGate, TektonPruner,
+and TektonResult are supported; other components will be added later.
 
 Configuration is available via `TektonConfig`:
 
@@ -33,8 +33,11 @@ spec:
 ```
 
 The `networkPolicy` field is propagated from `TektonConfig` to `TektonTrigger`,
-`TektonPipeline`, `TektonChain`, and `TektonPruner`. Users can also configure it
-directly on the individual component CRs.
+`TektonPipeline`, `TektonChain`, `TektonPruner`, and `TektonResult`. When those
+component CRs are managed by `TektonConfig` (the usual install path),
+**TektonConfig is the source of truth**: edits to `spec.networkPolicy` on the
+component CRs alone are overwritten on the next Config reconcile. Configure
+NetworkPolicy via `TektonConfig.spec.networkPolicy`.
 
 ## Default Policies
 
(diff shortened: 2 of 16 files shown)


docs/NetworkPolicy.md
```diff
@@ -85,6 +87,27 @@ to the operand namespace (e.g. `tekton-pipelines` or `openshift-pipelines`):
...
+| Policy | Direction | Port | Source / Destination |
+|---|---|---|---|
+| `results-default-deny` | deny all | — | Pods with `app.kubernetes.io/name` in Results API, watcher, retention-policy-agent, postgres |
+| `results-api` | ingress | TCP/8080 | All namespaces (Console Plugin, CLI, watcher, internal clients) |
+| | ingress | TCP/9090 | Prometheus namespace |
+| | egress | UDP+TCP/53 (K8s) or 5353 (OpenShift) | DNS resolver pods |
+| | egress | TCP/5432 | Postgres pods (`app.kubernetes.io/name: tekton-results-postgres`) |
+| | egress | all | API server (auth token review / impersonation) |
+| `results-watcher` | ingress | TCP/9090 | Prometheus namespace |
+| | egress | all | API server (all egress allowed — NP cannot select host-network endpoints; watcher also reaches Results API / DNS) |
```
khrm: While it's not required to have DNS ingress, it makes sense to keep a separate DNSIngress also for the watcher and results-retention-policy-agent, so that if we change the filter in DNS, it won't break.

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/tektoncd/operator/pkg/reconciler/common/networkpolicy"
+	"github.com/tektoncd/operator/pkg/reconciler/kubernetes/tektoninstallerset/client"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func resultsDefaultPolicies(params networkpolicy.PlatformParams) []networkingv1.NetworkPolicy {
+	apiPort := intstr.FromInt32(8080)
```
khrm: I think we should take this from the spec. If not available, then use default.

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
```diff
@@ -0,0 +1,209 @@
...
+	metricsPort := intstr.FromInt32(9090)
+	postgresPort := intstr.FromInt32(5432)
```
khrm: Also, here we should take from the spec. If not, then default.

docs/NetworkPolicy.md
khrm: Sorry. I meant DNS egress, not DNS ingress.

PS: I was reviewing it quite late and didn't check what I wrote.

adchauha: Added explicit DNSEgressRule on results-watcher (alongside allow-all for the API server).
results-retention-policy-agent already had DNS egress

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
adchauha: Done. results-api / metrics ports now come from Spec (server_port / prometheus_port) when set, otherwise defaults- 8080 / 9090 via portOrDefault.

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
adchauha: Done. DB port comes from Spec db_port when set, otherwise default 5432.

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
```diff
@@ -0,0 +1,225 @@
...
+				PodSelector: metav1.LabelSelector{
+					MatchLabels: map[string]string{"app": "tekton-results-api"},
+				},
+				PolicyTypes: []networkingv1.PolicyType{
+					networkingv1.PolicyTypeIngress,
+					networkingv1.PolicyTypeEgress,
+				},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						// Console Plugin, CLI, watcher, and other internal clients
```
enarha:
```suggestion
						// Console Plugin, CLI, watcher, routes and other internal clients
```

docs/NetworkPolicy.md
```diff
@@ -85,6 +87,28 @@ to the operand namespace (e.g. `tekton-pipelines` or `openshift-pipelines`):
...
+| | egress | UDP+TCP/53 (K8s) or 5353 (OpenShift) | DNS resolver pods |
+| | egress | all | API server (all egress allowed — NP cannot select host-network endpoints; also covers Results API) |
+| `results-retention-policy-agent` | egress | UDP+TCP/53 or 5353 | DNS resolver pods |
+| | egress | TCP/5432 | Postgres pods |
+| `results-postgres` | ingress | TCP/5432 | Results API and retention-policy-agent pods |
+| | egress | UDP+TCP/53 or 5353 | DNS resolver pods |
+
+For an external database, override `results-api` / `results-retention-policy-agent`
+egress via `spec.networkPolicy.policies` (CIDR or peer) so traffic is not limited
+to in-cluster postgres pods.
```
enarha: I understand that, but external user won't. It's bad user experience to expect users to write network policy as part of user configuration which should be as simple as `db_host: <postgres_instance>\ndb_port: <port>`. That especially given that we only support and recommend external DB for production environments, so most users will have to do it.
Furthermore `spec.networkPolicy.policies` applies to all components, not just Results which is far from ideal. This is not strictly related to this PR, but it worth escalating and fixing in a generic way for all components.

docs/NetworkPolicy.md
```diff
@@ -85,6 +87,28 @@ to the operand namespace (e.g. `tekton-pipelines` or `openshift-pipelines`):
...
+
```
enarha, line 245: No documented path for ad hoc/administrative Postgres access (migration jobs, psql debug pods). Some override mechanism should probably be documented for one-off admin workloads (DB migrations, psql debugging) — right now anything that isn't tekton-results-api or tekton-results-retention-policy-agent gets silently dropped by results-postgres's ingress allowlist once NetworkPolicy is on by default. Worth a line here pointing at `spec.networkPolicy.policies` or a temporary `disabled: true`.

config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
```diff
@@ -531,6 +531,8 @@ spec:
                           type: string
                       type: object
                     type: array
+                required:
+                - options
```
enarha: Where is this coming from? It changes hub and it's not related to results.

enarha (commented):
Can we get an e2e test here similar to test/e2e/common/03_tektontrigger_networkpolicy_test.go — verify Results still functions under the enforced policies and that toggling spec.networkPolicy.disabled adds/removes them?"

config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
enarha: I see the same thing in your other PR: https://github.com/tektoncd/operator/pull/3709/changes#diff-8509813ec32ee79b7ee4581a067b9db639fd10ff1fa862df58444182309424ccR534-R535

enarha: https://github.com/tektoncd/operator/pull/3842 seems related. We might need to merge it first, then you can rebase and regenerate the manifests.

docs/NetworkPolicy.md
khrm: We don't need to write the NP for the external DB. Just use the port that is being given and then write the egress policy on that port. K8s NP isn't URL-based. So traffic to every such port would be allowed.

enarha: But the PR does not have that currently. If you suggest that by default we add a NP allowing egress connections to the default SQL port, that's fine with me. IMO it's even better if we can take the values from the configuration from it. It'll also require constant monitoring of the configuration, so the NP policy is updated when the DB host and port change. The existing state where the user must write NP only to allow connections to external DB instance is IMO unacceptable.

docs/NetworkPolicy.md
khrm: DB migrations or PSQL debugging can be done by running a pod with the same label as the API server instead of creating an NP. 
Or we can have a dedicated NP for debugging preinstalled.

enarha: Using a label could confuse the api Service which also uses label selector to match the api pod(s). If anything, maybe a label to match the retention policy agent. Or a dedicated NP for that is also fine. My point is whatever we decide to use, to document it.

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
adchauha: Done

docs/NetworkPolicy.md
adchauha: Agreed. API and retention DB egress is now port-only using Spec db_port (default 5432), with no in-cluster postgres peer restriction, so external DBs work without a custom NetworkPolicy. Docs updated accordingly. In-cluster postgres remains protected by results-postgres ingress.

docs/NetworkPolicy.md
adchauha: Documented in docs/NetworkPolicy.md- use a dedicated temporary NetworkPolicy (or the same via spec.networkPolicy.policies) allowing pods labeled app: tekton-results-db-admin. We intentionally do not reuse the Results API labels, to avoid confusing the API Service selector. Also documented temporary networkPolicy.disabled: true as an emergency option.

config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
adchauha: Rebased onto main after #3842 and regenerated

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
```diff
@@ -0,0 +1,217 @@
...
+					networkpolicy.DNSEgressRule(params),
+				},
+			},
+		},
+	}
+}
+
+// portOrDefault returns *p as int32 when set and positive, otherwise def.
+func portOrDefault(p *int64, def int32) int32 {
+	if p != nil && *p > 0 && *p <= math.MaxInt32 {
```
jkhelil: portOrDefault accepts any value up to math.MaxInt32, which includes port numbers that Kubernetes rejects (valid port range is 1–65535). A user setting db_port: 99999 silently produces an unappliable NetworkPolicy.

pkg/reconciler/kubernetes/tektonresult/tektonresult.go
```diff
@@ -408,6 +416,16 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, tr *v1alpha1.TektonResul
 	tr.Status.MarkInstallerSetReady()
 	logger.Infow("Installer set is ready", "name", installedTIS.Name)
 
+	if err := r.reconcileNetworkPolicies(ctx, tr); err != nil {
+		if err == v1alpha1.REQUEUE_EVENT_AFTER {
```
jkhelil: reconcileNetworkPolicies doesnt retrun REQUEUE_EVENT_AFTER, this is deead code

test/e2e/common/03_tektonresult_networkpolicy_test.go
```diff
@@ -0,0 +1,155 @@
...
+// reach Postgres; watcher can reach the API server), and that toggling
+// spec.networkPolicy.disabled correctly adds and removes the policies.
+func TestTektonResultNetworkPolicy(t *testing.T) {
+
+	crNames := utils.GetResourceNames()
+	clients := client.Setup(t, crNames.TargetNamespace)
+
+	utils.CleanupOnInterrupt(func() { utils.TearDownPipeline(clients, crNames.TektonPipeline) })
+	utils.CleanupOnInterrupt(func() { utils.TearDownResult(clients, crNames.TektonResult) })
+	defer utils.TearDownResult(clients, crNames.TektonResult)
```
khrm: Can you change this order? Run this after TearDownPipeline.  Otherwise, it might leave TektonResult in some broken state which isn't deleted.

pkg/reconciler/kubernetes/tektonresult/networkpolicies_test.go
```diff
@@ -0,0 +1,194 @@
...
+				return
+			}
+		}
+	}
+	t.Errorf("%s: expected port-only egress TCP/%d (no To peer)", policy, port)
+}
+
+func assertEgressHasDNS(t *testing.T, policy string, rules []networkingv1.NetworkPolicyEgressRule) {
+	t.Helper()
+	for _, rule := range rules {
```
jkhelil, line 177: The rule doesnt verify dns port

pkg/reconciler/kubernetes/tektonresult/networkpolicies.go
adchauha: Fixed. portOrDefault now only accepts ports in the Kubernetes range 1–65535. Values like 99999 or 0 fall back to the default so we don’t emit an unappliable NetworkPolicy. Added unit tests for those cases.

pkg/reconciler/kubernetes/tektonresult/tektonresult.go
adchauha: Removed. reconcileNetworkPolicies never returns REQUEUE_EVENT_AFTER.

pkg/reconciler/kubernetes/tektonresult/networkpolicies_test.go
adchauha: Updated assertEgressHasDNS to require a DNS egress rule with both UDP and TCP on the platform DNS port (5353 for the OpenShift defaults used in this test), matching DNSEgressRule.

test/e2e/common/03_tektonresult_networkpolicy_test.go
adchauha: Thanks for the suggestion. I tried adjusting the teardown order first, but that alone did not fix the neighbour failure.
Root cause was the probe TaskRun.Results adds finalizer results.tekton.dev/taskrun. If Result is deleted while that TaskRun is still present, the finalizer never clears, tekton-pipelines stays terminating, and TestTektonTriggerNetworkPolicy fails.

Kept Result-before-Pipeline  and additionally delete the probe TaskRun, waiting until it is fully gone while Results is still running so the watcher can clear the finalizer.

PR #3810, opened 2026-07-27
fix(tektonchain): sync leader-election configmap

### LINKED ISSUE #2884: Tekton Chains ConfigMap tekton-chains-config-leader-election not Updated with TektonConfig Performance Values
# Expected Behavior
Performance values set in TektonConfig in chains spec should be reflected both in:

- Chains Controller pod labels and args
-  corresponding ConfigMap- tekton-chains-config-leader-election should be kept in sync with current values

# Actual Behavior
- Controller pod has performance labels and args correctly set as per TektonConfig.
- ConfigMap is not updated with the current TektonConfig values.

1. Launch kind cluster
2. Install Tekton operator -` kubectl apply -f https://storage.googleapis.com/tekton-releases/operator/latest/release.yaml`
3. Edit config: `kubectl edit tektonconfig config`
4. Set the following performance attributes
 ` 
 chain:
    artifacts.oci.format: simplesigning
    artifacts.oci.storage: oci
    artifacts.pipelinerun.format: in-toto
    artifacts.pipelinerun.storage: oci
    artifacts.taskrun.format: in-toto
    artifacts.taskrun.storage: oci
    disabled: false
    options: {}
    performance:
      buckets: 3
      disable-ha: true
      kube-api-burst: 20
      kube-api-qps: 10
      replicas: 3
      statefulset-ordinals: false
      threads-per-controller: 4
`
5. Verify that the number of chains controller pods reflects the replica count set in config and the labels + args of each pod include the correct config values.
6. However the configmap `tekton-chains-config-leader-election` does not reflect the current state of any of the performance configuratios

# Additional Info

- Kubernetes version:
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.31.0

- Tekton Operator version:
  **Output of `kubectl get pods -n tekton-operator -l app=tekton-operator   -o=jsonpath='{.items[0].spec.containers[0].env[?(@.name=="VERSION")].value}'**
v0.77.0

pujitha24:
Motivation:
TektonConfig.spec.chain.performance values (e.g. buckets) were
correctly propagated to the tekton-chains-controller Deployment's
pod labels and container args, but the ConfigMap
tekton-chains-config-leader-election was never updated to match,
so it kept stale/default data instead of reflecting the current
TektonConfig. This is a data-consistency bug in that ConfigMap,
not a functional HA failure: the controller itself already
receives the tuned values via the Deployment's args and pod
labels, which are unaffected by this fix.

Approach:
tektonpipeline and tektonresult already call
common.AddConfigMapValues(<leaderElectionConfigMapName>, ...
Performance.PerformanceLeaderElectionConfig) alongside
common.UpdatePerformanceFlagsInDeploymentAndLeaderConfigMap to
keep their leader-election ConfigMaps in sync. The tektonchain
transformer chain was missing that call. Add it, mirroring the
existing pipeline/result behavior.

Validation:
Added TestUpdateLeaderElectionConfigMapForChains to
pkg/reconciler/kubernetes/tektonchain/transform_test.go, with a
new minimal fixture testdata/test-chain-leader-election-configmap.yaml.
Confirmed (by temporarily reverting transform.go) that this test
fails without the fix and passes with it.

  go build ./...
  go test ./pkg/reconciler/kubernetes/tektonchain/... ./pkg/reconciler/common/...

Both commands pass. golangci-lint could not be downloaded in this
sandbox (checksum mismatch fetching the pinned release binary,
unrelated to this change); go vet and go build are clean.

Fixes #2884

```release-note
Fix `tekton-chains-config-leader-election` ConfigMap not being
updated with `TektonConfig.spec.chain.performance` values.
```

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

---

### DIFF
--- a/pkg/reconciler/common/testdata/test-chain-leader-election-configmap.yaml
+++ b/pkg/reconciler/common/testdata/test-chain-leader-election-configmap.yaml
@@ -1,0 +1,43 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: tekton-chains-controller
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/name: tekton-chains
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-chains
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: tekton-chains
+      app.kubernetes.io/component: controller
+      app.kubernetes.io/instance: default
+      app.kubernetes.io/part-of: tekton-chains
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: tekton-chains
+        app.kubernetes.io/component: controller
+        app.kubernetes.io/instance: default
+        app.kubernetes.io/part-of: tekton-chains
+    spec:
+      serviceAccountName: tekton-chains-controller
+      containers:
+      - name: tekton-chains-controller
+        image: gcr.io/tekton-releases/github.com/tektoncd/chains/cmd/controller:latest
+        args:
+        - -disable-ha=true
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: tekton-chains-config-leader-election
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-chains
+data:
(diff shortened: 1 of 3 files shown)


jkhelil:
hi @pujitha24 Can you go through easycla ?

khrm (commented):
/lgtm

PR #3828, opened 2026-07-28
feat(openshiftpipelinesascode): add NetworkPolicy support

zakisk:
Add default-deny and per-pod NetworkPolicies for the three OpenShiftPipelinesAsCode workloads (controller, watcher, webhook), following the pattern established by TektonPipeline and TektonTrigger.

Default policies:
- pac-default-deny: deny all traffic to pods with app.kubernetes.io/part-of=pipelines-as-code
- pac-controller: allow Prometheus metrics ingress (TCP/9090), Git provider webhook ingress (TCP/8082), DNS egress, API server egress, and internet egress (TCP/80,443)
- pac-watcher: allow Prometheus metrics ingress (TCP/9090), DNS egress, API server egress, and internet egress (TCP/80,443)
- pac-webhook: allow admission webhook ingress (TCP/8443), Prometheus metrics ingress (TCP/9090), DNS egress, and API server egress

Policies are reconciled via InstallerSet.CustomSet and respect the spec.networkPolicy.disabled toggle. The NetworkPolicy field is propagated from TektonConfig to the OpenShiftPipelinesAsCode CR.

Includes unit tests for NetworkPolicyConfig validation, an E2E test for policy creation/disable/re-enable, and updated NetworkPolicy documentation.

# Changes

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Run `make test lint` before submitting a PR
- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/operator/blob/master/CONTRIBUTING.md) for more details._

# Release Notes

```release-note
NONE
```

PR #3842, opened 2026-07-30
fix(hack): stop sync-helm-crds.sh reviving dead manual CRDs

jkhelil:
# Changes

## 1. `hack/sync-helm-crds.sh` still regenerated dead manual CRD copies

`hack/sync-helm-crds.sh` still regenerated the manual per-component CRD
copies under `config/base/`, `config/kubernetes/base/`, and
`config/openshift/base/` (`300-operator_v1alpha1_<component>_crd.yaml`).

Commit 1ae0906b3 ("Cleanup manual CRDs and use generated CRDs in
kustomize") deleted those files and pointed every `kustomization.yaml`
at `config/base/generated-crds/*.yaml` directly, but never updated
this script to match. As a result, running `make sync-helm-crds` (a
documented prerequisite for any CRD-affecting API change) silently
resurrects ~4,700 lines of unreferenced, dead manifests on every use
— nothing in the repo consumes them (verified via `git grep` across
the tree; the only reference was the script itself).

This surfaced concretely in
https://github.com/tektoncd/operator/pull/3828, where an unrelated
`OpenShiftPipelinesAsCode` NetworkPolicy change ballooned into a
~8,900-line diff because the author correctly ran `make
sync-helm-crds` and the stale script did the rest.

This PR drops the dead `write_config_crd` step (and the now-unused
`BASE_DIR`/`K8S_DIR`/`OPENSHIFT_DIR` variables) so the script only
does what's still needed: assembling the Helm chart CRD bundles
(`charts/tekton-operator/templates/{kubernetes,openshift}-crds.yaml`)
from `config/base/generated-crds/`, which remains kustomize's sole
source of truth.

## 2. `TektonConfig.spec.hub.options` was implicitly required

While investigating #1, I found that running `make generate-crds` on
unmodified `main` (independent of any other PR) produces a
`required: [options]` entry under `spec.hub` in
`config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml` —
the file kustomize applies directly to clusters. That would reject
any `TektonConfig` CR that doesn't explicitly set `spec.hub.options`,
a breaking regression for existing users on upgrade.

Root cause: `Hub.Options` in `tektonconfig_types.go` was the only
`Options AdditionalOptions` field across all components missing the
`// +optional` marker (every other component — Pipeline, Trigger,
Chain, Dashboard, Addon, ManualApprovalGate, Pruner, Scheduler,
MulticlusterProxyAAE, SyncerService — has it). Almost certainly an
oversight from the earlier "fix: options field should be optional in
all components" work.

Fixed by adding `+optional`/`omitempty` to match every other
component, then regenerating `operator.tekton.dev_tektonconfigs.yaml`
and `charts/tekton-operator/templates/kubernetes-crds.yaml` via
`make sync-helm-crds`. The regeneration also picks up an unrelated,
already-correct doc fix: the `networkPolicy` field description now
mentions `TektonPipeline` alongside `TektonTrigger`.

`charts/tekton-operator/templates/openshift-crds.yaml` is
intentionally left untouched: it has independently drifted to bundle
only 3 of the 13 CRDs it should, and re-syncing it pulls in a large,
unrelated diff that deserves its own review/PR.

## Verification

- `bash -n hack/sync-helm-crds.sh` — syntax OK
- `go build ./pkg/...` and `go vet ./pkg/apis/operator/v1alpha1/...` — clean
- `go test ./pkg/apis/operator/v1alpha1/...` — no new failures (3
  pre-existing failures in `TestSetPACControllerDefaultSettings*`
  remain, caused by an unrelated `pipelines-as-code` v0.48.0→v0.48.1
  default `hub-catalog-type` change, already broken on unmodified
  `main`)
- Ran the fixed script against `main`: no `config/*/base/*_crd.yaml`
  files are recreated, and (before the hub.options fix)
  `kubernetes-crds.yaml` regenerated byte-for-byte identical to
  what's committed, confirming it was already in sync

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/charts/tekton-operator/templates/kubernetes-crds.yaml
+++ b/charts/tekton-operator/templates/kubernetes-crds.yaml
@@ -1446,9 +1446,9 @@ spec:
               networkPolicy:
                 description: |-
                   NetworkPolicy configures NetworkPolicy resources for the operand namespace.
-                  This field is propagated to TektonTrigger, which is the only component with
-                  NetworkPolicy reconciliation implemented. Other components (Pipeline, Chains,
-                  Results, Dashboard) do not yet act on this field.
+                  This field is propagated to TektonTrigger and TektonPipeline, which are the
+                  only components with NetworkPolicy reconciliation implemented. Other
+                  components (Chains, Results, Dashboard) do not yet act on this field.
                 properties:
                   disabled:
                     description: |-
--- a/config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
+++ b/config/base/generated-crds/operator.tekton.dev_tektonconfigs.yaml
@@ -573,9 +573,9 @@ spec:
               networkPolicy:
                 description: |-
                   NetworkPolicy configures NetworkPolicy resources for the operand namespace.
-                  This field is propagated to TektonTrigger, which is the only component with
-                  NetworkPolicy reconciliation implemented. Other components (Pipeline, Chains,
-                  Results, Dashboard) do not yet act on this field.
+                  This field is propagated to TektonTrigger and TektonPipeline, which are the
+                  only components with NetworkPolicy reconciliation implemented. Other
+                  components (Chains, Results, Dashboard) do not yet act on this field.
                 properties:
                   disabled:
                     description: |-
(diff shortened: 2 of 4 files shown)


enarha (commented):
/approve

PR #3848, opened 2026-08-03
feat(openshift): secure component metrics endpoints with mutual TLS

jkhelil:
# Changes

Secure component metrics endpoints with mutual TLS (mTLS) on OpenShift, gated
behind an opt-in feature flag.

Each reconciler syncs the Prometheus client CA bundle from
`kube-system/extension-apiserver-authentication` into the component namespace
as a `metrics-client-ca` ConfigMap. OpenShift's serving-cert controller is
triggered via annotation on each metrics `Service` to provision a
per-component TLS `Secret`. Manifest transformers (`AnnotateMetricsServingCert`,
`RenameServicePort`, `ApplyMetricsTLS`) wire the Secret and ConfigMap as
volumes and inject `METRICS_PROMETHEUS_TLS_*` env vars so `knative/pkg`'s
`prometheus.Server` enables mTLS with required client auth.
`ServiceMonitor` resources are updated with `scheme: https`,
`tlsConfig.serverName`, and the correct port name.

The feature is opt-in via `spec.platforms.openshift.enableMetricsMTLS` on
`TektonConfig` (default `false`). When disabled, all mTLS transformers are
skipped and metrics stay on plain HTTP — existing clusters are unaffected.

## Known limitation: `tekton-results-api`

`tekton-results-api`'s Prometheus metrics endpoint
(`cmd/api/main.go` in `tektoncd/results`) is a bare `net/http` server that
never reads `METRICS_PROMETHEUS_TLS_*` env vars, so it cannot serve mTLS.
It is therefore **excluded** from the mTLS upgrade: its `Service` and
`ServiceMonitor` stay on plain HTTP regardless of the flag.
`tekton-results-watcher` is unaffected by this and gets full mTLS like every
other component, since it is a standard Knative-based reconciler. Fixing
`tekton-results-api` itself would require an upstream change in
`tektoncd/results`, tracked separately.

## Testing performed

- Enabled the flag on a live OpenShift cluster and verified, for every
  mTLS-capable component (pipeline, triggers, chains, pruner,
  results-watcher): the `metrics-client-ca` ConfigMap synced, the metrics
  `Service` port renamed to `https-*`, the serving-cert `Secret` was
  provisioned, and the `ServiceMonitor` was updated to `https` + `tlsConfig`.
- Verified mTLS enforcement with `curl`: requests with a valid Prometheus
  client cert succeed (`200`, real metrics returned); requests without a
  client cert fail the TLS handshake (explicit "certificate required" alert).
- Verified `tekton-results-api` metrics remain reachable over plain HTTP
  with the flag enabled (no regression from the mTLS work).
- Disabled the flag and verified a clean rollback to plain HTTP for all
  components, with matching `Service`/`ServiceMonitor` ports.
- Repeated the on/off toggle to confirm idempotency.

# Submitter Checklist

- [x] Includes tests (if functionality changed/added)
- [ ] Includes docs (if user facing)
- [x] Commit messages follow commit message best practices

# Release Notes

```release-note
Add opt-in `spec.platforms.openshift.enableMetricsMTLS` field to
`TektonConfig` to secure component Prometheus metrics endpoints with
mutual TLS on OpenShift. Disabled by default; existing clusters are
unaffected. Not supported for `tekton-results-api`, whose metrics
endpoint stays on plain HTTP.
```

### DIFF
--- a/pkg/reconciler/openshift/common/metricsca.go
+++ b/pkg/reconciler/openshift/common/metricsca.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2026 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/pkg/logging"
+)
+
+const (
+	// MetricsClientCAConfigMap is the name of the ConfigMap synced into each
+	// component namespace so that the Prometheus metrics server can verify
+	// the client certificate presented by Prometheus during mTLS scraping.
+	MetricsClientCAConfigMap = "metrics-client-ca"
+
+	// MetricsClientCAKey is the data key within MetricsClientCAConfigMap that
+	// holds the PEM-encoded CA bundle.
+	MetricsClientCAKey = "client-ca-file"
+
+	// SourceConfigMapName is the authoritative ConfigMap in kube-system that
+	// contains the Prometheus client CA bundle on OpenShift.
+	//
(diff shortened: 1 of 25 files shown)


pkg/reconciler/openshift/common/metricsca.go
```diff
@@ -0,0 +1,164 @@
...
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("reading TektonConfig: %w", err)
+	}
+	if tc.Spec.Platforms.OpenShift.EnableMetricsMTLS == nil || !*tc.Spec.Platforms.OpenShift.EnableMetricsMTLS {
+		return false, nil
+	}
+
+	_, err = kubeClient.CoreV1().ConfigMaps(targetNamespace).Get(
```
pratap0007, line 74: Why are we ResolvingMTLS based on the ConfigMap's availability instead of using tc.Spec.Platforms.OpenShift.EnableMetricsMTLS?

PR #3873, opened 2026-08-06
revert(operator): remove NetworkPolicy from operator bundle manifests

jkhelil:
## Changes

Removes the static NetworkPolicy manifests added to the operator bundle by #3714:

- `config/kubernetes/base/networkpolicy.yaml` (deleted)
- `config/openshift/base/networkpolicy.yaml` (deleted)
- `config/kubernetes/base/kustomization.yaml` (entry removed)
- `config/openshift/base/kustomization.yaml` (entry removed)
- `docs/NetworkPolicy.md` (operator namespace section removed)

## Why

OLM v1 does not support `NetworkPolicy` resources in operator bundles. Shipping these as static manifests alongside the operator Deployment/RBAC causes install failures on OLM v1 clusters.

The reconciler-managed NetworkPolicies for operand namespaces (proxy-webhook, pipeline core, triggers, chains, PAC, pruner, results, multicluster, etc.) are **not** affected — those are applied by the operator's own reconcilers and are unrelated to the bundle manifests.

## Related

Partial revert of #3714 (static-manifest portion only).


khrm (commented):
/lgtm

PR #3884, opened 2026-08-10
fix(tektonresult): default watcher logs_api from spec

### LINKED ISSUE #2600: The TektonResults component cannot store logs information.
# Expected Behavior

After enabling the logs API feature of TektonResults, logs can be stored normally.

# Actual Behavior

Even if the switch is enabled in TektonConfig, logs still cannot be stored.

1. Adjust TektonResults configuration
```yaml
kind: TektonResult
metadata:
  name: result
spec:
  auth_disable: true
  db_enable_auto_migration: true
  db_host: 192.168.188.7
  db_name: qingliu_tektoncd_results_202502171201
  db_port: 5432
  is_external_db: true
  log_level: debug
  logging_pvc_name: tekton-logs
  logs_api: true        # <- enable logs api
  logs_path: /logs
  logs_type: File
  options: {}
  secret_name: tekton-results-postgres
  server_port: 8080
```
2. Restart the pod
```yaml
$ kubectl delete pods -n tekton-pipelines -l app.kubernetes.io/name=tekton-results-watcher

$ kubectl delete pods -n tekton-pipelines -l app.kubernetes.io/name=tekton-results-api
```
3. Create TaskRun
```sh
$ cat <<'EOF' | kubectl create -f -
apiVersion: tekton.dev/v1
kind: TaskRun
metadata:
  name: hello
spec:
  taskSpec:
    steps:
      - name: hello
        image: alpine
        command: ["echo", "hello"]
EOF
```

# Additional Info

- Kubernetes version:

  ```
  Client Version: v1.32.2
  Kustomize Version: v5.5.0
  Server Version: v1.28.13
  WARNING: version difference between client (1.32) and server (1.28) exceeds the supported minor version skew of +/-1
  ```

- Tekton Operator version:
  - v0.75.0

# Reference

- Changed the default startup argument `logs_api ` to false.
  - https://github.com/tektoncd/results/pull/782
- Solutions from others
  - https://github.com/tektoncd/results/pull/910
- The TektonOperator does not modify the startup parameters of the `tekton-results-watcher` based on the logs_api value in TektonResults when deploying TektonResults. This results in the watcher service always having this feature disabled.
  - https://github.com/tektoncd/operator/blob/07fcbf76face59c774f9d32d1eec786673c86131/pkg/reconciler/kubernetes/tektonresult/transform.go#L85-L95

pujitha24:
# Changes

Fixes a bug where setting `spec.logs_api: true` on a `TektonResult` (or via `TektonConfig`) enabled the `LOGS_API` env var on the `tekton-results-api` deployment, but did not propagate to the `tekton-results-watcher` deployment's separate `-logs_api` flag (which defaults to `false`). As a result, users following the documented example of only setting the top-level `logs_api` ended up with an API server that had logging enabled but a Watcher that never forwarded logs, so logs were never stored.

`Result.setDefaults()` now defaults `Result.Watcher.LogsAPI` from the top-level `ResultsAPIProperties.LogsAPI` whenever the user has not explicitly set `spec.watcher.logs_api`, mirroring the existing `RouteEnabled`/`RouteTLSTermination` defaulting in the same function. An explicit `spec.watcher.logs_api` value is never overridden.

Note: this defaulting runs wherever `Result.setDefaults()` runs today (i.e. when a `TektonResult` is created/updated via `TektonConfig`, the documented install path). `TektonResult.SetDefaults` does not call `Result.setDefaults()`, so a CR applied directly as a standalone `kind: TektonResult` does not get this (or the pre-existing `RouteEnabled`) default — that gap predates this change and is not addressed here.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix TektonResult so that setting `spec.logs_api: true` also defaults the tekton-results-watcher's log forwarding flag, ensuring TaskRun/PipelineRun logs are actually stored when the top-level `logs_api` option is enabled.
```

---

Fixes #2600

### DIFF
--- a/pkg/apis/operator/v1alpha1/tektonresult_defaults.go
+++ b/pkg/apis/operator/v1alpha1/tektonresult_defaults.go
@@ -41,6 +41,14 @@ func (c *Result) setDefaults() {
 		c.RouteTLSTermination = "edge"
 	}
 
+	// The Watcher only forwards logs to the API server when its own
+	// logs_api flag is set; enabling the API server's logs_api alone does
+	// not do this. Default the Watcher to match unless the user set
+	// watcher.logs_api explicitly.
+	if c.LogsAPI != nil && c.Watcher.LogsAPI == nil {
+		c.Watcher.LogsAPI = c.LogsAPI
+	}
+
 	c.setPerformanceDefaults()
 }
 
--- a/pkg/apis/operator/v1alpha1/tektonresult_defaults_test.go
+++ b/pkg/apis/operator/v1alpha1/tektonresult_defaults_test.go
@@ -127,6 +127,98 @@ func TestResult_SetDefaultsRoutes(t *testing.T) {
 	}
 }
 
+func TestResult_SetDefaultsWatcherLogsAPI(t *testing.T) {
+	tests := []struct {
+		name string
+		r    *Result
+		want *Result
+	}{
+		{
+			name: "propagates enabled logs_api to watcher when unset",
+			r: &Result{
+				ResultsAPIProperties: ResultsAPIProperties{
+					LogsAPI:      ptr.Bool(true),
+					RouteEnabled: ptr.Bool(true),
+				},
+			},
+			want: &Result{
+				ResultsAPIProperties: ResultsAPIProperties{
+					LogsAPI:             ptr.Bool(true),
+					RouteEnabled:        ptr.Bool(true),
+					RouteTLSTermination: "edge",
+				},
+				Watcher: ResultsWatcherProperties{
+					LogsAPI: ptr.Bool(true),
+				},
+			},
+		},
+		{
+			name: "propagates disabled logs_api to watcher when unset",
+			r: &Result{
+				ResultsAPIProperties: ResultsAPIProperties{
+					LogsAPI:      ptr.Bool(false),
+					RouteEnabled: ptr.Bool(true),
+				},
+			},
+			want: &Result{
+				ResultsAPIProperties: ResultsAPIProperties{
+					LogsAPI:             ptr.Bool(false),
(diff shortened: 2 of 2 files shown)


jkhelil:
@enarha @khrm PTAL

pujitha24:
Tide was flagging this as unmergeable because the branch had a merge commit from syncing upstream/main, which a clean rebase can't replay past a later commit (`39604bf5d`) touching the same function. Rebased onto current `main` and resolved it so history is linear again — no logic changes, the two touched files are byte-identical to what was already passing CI here. Tests and build still pass locally.

pujitha24:
/assign @jkhelil

Rebased and green — let me know if anything else would help move this toward approval.

pratap0007:
@enarha @khrm Could you please take a look?

khrm (commented):
/hold

khrm (commented):
Existing behaviour is correct. Watcher forwarder is not required, you can have a different forwarder like vector.

pujitha24:
Fair point — if logs already land in storage some other way (an external forwarder, or a LokiStack-based viewer), pushing through the Watcher too would be redundant, and leaving watcher.logs_api off by default there is correct. The conflict is with #2600 itself and the PVC example in TektonResult.md: those only set the top-level logs_api with nothing else configured to get logs in, so the feature silently never stores anything. Would you rather I drop this default and instead fix the docs to have users set watcher.logs_api explicitly, so we don't risk flipping behavior on upgrade for anyone currently relying on the existing default (e.g. LokiStack users)?

khrm (commented):
Yes, maybe update docs. Also, watcher forwarding logs would be removed in future. That point can also be added.

PR #3904, opened 2026-08-11
fix(common): Preserve param substitution in step images during replacement

anchi205:
# Changes

The replaceStepsImages transformer unconditionally overwrites step images with pinned digests from environment variables. This breaks tasks like maven that use  in the image field
to allow users to select a Java version at runtime.

Add a guard to skip image replacement when the existing step image contains Tekton parameter references, so that parameter substitution is preserved and resolved correctly at runtime.

# Submitter Checklist

review them:

- [X] Run `make test lint` before submitting a PR

# Release Notes

```release-note
Fix operator image replacement overwriting parameter substitution in Task step images
```

### DIFF
--- a/pkg/reconciler/common/transformers.go
+++ b/pkg/reconciler/common/transformers.go
@@ -443,6 +443,10 @@ func replaceStepsImages(steps []interface{}, override map[string]string, logger 
 			}
 			continue
 		}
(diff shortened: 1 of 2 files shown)


jkhelil:
@anchi205 Please confirm the PR is finxing the issue with a local test

PR #3940, opened 2026-08-17
fix(chart): stamp CRD labels with chart appVersion instead of hardcoded devel

aqeelat:
# Changes

The OCI-published chart (e.g. `ghcr.io/tektoncd/operator/charts/tekton-operator:0.81.0`) ships CRDs with labels `version: "devel"` and `operator.tekton.dev/release: "devel"` even though the chart itself is correctly stamped (`version: 0.81.0`, `appVersion: v0.81.0`) and all deployment images pin `:v0.81.0`. For GitOps users (ArgoCD), every CRD diff shows `release: devel` while everything else says the real version — misleading for auditing.

Root cause: `hack/sync-helm-crds.sh` (which generates `charts/tekton-operator/templates/{kubernetes,openshift}-crds.yaml`) injects hardcoded `"devel"` labels, and the release workflow (`.github/workflows/helm-release.yaml`) stamps only `Chart.yaml` — the CRD templates are never touched.

Fix: the label injection now emits `{{ .Chart.AppVersion | quote }}`. In-tree/dev renders keep `devel` (Chart.yaml `appVersion: "devel"`), released OCI charts stamp the real version. No workflow change needed.

These labels are inert metadata: the operator reads `operator.tekton.dev/release-version` (a different key) on InstallerSets; nothing selects on these CRD labels.

Render proof (13 CRDs per platform, both kubernetes and openshift sets):

| Chart.yaml appVersion | Rendered CRD label |
|---|---|
| `"devel"` (in-tree) | `operator.tekton.dev/release: "devel"` |
| `"v0.82.0"` (release-stamped) | `operator.tekton.dev/release: "v0.82.0"` |

Note: 0.81.0 is already published and stays as-is; this lands in the next release (maintainers may cherry-pick).

# Submitter Checklist

- [x] Run `make test lint` before submitting a PR (chart-only change; `lint-go` 0 issues, `yamllint` clean on changed files. `helm lint` fails on main pre-existing — `version: "devel"` is not valid semver)
- [ ] Includes tests (chart templates — verified via `helm template` renders above)
- [ ] Includes docs (no user-facing behavior change beyond label values)

# Release Notes

```release-note
Helm chart: CRDs now carry the chart appVersion in their `version` / `operator.tekton.dev/release` labels instead of the hardcoded `devel`, so released charts show the actual release version.
```

### DIFF
--- a/charts/tekton-operator/templates/kubernetes-crds.yaml
+++ b/charts/tekton-operator/templates/kubernetes-crds.yaml
@@ -7,8 +7,8 @@ metadata:
     controller-gen.kubebuilder.io/version: v0.18.0
   name: manualapprovalgates.operator.tekton.dev
   labels:
-    version: "devel"
-    operator.tekton.dev/release: "devel"
+    version: {{ .Chart.AppVersion | quote }}
+    operator.tekton.dev/release: {{ .Chart.AppVersion | quote }}
 spec:
   group: operator.tekton.dev
   names:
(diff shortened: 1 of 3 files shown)


waveywaves (approved):

infernus01 (approved):
/lgtm

PR #3945, opened 2026-08-17
fix(chart): make the openshift flavor functional and align with kustomize

aqeelat:
# Changes

The openshift flavor of the Helm chart (`openshift.enabled=true`) did not work out of the box, and the chart had drifted from the kustomize installation manifests. This PR fixes the openshift flavor and aligns both flavors with `config/kubernetes/base` and `config/openshift/base` (verified by diffing `helm template` output against `kubectl kustomize` for both flavors; remaining differences are limited to release-name-parameterized resource names, digest-pinned images instead of `ko://` refs, and inlined ConfigMap defaults).

Commits are separated so anything unneeded can be dropped:

1. **Template the openshift ServiceAccount name** — the openshift RBAC file hardcoded `openshift-pipelines-operator` while the Deployments resolve the SA via the fullname helper, so the pods referenced an SA the chart never created unless the user set `rbac.serviceAccount.name` manually.
2. **Set `PLATFORM=openshift`** on the lifecycle and webhook containers in the openshift flavor, so `IsOpenShiftPlatform()` activates the openshift reconciler branches, defaults and validation.
3. **Add missing RBAC** — pac-controller (kubernetes), result-read and scheduler roles/bindings (both flavors, in a new `common-rbac.yaml`, fullname-prefixed), plus openshift ClusterRole rule alignment (nodes:list, SCC list, config.openshift.io reads, routes/custom-host, events, consoleplugins, pipelinesascode.tekton.dev group, resolutionrequests/status).
4. **Use platform images and controllers per flavor** — the openshift branches of the image helpers now select the openshift operator/webhook/proxy images (digest-pinned from the v0.81.0 openshift release), and `-controllers` gets a flavor-aware default matching the kustomize deployments. Note: the previous kubernetes default was missing `tektonscheduler` and `tektonmulticlusterproxyaae`; the new default adds them (additive change, nothing is removed from vanilla).
5. **Wire leader election and add missing ConfigMaps** — controller/webhook leader-election ConfigMaps + `CONFIG_LEADERELECTION_NAME` on all containers, and the `tekton-operator-info` ConfigMap consumed by `tkn version` (the chart already shipped the Role granting read access to it).
6. **Fix webhook logLevel wiring** — the webhook container was named after the operator name while the logging ConfigMap keys off the fullname, so `webhook.logLevel` never applied.
7. **Reject a custom `operator.defaultTargetNamespace` on the openshift flavor** — the openshift addon sample pipelines hardcode the `openshift-pipelines` namespace in their cluster-resolver taskRefs, so any other value ships a broken config; the chart now fails the render with an explanatory message instead. The kubernetes flavor keeps accepting any namespace. Also documented in values.yaml: the value is **first-install-only** — it seeds the auto-created TektonConfig CR, and once the CR exists the operator never re-reads it (retargeting requires deleting and recreating the CR; `spec.targetNamespace` updates are rejected by validation).

**Decisions worth flagging for reviewers:**
- `IMAGE_ADDONS_*` env vars are deliberately not set: addon image defaults come from the operator's built-in manifests and registries can be rewritten with `TEKTON_REGISTRY_OVERRIDE` or per-image envs via `operator.additionalEnvs` (documented in values.yaml; see AirGapImageConfiguration.md).
- The openshift flavor's ServiceMonitor remains behind `service.createServiceMonitor` (default off) — upstream ships one unconditionally; happy to flip the default if parity is wanted.
- The new roles are not gated on the `controllers` value; happy to follow up if per-controller RBAC gating is preferred.
- The new `result-read-role` grants `system:authenticated` get/watch/list on `tektonresults` CRs — exact parity with upstream `config/base/tekton_result_role_binding.yaml` and the same pattern as the chart's existing `tekton-config-read-role`, but it does widen read access for chart upgrades that previously lacked it.
- **Upgrade note (renamed RBAC):** `tekton-config-read-role`/`-rolebinding` (both flavors) and the openshift `tekton-operator-info` Role/RoleBinding are renamed to fullname-prefixed names. `helm upgrade` (and GitOps renderers like ArgoCD) delete the old-named objects automatically — helm diffs the stored release manifest against the new render and deletes anything removed, so no manual cleanup is needed. Only clusters managed with `helm template | kubectl apply` (no prune) keep stale copies; delete them manually if that is your setup. Anything referencing the old literal names (e.g. custom RoleBindings binding users to `tekton-config-read-role`) must be updated, since those objects disappear on upgrade.
- A render-time guard (4bdaa7e5f) fails the chart when `controllers` contains a name the selected binary doesn't register — this catches the v0.81.1 `--reuse-values` upgrade path, where the old hardcoded kubernetes list (incl. `tektondashboard`) crash-loops the openshift operator (verified live; workaround `--set controllers=""`). The commit is droppable; a fuller allowlist variant (typo detection, `tektoninstallerset` reservation) is available as a follow-up. Related upstream finding filed as #4123.
- The `operator.defaultTargetNamespace` guard exists because the upstream openshift addon reconciler hardcodes `openshift-pipelines` in the generated sample pipelines (`pkg/reconciler/openshift/tektonaddon/pipelinetemplates/pipelinetemplates.go`, `getTaskRef`). Upstream fix: thread the addon CR's `targetNamespace` through `GeneratePipelineTemplates`; once landed, this guard can be relaxed. Relatedly, upstream reads `DEFAULT_TARGET_NAMESPACE` only once when auto-creating the config CR — changing the env afterwards is silently ignored.

# Submitter Checklist

Chart-only change; verified via `helm template` for both flavors and value overrides, and yamllint.

# Release Notes

```release-note
Fix the openshift flavor of the Helm chart: openshift operator/webhook/proxy images and controller list are now selected per flavor, PLATFORM=openshift is set, the operator ServiceAccount follows the chart fullname, missing RBAC roles (pac, result-read, scheduler) and ConfigMaps (leader election, operator-info) are shipped, and webhook.logLevel is wired correctly. The kubernetes controllers default additionally includes tektonscheduler and tektonmulticlusterproxyaae. The openshift flavor now rejects a custom operator.defaultTargetNamespace while the addon sample pipelines hardcode the openshift-pipelines namespace (first-install-only value; documented in values.yaml).
```

---

**AI assistance:** developed with [opencode](https://opencode.ai) (GLM model) under my direction; every commit carries an `Assisted-by:` trailer per the [AI contribution policy](https://github.com/tektoncd/community/blob/main/ai-contribution-policy.md). I drove scoping and decisions, reviewed every change, and verified each push — this PR was validated by diffing `helm template` output against `kubectl kustomize` for both flavors, plus yamllint.


charts/tekton-operator/templates/deployment.yaml
```diff
@@ -48,6 +48,10 @@ spec:
               value: tekton-operator
               {{- end }}
               {{- end }}
+            {{- if .Values.openshift.enabled }}
+            - name: PLATFORM
```
waveywaves, line 53: `PLATFORM` does not switch the compiled operator binary. The OpenShift branch in `_helpers.tpl` still selects the Kubernetes operator image. The correct OpenShift binary also rejects the default `tektondashboard` controller. Please make the image and controller defaults flavor-aware, or narrow the PR title, description, and release note.

charts/tekton-operator/templates/deployment.yaml
```diff
@@ -209,6 +213,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            {{- if .Values.openshift.enabled }}
+            - name: PLATFORM
```
waveywaves, line 210: The OpenShift branches in `_helpers.tpl` also select the Kubernetes webhook and proxy images. Please use the OpenShift image repositories, or narrow the PR scope and release note.

waveywaves (changes requested):

charts/tekton-operator/templates/deployment.yaml
aqeelat: Addressed in 59d667f: the openshift webhook and proxy-webhook image helpers now select the openshift image repositories, digest-pinned from the v0.81.0 openshift release.

charts/tekton-operator/templates/deployment.yaml
aqeelat: Addressed in 59d667f: the openshift branches of the image helpers now select the openshift operator/webhook/proxy images (digest-pinned from the v0.81.0 openshift release), and the `-controllers` flag gets a flavor-aware default matching config/openshift/base/operator.yaml (tektonaddon and syncerservice instead of tektondashboard, plus tektonscheduler and tektonmulticlusterproxyaae). IMAGE_ADDONS_* is deliberately not set — defaults come from the operator's built-in manifests and TEKTON_REGISTRY_OVERRIDE covers registry rewrites; noted in values.yaml and the PR description.

charts/tekton-operator/templates/_helpers.tpl
```diff
@@ -96,17 +96,31 @@ tekton-operator
...
+
 {{- define "tekton-operator.operator-image" -}}
 {{- $tag := default .Chart.AppVersion .Values.operator.image.tag -}}
 {{- $image := "" -}}
 {{- if .Values.operator.image.repository -}}
   {{- $image = .Values.operator.image.repository }}
 {{- else -}}
 {{- if .Values.openshift.enabled -}}
+  {{- $image = "ghcr.io/tektoncd/operator/operator-1d69a75f22dd094880847eac907fb2c1" -}}
+  {{- else -}}
```
jkhelil: ident, remove space for readbility

charts/tekton-operator/templates/_helpers.tpl
```diff
@@ -96,17 +96,31 @@ tekton-operator
...
   {{- $image = "ghcr.io/tektoncd/operator/operator-303303c315a48490ba6517859ef65b77" -}}
-{{- else -}}
-  {{- $image = "ghcr.io/tektoncd/operator/operator-303303c315a48490ba6517859ef65b77" -}}
-{{- end -}}
+  {{- end -}}
```
jkhelil: same identation for readbility

charts/tekton-operator/templates/_helpers.tpl
```diff
@@ -96,17 +96,31 @@ tekton-operator
...
+{{- end -}}
+
 {{- define "tekton-operator.operator-image" -}}
 {{- $tag := default .Chart.AppVersion .Values.operator.image.tag -}}
 {{- $image := "" -}}
 {{- if .Values.operator.image.repository -}}
   {{- $image = .Values.operator.image.repository }}
 {{- else -}}
 {{- if .Values.openshift.enabled -}}
+  {{- $image = "ghcr.io/tektoncd/operator/operator-1d69a75f22dd094880847eac907fb2c1" -}}
```
jkhelil: Add comment for digest pinning 
{{/* OpenShift image repositories are digest-pinned to the v0.81.0 .
     Update all three slugs (operator/webhook/proxy-webhook) when bumping AppVersion. */}}

charts/tekton-operator/templates/common-rbac.yaml
```diff
@@ -0,0 +1,156 @@
...
+      - tektonconfigs
+    verbs:
+      - get
+      - watch
+      - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: tekton-config-read-rolebinding
```
jkhelil: may be not hardcode the name , two simultaneous chart installs in the same cluster will conflict on this ClusterRole and ClusterRoleBinding, use
{{ include "tekton-operator.fullname" . }}

charts/tekton-operator/templates/common-rbac.yaml
```diff
@@ -0,0 +1,156 @@
+{{- if .Values.rbac.create -}}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: tekton-operator-info
```
jkhelil: Orphaned resource risk: kubernetes-rbac.yaml to common-rbac.yaml rename

⚠️ Issue: Before this PR, kubernetes-rbac.yaml shipped:

RoleBinding named {{ include "tekton-operator.fullname" . }}-info
ClusterRole named tekton-config-read-role
ClusterRoleBinding named tekton-config-read-rolebinding
After this PR, the equivalent resources move to common-rbac.yaml with different names (tekton-operator-info for the RoleBinding). Helm upgrade will create the new resources but will not delete the old <fullname>-info RoleBinding because it no longer appears in any template. Users upgrading will accumulate stale RBAC objects.

jkhelil:
@aqeelat 
ince the repo has no Helm e2e tests, could you provide manual verification evidence for the following scenarios? A helm template diff is already a good start, but we need proof the chart is functional end-to-end:

1. Fresh install — Kubernetes

helm install tekton-operator ./charts/tekton-operator -n tekton-operator --create-namespace
kubectl get pods -n tekton-operator
kubectl get tektonconfig config -o jsonpath='{.status.conditions}'
2. Fresh install — OpenShift

helm install tekton-operator ./charts/tekton-operator -n openshift-operators \
  --set openshift.enabled=true
kubectl get pods -n openshift-operators
kubectl get tektonconfig config -o jsonpath='{.status.conditions}'
3. Upgrade from the previous chart version (both flavors)

helm upgrade tekton-operator ./charts/tekton-operator --reuse-values
kubectl get pods -n <namespace>   # confirm no CrashLoopBackOff
4. Confirm webhook.logLevel is now wired

helm upgrade --set webhook.logLevel=debug ...
kubectl exec -n <ns> deploy/tekton-operator-webhook -- env | grep -i log
5. Confirm PLATFORM=openshift activates the OpenShift reconciler

kubectl exec -n openshift-operators deploy/tekton-operator -- env | grep PLATFORM
kubectl get tektonaddons                   # should exist only on OpenShift flavor
Screenshots or kubectl get/describe output for each scenario would be sufficient. Alternatively, a short recording or CI run against a real cluster is also acceptable.

jkhelil:
hey @aqeelat Could you share how you raised this PR ?

charts/tekton-operator/templates/_helpers.tpl
aqeelat: Flattened to match the chart style — fixed in 42b857837.

charts/tekton-operator/templates/_helpers.tpl
aqeelat: Same — flattened in 42b857837.

charts/tekton-operator/templates/_helpers.tpl
aqeelat: Added in 42b857837, with one correction to the wording: the repositories are the ko-published binary slugs from the v0.81.0 release assets and the tag follows Chart.AppVersion (no @sha256 pinning), so the comment now says where the slugs come from and that all three must be updated together if a release renames them.

charts/tekton-operator/templates/common-rbac.yaml
aqeelat: Good catch — fixed in 42b857837: the info Role/RoleBinding and config-read ClusterRole/ClusterRoleBinding are now fullname-prefixed, so two installs no longer collide on cluster-scoped names. One deliberate exception: the Role's resourceNames entry stays the literal tekton-operator-info, because it refers to the ConfigMap that tkn version reads — that object keeps a fixed name.

charts/tekton-operator/templates/common-rbac.yaml
aqeelat: Valid concern — addressed in two parts: the kubernetes flavor's <fullname>-info Role/RoleBinding went back to their pre-PR names (so that particular orphan no longer occurs for kubernetes upgrades from released charts), and the PR description now carries an upgrade note with the exact kubectl cleanup commands for the objects that do get orphaned (the literal config-read pair on both flavors, and the literal info pair on openshift). A pre-upgrade hook felt like the wrong trade: it would need its own RBAC to delete RBAC.

aqeelat: Correction to my earlier reply — and to the orphan concern itself: `helm upgrade` *does* delete renamed objects. Helm diffs the stored release manifest against the newly rendered one and deletes anything removed (`originals.Difference(targets)` → delete in helm/pkg/kube/client.go; the `helm.sh/resource-policy: keep` annotation exists precisely to opt out of this, and same behavior in Helm 4 where server-side apply is now the default). Orphans only occur with `helm template | kubectl apply` (no prune) or adopted non-helm installs. I've updated the PR description accordingly — the multi-install collision fix from this review stands and was the right call.

aqeelat:
Developed with [opencode](https://opencode.ai) (a CLI coding agent, GLM model) under my direction — each commit carries an `Assisted-by` trailer per the [AI contribution policy](https://github.com/tektoncd/community/blob/main/ai-contribution-policy.md). I reviewed and verified everything before asking for review (chart PRs were validated by diffing `helm template` output against `kubectl kustomize`). Happy to walk through any specific change.

jkhelil:
@aqeelat We dont have e2e for helm
C an you please post proof that install is working on both openshift and kubernetes

Thank you

aqeelat:
**Manual verification against real clusters** — kubernetes flavor on kind (v1.34.0), openshift flavor on OKD (k8s v1.35.3). Chart built from this branch with `appVersion: v0.81.1` (the in-repo `devel` tag isn't published to ghcr); upgrade baseline is the released v0.81.1 chart.

**1. Fresh install — Kubernetes** (`release: tekton-operator`, `installCRDs=true`)
`tekton-operator` 2/2, `tekton-operator-webhook` 1/1, zero restarts. `tektonconfig/config`: `Ready=True ComponentsReady=True` (~4 min). 18 component pods Running in `tekton-pipelines` (pipelines, triggers, chains, results + postgres, dashboard, PAC). CM key `loglevel.tekton-operator-webhook` == container name.

**2. Fresh install — OpenShift** (`-n openshift-operators --set openshift.enabled=true`)
Operator 2/2, webhook 1/1; `tektonconfig/config` `v0.81.1 Ready=True` (~3.5 min). Full openshift set in `openshift-pipelines` incl. PAC and the console plugin. SA is fullname-templated (`tekton-operator`), pod image is the openshift slug (`operator-1d69a75f…:v0.81.1`).

**3. Upgrade v0.81.1 → this chart (both flavors)**
- OpenShift baseline demonstrates the original bug live: v0.81.1 creates SA `openshift-pipelines-operator` while deployments reference `tekton-operator` → pods `Forbidden: serviceaccount "tekton-operator" not found`, operator never starts. After upgrade: correct SA created, old literal SA auto-deleted, operator 2/2, `tektonconfig Ready=True`, `tektonaddons` Ready.
- Kubernetes: no CrashLoopBackOff, `tektonconfig` stays Ready through the upgrade.
- Renamed RBAC (`tekton-config-read-role` pair on both flavors) was **deleted automatically by `helm upgrade`** — no manual cleanup needed (helm diffs the stored release manifest; `helm.sh/resource-policy: keep` is the opt-out).

**4. `webhook.logLevel` wiring**
With a release name that makes fullname ≠ operator-name (`to`): v0.81.1 writes CM key `loglevel.to-tekton-operator-webhook` but names the container `tekton-operator-webhook` → value silently dead. This chart names the container `<fullname>-webhook` → match. `--set webhook.logLevel=debug` → CM flips to `debug`, webhook logs emit `"level":"debug"` lines.

**5. `PLATFORM=openshift`**
`env | grep PLATFORM` → `PLATFORM=openshift`; `tektonaddons/addon Ready=True`; `tektondashboards` CRD absent entirely (openshift CRD set).

**Two issues surfaced (not regressions of this PR):**
- **Upgrade caveat:** v0.81.1 ships `controllers` as an explicit value (the k8s list incl. `tektondashboard`). `helm upgrade --reuse-values` carries it into the openshift flavor, and the openshift binary exits fatally on `tektondashboard` (`validateControllerNamesOrDie`). Verified live: instant CrashLoop. Workaround `--set controllers=""` — verified green. Added a render-time guard in 4bdaa7e5f that fails the chart instead; it's the minimal flavor-mismatch blocklist — a fuller allowlist variant (typo detection, `tektoninstallerset` reservation) is available as a follow-up if maintainers prefer, and the commit is droppable if this is out of scope.
- Unrelated to this PR's changes, also present in v0.81.1 — filed separately as #4123: the kodata admission-webhook manifests hardcode `service.name: tekton-operator-webhook`, so with any helm release name ≠ `tekton-operator`, tektonconfig auto-install fails on the defaulting webhook.

PR #3958, opened 2026-08-20
fix(pipeline): validate controller ordinal buckets

### LINKED ISSUE #3957: Resolver requests time out when ordinal-mode bucket overrides exceed replicas
## Expected behavior

When StatefulSet ordinal mode is enabled, every configured resolver bucket should have an owning resolver replica. The Operator should reject or normalize configurations where the effective `config-leader-election-resolvers` bucket count differs from the effective `tekton-pipelines-remote-resolvers` StatefulSet replica count.

## Actual behavior

Top-level validation enforces `spec.pipeline.performance.buckets == spec.pipeline.performance.replicas` when `statefulset-ordinals` is enabled. However, `spec.pipeline.options` is applied after the performance transforms and can override either the resolver leader-election ConfigMap or resolver StatefulSet replicas, bypassing that invariant.

For example:

```yaml
kind: TektonConfig
metadata:
  name: config
spec:
  pipeline:
    performance:
      buckets: 4
      replicas: 4
      statefulset-ordinals: true
    options:
      configMaps:
        config-leader-election-resolvers:
          data:
            buckets: "8"
      statefulSets:
        tekton-pipelines-remote-resolvers:
          spec:
            replicas: 4
```

This configuration passes the top-level invariant but renders an eight-bucket resolver universe with only four ordinal replicas.

In StatefulSet ordinal mode, replica ordinal `N` owns only bucket `N`. With replicas `0` through `3`, buckets `4` through `7` have no owner. Since tektoncd/pipeline#10480, resolver reconcilers correctly honor bucket ownership and return `NewSkipKey` for non-owned requests. Requests assigned to the unowned buckets are therefore skipped by every resolver pod and eventually fail at the global ResolutionRequest timeout.

This presents as:

- ResolutionRequests remaining unresolved until the global timeout
- PipelineRuns failing with `CouldntGetTask`
- resolver workqueues appearing empty because skipped keys are forgotten
- no resolver-specific timeout errors because resolution never starts for those requests
- partial ResolutionRequest success, but near-total PipelineRun failure when each run needs several remote resources

## Steps to reproduce

1. Install the Operator with a Pipeline payload containing tektoncd/pipeline#10480.
2. Apply the `TektonConfig` above.
3. Verify the effective resources:

   ```console
   $ kubectl get configmap config-leader-election-resolvers -o jsonpath='{.data.buckets}'
   8
   $ kubectl get statefulset tekton-pipelines-remote-resolvers -o jsonpath='{.spec.replicas}'
   4
   ```

4. Create multiple PipelineRuns that resolve Tasks through a remote resolver.
5. Observe requests assigned to buckets without matching ordinals remain unresolved and eventually time out.

## Relevant implementation

- StatefulSet ordinal mode is documented as requiring buckets to equal replicas:
  https://github.com/tektoncd/operator/blob/6a5b57b302528874834bb6ec21f904475566873c/docs/TektonPipeline.md#L341-L350
- Current validation checks only the top-level performance values:
  https://github.com/tektoncd/operator/blob/6a5b57b302528874834bb6ec21f904475566873c/pkg/apis/operator/v1alpha1/performance_validation.go#L36-L44
- Additional options run after the performance and StatefulSet transformations:
  https://github.com/tektoncd/operator/blob/6a5b57b302528874834bb6ec21f904475566873c/pkg/reconciler/kubernetes/tektonpipeline/transform.go#L92-L115
- The resolver framework now correctly enforces bucket ownership:
  https://github.com/tektoncd/pipeline/blob/7289f7d7e1ee24cbc84a58b9f763e3869db8ab2e/pkg/remoteresolution/resolver/framework/reconciler.go#L89-L105
- Knative ordinal assignment creates the configured bucket universe but assigns only `Buckets()[ordinal]` to each pod:
  https://github.com/tektoncd/pipeline/blob/7289f7d7e1ee24cbc84a58b9f763e3869db8ab2e/vendor/knative.dev/pkg/leaderelection/context.go#L224-L246

## Suggested fix

Validate the effective rendered resolver bucket and replica values after applying additional options, or reject conflicting overrides while StatefulSet ordinal mode is enabled. A render-level regression test should cover both ConfigMap and StatefulSet overrides.

The existing documentation already states the invariant, so this should not be treated as documentation-only. A short warning about additional-option overrides could complement the validation fix.

waveywaves:
# Changes

- Validate the effective leader-election bucket counts for both the main Pipelines controller and remote-resolver controller against their rendered StatefulSet replica counts after additional options are applied.
- Preserve matching overrides and dynamic leader-election configurations.
- Reject mismatched or out-of-range bucket overrides in StatefulSet ordinal mode.
- Cover both controllers with render-level tests for bucket overrides, replica overrides, matching values, and dynamic leader election.

The existing performance documentation already states that buckets must match replicas in ordinal mode.

Fixes #3957.

# Submitter Checklist

review them:

# Release Notes

```release-note
Validate that Pipelines and resolver leader-election buckets match their controller replicas when StatefulSet ordinal mode is enabled.
```

### DIFF
--- a/pkg/reconciler/kubernetes/tektonpipeline/transform.go
+++ b/pkg/reconciler/kubernetes/tektonpipeline/transform.go
@@ -18,7 +18,9 @@ package tektonpipeline
 
 import (
 	"context"
+	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 
 	mf "github.com/manifestival/manifestival"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/kubernetes/tektonpipeline/transform.go
```diff
@@ -113,11 +115,53 @@ func filterAndTransform(extension common.Extension) client.FilterAndTransform {
 		if err := common.ExecuteAdditionalOptionsTransformer(ctx, manifest, pipeline.Spec.GetTargetNamespace(), pipeline.Spec.Options); err != nil {
 			return &mf.Manifest{}, err
 		}
+		if pipeline.Spec.Performance.StatefulsetOrdinals != nil && *pipeline.Spec.Performance.StatefulsetOrdinals {
+			if err := validateResolverStatefulSetOrdinals(manifest); err != nil {
```
jkhelil: @waveywaves We could have same kind of bug for pipelineruns like for resolutionrequest
Shoul we add same kind of validation for pipelinerun controller

waveywaves: Added. The post-options render validation is now shared by both ordinal-mode controller pairs:

- `config-leader-election-controller` / `tekton-pipelines-controller`
- `config-leader-election-resolvers` / `tekton-pipelines-remote-resolvers`

The regression matrix now exercises mismatched bucket overrides, mismatched replica overrides, matching overrides, out-of-range buckets, and dynamic leader election for both controllers. `make test` and `make lint` pass locally.

PR #4001, opened 2026-08-30
fix(networkpolicy): allow DNS egress to link-local resolvers (GKE Cloud DNS, NodeLocal DNSCache)

### LINKED ISSUE #4000: Generated NetworkPolicies break DNS on GKE clusters using Cloud DNS (link-local resolver cannot be matched by pod selector)
<details>
<summary>Environment</summary>

- operator v0.81.1, GKE Standard, Dataplane V2
- cluster created/updated with `--cluster-dns=clouddns --cluster-dns-scope=cluster`
- `profile: all` (NetworkPolicy defaults enabled)

</details>

## Symptom

`tekton-results-retention-policy-agent` crashloops:

```
lookup tekton-results-postgres-service.tekton-pipelines.svc.cluster.local
on 169.254.169.254:53: read udp ...->169.254.169.254:53: i/o timeout
fatal: Failed to connect to database: timed out waiting for the condition
```

## Root cause

On GKE clusters with Cloud DNS, kubelet writes `nameserver 169.254.169.254` (the node-local metadata server running the Cloud DNS data plane) into every pod's resolv.conf. The CoreDNS/kube-dns pods keep running but are unused.

`networkpolicy.DNSEgressRule` allows DNS egress only to a pod peer (`kube-system` / `k8s-app=kube-dns` on port 53). A podSelector/namespaceSelector peer can never match 169.254.169.254 — it is not a pod. Only an `ipBlock` peer can allow it. On Dataplane V2 the packets are dropped, causing total DNS failure for any pod whose only DNS allow is that rule.

Most generated policies also include `APIServerEgressRule()` (allow-all, since NP cannot select host-network endpoints), which incidentally permits the link-local resolver and masks the bug. The two policies without it fail visibly:

- `results-retention-policy-agent` (DNS rule + db_port egress only)
- `results-postgres` (DNS rule only; latent — postgres does no runtime lookups)

The same applies to NodeLocal DNSCache (`169.254.20.10`, Autopilot default / opt-in on Standard): it is a host-network DaemonSet unreachable by pod selectors.

## Workaround

Override the generated policy by name via `spec.networkPolicy.policies` (on `TektonConfig`) and add an `ipBlock: 169.254.169.254/32` peer on UDP+TCP 53. Takes effect on next reconcile; no pod restart needed.

## Proposed fix

In `DNSEgressRule` (Kubernetes platform only), add ipBlock peers `169.254.169.254/32` and `169.254.20.10/32` on port 53 alongside the existing kube-dns pod peer. Additive and zero-risk on clusters not using these resolvers — the extra peers simply never match. OpenShift path unchanged.

aqeelat:
# Changes

`DNSEgressRule` allowed DNS egress only to the kube-dns/CoreDNS pod peer. On GKE clusters using Cloud DNS, kubelet points every pod's resolv.conf at the link-local metadata server `169.254.169.254`, which no pod selector can match — so on Dataplane V2 clusters DNS egress is dropped and e.g. `tekton-results-retention-policy-agent` crashloops. NodeLocal DNSCache (`169.254.20.10`) has the same host-network property.

Adds `ipBlock` peers `169.254.169.254/32` and `169.254.20.10/32` (UDP+TCP 53) to the Kubernetes-platform DNS rule, alongside the existing pod peer. Purely additive — on clusters not using these resolvers the peers never match. OpenShift unchanged. Fixes #4000.

# Submitter Checklist

review them:

# Release Notes

```release-note
NetworkPolicies generated by the operator now allow DNS egress to link-local resolvers (GKE Cloud DNS 169.254.169.254 and NodeLocal DNSCache 169.254.20.10), fixing DNS failures on GKE clusters using Cloud DNS.
```

---

**AI assistance:** developed with [opencode](https://opencode.ai) (GLM model) under my direction; every commit carries an `Assisted-by:` trailer per the [AI contribution policy](https://github.com/tektoncd/community/blob/main/ai-contribution-policy.md). I drove scoping and decisions, reviewed every change, and verified each push — unit tests and `make test lint` for the touched packages.

### DIFF
--- a/pkg/reconciler/common/networkpolicy/platform.go
+++ b/pkg/reconciler/common/networkpolicy/platform.go
@@ -19,8 +19,15 @@ package networkpolicy
 // PlatformParams holds platform-specific values for building default NetworkPolicy rules.
 // It is an internal type — it never appears in CRD API fields.
 type PlatformParams struct {
-	DNSResolverNamespace     string
-	DNSResolverPodLabel      map[string]string
+	DNSResolverNamespace string
+	DNSResolverPodLabel  map[string]string
+	// DNSResolverIPBlocks are link-local resolver IPs kubelet may write into pod
+	// resolv.conf instead of the kube-dns Service: GKE Cloud DNS (169.254.169.254),
+	// NodeLocal DNSCache (169.254.20.10, GKE Autopilot default) and the AWS VPC
+	// resolver (169.254.169.253, reached directly only with non-default DNS
+	// configuration). They are host-network endpoints no pod selector can match,
+	// so only ipBlock peers can allow them. Empty on OpenShift.
+	DNSResolverIPBlocks      []string
 	PrometheusNamespaceLabel map[string]string
 	// DNSPort is the DNS resolver pod port. Kubernetes CoreDNS uses 53; OpenShift DNS
 	// uses 5353 (OVN-K8s enforces NetworkPolicy after DNAT, so pod port applies).
@@ -30,8 +37,16 @@ type PlatformParams struct {
 // KubernetesPlatformDefaults returns PlatformParams for vanilla Kubernetes.
 func KubernetesPlatformDefaults() PlatformParams {
 	return PlatformParams{
-		DNSResolverNamespace:     "kube-system",
-		DNSResolverPodLabel:      map[string]string{"k8s-app": "kube-dns"},
+		DNSResolverNamespace: "kube-system",
+		DNSResolverPodLabel:  map[string]string{"k8s-app": "kube-dns"},
+		DNSResolverIPBlocks: []string{
+			// 169.254.169.254/32: GKE Cloud DNS forwarder (link-local, host-network).
+			"169.254.169.254/32",
+			// 169.254.169.253/32: AWS VPC resolver (Route 53 Resolver, link-local).
+			"169.254.169.253/32",
+			// 169.254.20.10/32: NodeLocal DNSCache default localip (GKE Autopilot).
+			"169.254.20.10/32",
+		},
 		PrometheusNamespaceLabel: map[string]string{"kubernetes.io/metadata.name": "monitoring"},
 		DNSPort:                  53,
 	}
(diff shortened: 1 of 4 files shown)


pkg/reconciler/common/networkpolicy/platform.go
```diff
@@ -32,6 +38,7 @@ func KubernetesPlatformDefaults() PlatformParams {
 	return PlatformParams{
 		DNSResolverNamespace:     "kube-system",
 		DNSResolverPodLabel:      map[string]string{"k8s-app": "kube-dns"},
+		DNSResolverIPBlocks:      []string{"169.254.169.254/32", "169.254.20.10/32"},
```
jkhelil: Add comment here
// 169.254.169.254/32: GKE Cloud DNS forwarder (link-local, host-network).

aqeelat: Not just GKE. I think aws and azure also have it.

BTW, I saw it in some charts as `169.254.0.0/16`
Do we want that?

jkhelil: the GCE/AWS IMDS IP is already in the list as a /32, so /16 doesnt add any functional value

aqeelat: Added in bfe8b88ed, along with per-entry comments. On the /16 question: it turns out /16 would add one thing the two /32s lack — 169.254.169.253, the AWS VPC resolver (Route 53 Resolver is link-local too). I've added that as a /32 rather than widening to /16: with the three documented resolvers covered, the /16 would only add non-default NodeLocal localips, at the cost of exposing port 53 to the entire link-local range. On whether AWS/Azure "also have" 169.254.169.254 — as a DNS endpoint only GCP does; on AWS that address is IMDS (no DNS listener), and Azure's IMDS is 169.254.169.50 with Azure DNS at 168.63.129.16 (not link-local at all, so out of scope here). The rule stays port-scoped to 53, so IMDS on 80 remains blocked either way.

jkhelil:
@aqeelat PLease squash your commit; and will be happy to approve

aqeelat:
Squashed to a single commit (bfe8b88ed) — thanks!

PR #4029, opened 2026-09-03
fix(tektonresult): default route TLS termination to reencrypt

divyansh42:
# Changes

The Results API route on OpenShift was defaulting to `edge` TLS termination,
which causes TLS handshake failures when clients try to connect. The router
terminates TLS at the edge and forwards **unencrypted** traffic to the gRPC
backend — but the backend only speaks TLS.

Additionally, when `passthrough` was used as a workaround, clients had to
pass `--insecure-skip-tls-verify` because the OpenShift service-serving
certificate was not trusted by external clients.

**`reencrypt`** is the correct termination type:
- The router presents a **trusted** certificate to external clients (no `--insecure` needed)
- The router **re-encrypts** to the backend using the OpenShift service-serving cert
- End-to-end encryption is maintained

### Upgrade migration

Changing the default only affects **new** installs — existing installs persist
the previous `edge` value in their `TektonConfig`/`TektonResult` CR, so they
would stay broken after upgrading. A version-gated pre-upgrade migration
(`migrateResultsRouteTLSToReencrypt`) updates the Results route TLS termination
from `edge` → `reencrypt` on `TektonConfig` (the source of truth), which then
syncs to the `TektonResult` CR. It runs on OpenShift only and migrates **only**
`edge`; values set manually to `reencrypt` or `passthrough` are left untouched.

### Changes made

| File | Change |
|---|---|
| `tektonresult_defaults.go` | Default `route_tls_termination` changed from `edge` → `reencrypt` |
| `route-rbac/rbac.yaml` (kodata) | Route TLS termination updated from `passthrough` → `reencrypt` |
| `route-rbac/rbac.yaml` (testdata) | RBAC `resourceNames` fixed: `tekton-results-api-service` → `tekton-results-api` (matches actual Route name) |
| `tektonresult_validation.go` | Extracted `validateRouteTLSTermination`; only `edge`, `reencrypt`, `passthrough` accepted |
| `tektonconfig_validation.go` | Reject invalid `spec.result.route_tls_termination` at the parent CR level too |
| `upgrade.go` / `pre_upgrade.go` | Added version-gated pre-upgrade migration `edge` → `reencrypt` |
| `docs/TektonResult.md`, `docs/TektonConfig.md` | Document that only `reencrypt` is supported on OpenShift; note the upgrade migration |
| `*_test.go` | Tests for defaults, `TektonResult`/`TektonConfig` validation, route injection, and the migration |

# Submitter Checklist

review them:

# Release Notes

```release-note
fix(tektonresult): change default OpenShift Results API route TLS
termination from `edge` to `reencrypt`, and add a pre-upgrade
migration that flips existing installs from `edge` to `reencrypt`.
This fixes TLS handshake failures and eliminates the need for
`--insecure-skip-tls-verify` when accessing the Results API via the
OpenShift route. Also adds validation to reject unsupported
`route_tls_termination` values.
```

### DIFF
--- a/cmd/openshift/operator/kodata/static/tekton-results/route-rbac/rbac.yaml
+++ b/cmd/openshift/operator/kodata/static/tekton-results/route-rbac/rbac.yaml
@@ -27,7 +27,7 @@ rules:
   verbs:
   - get
   resourceNames:
-  - tekton-results-api-service
+  - tekton-results-api
 ---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
@@ -58,5 +58,5 @@ spec:
   port:
     targetPort: 8080
   tls:
-    termination: passthrough
+    termination: reencrypt
     insecureEdgeTerminationPolicy: Redirect
--- a/docs/TektonConfig.md
+++ b/docs/TektonConfig.md
@@ -330,6 +330,8 @@ manualApproval:
 
 Result section allows user to customize the Tekton Result component, Refer to [Result Spec](https://github.com/tektoncd/operator/blob/main/docs/TektonResult.md#spec) section in TektonResult for available options.
 
+On OpenShift, `route_tls_termination` controls TLS termination on the Results route. Only `reencrypt` is supported (and is the default). `edge` breaks the route outright (no client flag can fix it). `passthrough` also works, but the client connects to the backend's service-serving certificate, so it must trust the cluster's service CA (or pass `--insecure`). `reencrypt` instead presents the ingress router certificate; note that by default this is cluster-signed rather than publicly trusted, so clients may still need the cluster's ingress CA (or `--insecure`) unless the router uses a publicly trusted certificate. See [route_tls_termination](https://github.com/tektoncd/operator/blob/main/docs/TektonResult.md#property-route_tls_termination-openshift-only) for details.
+
 Default Result configuration in TektonConfig looks like following if user doesn't specified any configuration options
 
 Example:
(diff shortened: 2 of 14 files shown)


divyansh42:
Addressed the Copilot review comments in the follow-up commit (06d301ff5):

1. **Stale test expectations** — Fixed all 6 occurrences of `RouteTLSTermination: "edge"` in `TestResult_SetDefaultsBucketsAutoDefaulting` to expect `"reencrypt"`.

2. **TektonConfig bypasses validation** — Extracted the TLS termination check into `ResultsAPIProperties.validateRouteTLSTermination()` and wired it into `TektonConfig.Validate` (at `spec.result.route_tls_termination`), with a corresponding test `Test_ValidateTektonConfig_InvalidResultRouteTLSTermination`.

3. **Existing installations not migrated** — The `edge` termination never worked correctly (it causes TLS handshake failures), so any existing cluster with `edge` is already broken. Rather than adding a forced migration (which risks overwriting deliberate user choices for users who set `edge` knowingly), we are treating this as a **known upgrade action**: users on existing installations who have `edge` persisted should manually patch their `TektonResult` or `TektonConfig` CR to `reencrypt`.

pratap0007:
Don't we also need to handle the upgrade scenario (updating TLS termination `edge` to `reencrypt`)?

jkhelil:
@divyansh42 can you adress @pratap0007 concerns? what happens on upgrade, please post a proof of work

enarha (approved):
/approve

jkhelil:
@divyansh42 can we deprecate passthroug ?
and eventually remove edge from valide values ?

divyansh42:
> @divyansh42 can you adress @pratap0007 concerns? what happens on upgrade, please post a proof of work

I have handled the upgrade but still need to test end to end. Will update once done.

divyansh42:
/hold cancel

I have verifed this on the cluster, by patching the version with the previous release version and it is able to upgrade to re-encrypt.
/cc @jkhelil @pratap0007

PR #4065, opened 2026-09-09
fix(pipelinesascode): guard nil Settings in validate

### LINKED ISSUE #4057: Cannot disable PipelinesAsCode in TektonConfig: validating webhook panics with "assignment to entry in nil map"
# Expected Behavior

Setting `spec.platforms.kubernetes.pipelinesAsCode.enable: false` (or the deprecated `spec.addon.enablePipelinesAsCode: false`) on `TektonConfig` should disable PipelinesAsCode and be admitted.

# Actual Behavior

The update is rejected, and `tekton-operator-webhook` panics on every attempt:

```
Error from server (InternalError): Internal error occurred: failed calling webhook
"validation.webhook.operator.tekton.dev": failed to call webhook: Post
"https://tekton-operator-webhook.tekton-operator.svc:443/resource-validation?timeout=10s":
EOF (retried 5 times)
```

```
http: panic serving [...]: assignment to entry in nil map
panic({0x193dc20?, 0x345c720?})
	runtime/panic.go:860 +0x12c
github.com/openshift-pipelines/pipelines-as-code/pkg/params/settings.getHubCatalogs(0x..., 0x..., 0x0, 0x...)
	github.com/openshift-pipelines/pipelines-as-code@v0.49.0/pkg/params/settings/default.go:21 +0x8c
github.com/openshift-pipelines/pipelines-as-code/pkg/params/settings.SyncConfig(0x..., 0x..., 0x0, 0x...)
	github.com/openshift-pipelines/pipelines-as-code@v0.49.0/pkg/params/settings/config.go:122 +0x3c
github.com/tektoncd/operator/pkg/apis/operator/v1alpha1.(*PACSettings).validate(...)
	github.com/tektoncd/operator/pkg/apis/operator/v1alpha1/openshiftpipelinesascode_validation.go:58 +0x204
github.com/tektoncd/operator/pkg/apis/operator/v1alpha1.(*TektonConfig).Validate(...)
	github.com/tektoncd/operator/pkg/apis/operator/v1alpha1/tektonconfig_validation.go:81 +0x568
```

1. Install operator v0.81.0 (or current `main`) on Kubernetes (non-OpenShift) with `profile: all`.
2. Let the default `TektonConfig` reconcile once, so PAC is auto-enabled and `platforms.kubernetes.pipelinesAsCode` is populated.
3. Run `kubectl patch tektonconfig config --type=merge -p '{"spec":{"platforms":{"kubernetes":{"pipelinesAsCode":{"enable":false}}}}}'` — it fails with the error above, and `kubectl logs -n tekton-operator deploy/tekton-operator-webhook` shows the panic.

# Additional Info

- Kubernetes version:

```
Client Version: v1.33.4
Server Version: v1.36.2-eks-bca9cf6
```

- Tekton Pipeline version:

```
➜ tkn version
Client version: 0.42.0
Pipeline version: v1.15.0
Triggers version: v0.37.0
Dashboard version: v0.71.0
```

- AI analysis

`PACSettings.Settings` is nil (third arg `0x0`), and `getHubCatalogs` writes into it unconditionally at `default.go:21`. `setPACDefaults()` is the only code that initializes `Settings`, and it is gated on PAC being *enabled* (`tektonconfig_defaults.go`, Kubernetes branch):

```go
if tc.Spec.Addon.EnablePAC != nil && !*tc.Spec.Addon.EnablePAC {
    tc.Spec.Platforms.Kubernetes.PipelinesAsCode.Enable = ptr.Bool(false)
    tc.Spec.Platforms.Kubernetes.PipelinesAsCode.PACSettings.Settings = nil
}

if *tc.Spec.Platforms.Kubernetes.PipelinesAsCode.Enable {
    tc.Spec.Platforms.Kubernetes.PipelinesAsCode.PACSettings.setPACDefaults(logger)  // skipped when disabled
}
```

It guards `Settings == nil` internally but never runs on the disable path, and `(*PACSettings).validate` then calls `SyncConfig(..., ps.Settings, ...)` unguarded — so defaulting produces an object validation cannot process.

Both routes to disabling hit this, so there is no field-level workaround: `addon.enablePipelinesAsCode: false` explicitly nils `Settings`, and `platforms.kubernetes.pipelinesAsCode.enable: false` leaves it never initialized. The OpenShift branch has the same pattern.

This is permanent rather than transient — every retry fails identically, and because the whole object is rejected, unrelated fields (`spec.pipeline.*`, `spec.trigger.*`, `spec.dashboard.*`, …) cannot be updated either while the disable is requested.

Initializing `Settings` to `map[string]string{}` instead of `nil` when disabling, or moving `setPACDefaults` outside the `*Enable` check, would fix it. `OpenShiftPipelinesAsCode.SetDefaults` in `openshiftpipelinesascode_defaults.go` already has the equivalent guard.

pujitha24:
# Changes

Fix a validating-webhook panic ("assignment to entry in nil map") that is
triggered whenever `TektonConfig.Spec.Platforms.{Kubernetes,OpenShift}.PipelinesAsCode`
is disabled, via either `pipelinesAsCode.enable: false` or the deprecated
`spec.addon.enablePipelinesAsCode: false`.

`SetDefaults` sets `PACSettings.Settings` to `nil` on the disable path (both
platform branches), but `(*PACSettings).validate` unconditionally passes
`ps.Settings` into the vendored `pipelines-as-code` `SyncConfig`, which
writes into that map unconditionally (`getHubCatalogs`, `default.go:21`).
A nil map panics on write.

The fix guards `ps.Settings` to an empty map before calling `SyncConfig`,
in `(*PACSettings).validate` — the single method used by both the
`OpenShiftPipelinesAsCode.Validate` and `TektonConfig.Validate` (Kubernetes
and OpenShift branches) entry points, so one guard covers all three call
sites.

A regression test reproduces the exact panic reported in the issue
(`PACSettings{Settings: nil}` through `OpenShiftPipelinesAsCode.Validate`)
and asserts it now returns cleanly.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix a validating webhook panic when disabling PipelinesAsCode in
TektonConfig (via `pipelinesAsCode.enable: false` or the deprecated
`addon.enablePipelinesAsCode: false`), which previously rejected the
update and blocked unrelated TektonConfig field changes as well.
```

---

Fixes #4057

### DIFF
--- a/pkg/apis/operator/v1alpha1/openshiftpipelinesascode_validation.go
+++ b/pkg/apis/operator/v1alpha1/openshiftpipelinesascode_validation.go
@@ -54,6 +54,12 @@ func (pac *OpenShiftPipelinesAsCode) Validate(ctx context.Context) *apis.FieldEr
 func (ps *PACSettings) validate(logger *zap.SugaredLogger, path string) *apis.FieldError {
 	var errs *apis.FieldError
 
+	// Settings may be nil when PAC is disabled (SetDefaults nils it out), but
+	// SyncConfig writes into it unconditionally, so guard against a nil map here.
+	if ps.Settings == nil {
+		ps.Settings = map[string]string{}
+	}
+
 	defaultPacSettings := pacSettings.Settings{}
 	if err := pacSettings.SyncConfig(logger, &defaultPacSettings, ps.Settings, pacSettings.DefaultValidators(), http.DefaultClient); err != nil {
 		errs = errs.Also(apis.ErrInvalidValue(err, fmt.Sprintf("%s.settings", path)))
--- a/pkg/apis/operator/v1alpha1/openshiftpipelinesascode_validation_test.go
+++ b/pkg/apis/operator/v1alpha1/openshiftpipelinesascode_validation_test.go
@@ -194,3 +194,26 @@ func TestValidateAddtionalPACControllerInvalidSetting(t *testing.T) {
 	err := opacCR.Validate(context.TODO())
 	assert.Equal(t, "invalid value: invalid value: invalid value for URL, error: parse \"test/path\": invalid URI for request: validation failed for field custom-console-url: spec.additionalPACControllers.settings", err.Error())
 }
+
+// TestValidateNilSettings guards against a regression where a nil
+// PACSettings.Settings map (as left behind when PAC is disabled via
+// SetDefaults) reaches SyncConfig and panics with "assignment to entry in
+// nil map".
+func TestValidateNilSettings(t *testing.T) {
+	opacCR := &OpenShiftPipelinesAsCode{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "name",
+			Namespace: "namespace",
+		},
+		Spec: OpenShiftPipelinesAsCodeSpec{
+			CommonSpec: CommonSpec{
+				TargetNamespace: "openshift-pipelines",
+			},
+			PACSettings: PACSettings{
+				Settings: nil,
+			},
+		},
+	}
+	err := opacCR.Validate(context.TODO())
+	assert.Assert(t, err == nil, "unexpected validation error: %v", err)
+}


theakshaypant (commented):
@pujitha24 Were you able to verify the nil panic is no longer happening on a live deployment?

Unrelated to the changes, what is a mutex in Go, and when would you use one?

pujitha24:
No, I haven't run this against a live deployment. What I verified is the regression test (`TestValidateNilSettings` in `openshiftpipelinesascode_validation_test.go`), which reproduces the exact panic path — `OpenShiftPipelinesAsCode.Validate` with `PACSettings.Settings: nil` — and confirms it now returns cleanly instead of panicking, plus the rest of the package's test suite and `make test lint` passing. If a live-cluster check is needed before merge I'm happy to do that, just say so.

On the mutex question: `sync.Mutex` gives mutual exclusion so only one goroutine holds the lock at a time — you'd use one when multiple goroutines read/write the same shared state (a map, counter, etc.) and you need to serialize access to avoid a data race. Wasn't needed here since `ps.Settings` is local to a single validation call, not shared across goroutines.

theakshaypant (commented):
@pujitha24 Would be very helpful if you can perform the check deploying your changes, maybe even on kind.

Additionally, can you provide an inline review of the changes in this PR?

PR #4067, opened 2026-09-09
fix(openshift): handle missing TektonConfig during TLS resolution

wangyusheng1985:
# Changes

Make OpenShift component TLS configuration resolution safe when the optional TektonConfig object is absent. A Kubernetes API NotFound result or a nil lister object now produces no TLS environment variables and no retry-triggering error; other lister errors continue to propagate.

# Submitter Checklist

- [x] Includes focused regression tests
- [x] Updated nearby behavior documentation
- [x] No API or dependency changes

# Validation

- `gofmt -d pkg/reconciler/openshift/common/tlsprofile.go pkg/reconciler/openshift/common/tlsprofile_test.go && git diff --check`
- `go test ./pkg/reconciler/openshift/common`
- `go test -race ./pkg/reconciler/openshift/common`

```release-note
NONE
```

Validation observed for this change:
- `gofmt -d pkg/reconciler/openshift/common/tlsprofile.go pkg/reconciler/openshift/common/tlsprofile_test.go && git diff --check`
- `go test ./pkg/reconciler/openshift/common`
- `go test -race ./pkg/reconciler/openshift/common`

### DIFF
--- a/pkg/reconciler/openshift/common/tlsprofile.go
+++ b/pkg/reconciler/openshift/common/tlsprofile.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


jkhelil:
Hi @wangyusheng1985, thank you for your first contribution we really appreciate it!
Could you please explain how you came across this issue and raised this PR? Did you encounter this bug in a real-world scenario while running Tekton Operator?
Since this function is specifically related to OpenShift, could you also provide some evidence or steps showing that the change resolves the issue?

